PR #24, opened 2015-02-10
fix failed to allocate mcode memory on some android devices

chrox:
This should fix #1416.


hwhw:
Did you have a chance to test it yet?

chrox:
Yes, it works on my Meizu MX4.

PR #27, opened 2015-04-15
Implement io.popen

chrox:
Unfortunately the screen brightness api is not ready for use now. The brightness is indeed written to system settings but the value is not applied for some reason.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -5,6 +5,7 @@
       android:versionName="1.0">
     <uses-sdk android:minSdkVersion="9" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+    <uses-permission android:name="android.permission.WRITE_SETTINGS" />
     <uses-permission android:name="android.permission.INTERNET" />
     <application android:label="@string/app_name"
         android:hasCode="false" android:debuggable="true"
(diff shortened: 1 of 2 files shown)


hwhw:
Maybe it's switched to automatic brightness? Will manual setting some other value override that?

chrox:
I have tried disable automatic brightness but without luck. It seems the
limitation of Android system. And I also tried to set screen brightness
attribute (see the commented part in this PR) according to this:
http://stackoverflow.com/questions/7646865/changing-screen-brightness-programmatically-as-with-the-power-widget/7658364#7658364,
but it raises an exception saying that the native activity thread is not
the UI thread and cannot touch the layout attributes. Maybe we need a way
to run code in the UI thread. Is this possible?

On Wed, Apr 15, 2015 at 6:43 PM, HW <notifications@github.com> wrote:

> Maybe it's switched to automatic brightness? Will manual setting some
> other value override that?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/koreader/android-luajit-launcher/pull/27#issuecomment-93317141>
> .
>



-- 
Huang Xin, Ph.D. student
Laboratory of Vision Restoration
School of Basic Medical Sciences
Peking University
38 Xueyuan Road
Beijing, China 100083
Tel 86-10-8280-2972
Fax 86-10-8280-2877

houqp:
We can use message passing to communicate with UI thread and write the brightness adjustment code in Java. Also from what I heard, Mozilla Fennec is also completely written in native code and seems like it supports screen brightness setting [1][2]. So maybe there is a way to do it in native code.

[1]: https://wiki.mozilla.org/Fennec/Features/Reader
[2]: https://bugzilla.mozilla.org/show_bug.cgi?id=616664

hwhw:
It should be possible to join the main UI thread in order to make such a change in native code. I have a strong feeling that I already did such a thing or at least have seen it somewhere. Now, where would that be?

PR #29, opened 2015-06-15
fix serious bug that accessing deleted local reference

chrox:


hwhw:
Whoa, good find! This might have caused all kinds of weird problems.

PR #32, opened 2015-09-12
update luajit to the same version as in koreader base

chrox:
Hope it will improve the poor usability of recent releases.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "luajit-2.0"]
 	path = luajit-2.0
-	url = http://luajit.org/git/luajit-2.0.git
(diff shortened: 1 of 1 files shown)


houqp:
LuaJIT recently just moved to Github, should we use that repo instead?
https://www.reddit.com/r/Compilers/comments/3i2n9x/luajit_moved_to_github/

chrox:
That will be great. Here we go https://github.com/LuaJIT/LuaJIT.

houqp:
This is going to be the main official repo from now on. Not sure how the future of LuaJIT goes since Mike won't maintain it anymore ;(

PR #43, opened 2016-09-09
Adjust for latest NDK

tobias-93:
In the latest NDK, the toolchain version is 4.9
However, this script uses the folder with 4.6, so it cannot find the files.
When adjusted to 4.9, the android release of KOreader builds fine.

### DIFF
--- a/mk-luajit.sh
+++ b/mk-luajit.sh
@@ -19,7 +19,7 @@ clean)
 armeabi)
 	# Android/ARM, armeabi (ARMv5TE soft-float), Android 2.2+ (Froyo)
 	NDKABI=8
-	NDKVER=$NDK/toolchains/arm-linux-androideabi-4.6
+	NDKVER=$NDK/toolchains/arm-linux-androideabi-4.9
 	NDKP=$NDKVER/prebuilt/$BUILD_ARCH/bin/arm-linux-androideabi-
 	NDKF="--sysroot $NDK/platforms/android-$NDKABI/arch-arm"
 	rm -rf "$DEST"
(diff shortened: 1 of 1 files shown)


houqp:
Thanks for the update. Could you also document the minimal working NDK revision that has 4.9 TC? So that people know they are using an unsupported version.

tobias-93:
This is from NDK 11 onwards. Do you want me to document that in the koreader main repo or in this repo?

houqp:
It's better if you can detect the revision at the beginning of the script and print out an error message. But documenting it in this repo's README is good enough.

houqp:
Ideally, we should document all android toolchain dependency related stuff in this repo and link it from KOReader's README.

tobias-93:
@houqp I agree, but I'm not sure what is related to the android toolchain version as I'm not an expert on this project :wink:.
I did add some check to the script, as you can see.

houqp:
LGTM, thanks!

Hzj-jie:
@tobias-93 Are you sure koreader can build with NDK11? I found some errors which blocks us from using NDK higher than 9d, but I remember I have fixed part of them.

chrox:
Surely we can as in the unit test https://github.com/koreader/koreader-base/blob/master/.travis.yml#L19.

Hzj-jie:
Emmm, it looks weird, 4.9 toolchain is not by default included in NDK 9c. Have you installed standalone toolchains separately?

tobias-93:
@Hzj-jie As I said in the discussion already, it works from the NDK version 11 onwards. I didn't use version 9c or lower, I've only used version 12b and I've got it working using that.

houqp:
I just tried 12b, I am surprised it works for you. I was getting all kinds of error while building glib and gettext :P

tobias-93:
@houqp I got some errors when I tried at first, but then I tried starting anew with a fresh checkout and it all went fine. One of the things I messed up was using the NDK through a 'shared folder' on Virtualbox (I'm developing on a VM on my Windows machine), but I'm sure nobody will be as stupid as I am 😄

houqp:
@Hzj-jie I adapted the script so it should work for r11c now. See https://github.com/koreader/android-luajit-launcher/pull/45 

We need to setup travis build for this repo too since the build in koreader-base does not run the mk-luajit.sh script in this repo.

@tobias-93 turns out it was me being stupid and added a wrong configure option in gettext build. Now I am back to r12b :)

Hzj-jie:
Got it, thank you.

PR #50, opened 2017-05-29
[wip] Android Fullscreen switcher

mwoz123:
Initial commit for Android Fullscreen on/off switch (enable showing android top menu bar).

I'm not sure if those are correct:
"(I)V",
ffi.new('bool'

Help appreciated.

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1339,13 +1339,24 @@ local function run(android_app_state)
             )
         end)
     end
+    android.isFullscreen = function()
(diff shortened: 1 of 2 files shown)


houqp:
`(I)V` should probably work, but `(Z)V` is more accurate since the java signature actually takes a boolean, see: http://www.rgagnon.com/javadetails/java-0286.html.

mwoz123:
@houqp Thanks, fixed.

houqp:
Thanks for the patch!

PR #51, opened 2017-06-05
[fix] fix #50

chrox:


houqp:
ooops, we need CI for this repo :(

PR #54, opened 2017-07-11
Get screen brightness from Android API

Hzj-jie:
This is one of the changes to fix issue https://github.com/koreader/koreader/issues/3004.

### DIFF
--- a/src/org/koreader/launcher/MainActivity.java
+++ b/src/org/koreader/launcher/MainActivity.java
@@ -10,7 +10,13 @@ import android.os.Bundle;
 import android.view.View;
 import android.util.Log;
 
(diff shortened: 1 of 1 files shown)


Hzj-jie:
p.s. my android build breaks,

configure: error: in `/home/hzj_jie/git/koreader/koreader/base/thirdparty/libpng/build/arm-linux-androideabi/libpng-prefix/src/libpng-build':
configure: error: C compiler cannot create executables

Any thoughts?

houqp:
Do you get this error from a clean build? Do you get more info with `-v` flag enabled?

Hzj-jie:
I remember the error log mentioned that some functions could not be found.
I cannot connect to my machine now. I will provide more information later.

PR #55, opened 2017-07-18
adding mimetypes to open more supported files

silberzwiebel:
- based on https://github.com/SufficientlySecure/document-viewer/blob/master/document-viewer/src/main/AndroidManifest.xml
- fixes https://github.com/koreader/koreader/issues/3006 for PDFs (tested using the approach written in the issue; I did not build the app by myself, just changed this file in the .apk)
- I added other mimetypes based on the supported file types as written in https://github.com/koreader/koreader/blob/master/README.md but couldn't test them; you might want to add more

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -24,6 +24,22 @@
             </intent-filter>
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
+                <action android:name="android.intent.action.EDIT" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:mimeType="application/pdf" />
+                <data android:mimeType="image/djvu" />
+                <data android:mimeType="image/vnd.djvu" />
+                <data android:mimeType="image/x-djvu". />
+                <data android:mimeType="application/x-cbz" />
+                <data android:mimeType="application/epub+zip" />
+                <data android:mimeType="application/x-fb2" />
+                <data android:mimeType="application/x-mobipocket-ebook" />
+                <data android:mimeType="application/msword" />
+                <data android:mimeType="application/x-chm" />
+                <data android:mimeType="text/plain" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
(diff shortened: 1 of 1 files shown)


AndroidManifest.xml
```diff
@@ -24,6 +24,22 @@
             </intent-filter>
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
+                <action android:name="android.intent.action.EDIT" />
```
Frenzie: Edit? Odd to see it here; no wonder that wasn't in there.

Frenzie:
Thanks!

AndroidManifest.xml
silberzwiebel: Oh, yeah. I deleted that line and tested again -- apparaently, it is not needed for the bugfix. And I assume that KOreader can't edit files, right? So, feel free to just delete the line ...

Frenzie:
Oh, I'd assumed it was a necessary workaround for something. D'oh!

On Jul 19, 2017 12:22 PM, "silberzwiebel" <notifications@github.com> wrote:

*@silberzwiebel* commented on this pull request.
------------------------------

In AndroidManifest.xml
<https://github.com/koreader/android-luajit-launcher/pull/55#discussion_r128208274>
:

> @@ -24,6 +24,22 @@
             </intent-filter>
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
+                <action android:name="android.intent.action.EDIT" />

Oh, yeah. I deleted that line and tested again -- apparaently, it is not
needed for the bugfix. And I assume that KOreader can't edit files, right?
So, feel free to just delete the line ...

—
You are receiving this because you modified the open/close state.

Reply to this email directly, view it on GitHub
<https://github.com/koreader/android-luajit-launcher/pull/55#discussion_r128208274>,
or mute the thread

AndroidManifest.xml
Frenzie: If you save your highlights then KOReader generates a new PDF file (although I just noticed those highlights show up really strangely in Evince, but that aside), so for some files it can technically edit it. But I do think it'd be odd to advertise it as an actual editor so I'll take it out.

PR #59, opened 2017-09-16
[fix] Normalize Android build

Frenzie:
There are different numbers all over the place, but I *think* the code points toward Android 4 (platform-14) as an intended target.

Also, by using `NDKABI=${NDKABI:-14}` we allow for environment variable overrides. It only sets the default of 14 if $NDKABI isn't set. This is in the same spirit as https://github.com/koreader/koreader/pull/3064


Makefile
```diff
@@ -9,7 +9,7 @@ apk: local.properties project.properties
 	ant debug
 
 local.properties project.properties:
-	android update project --path . -t android-19
+	android update project --path . -t android-14
```
Frenzie: Ah, never mind this one. You need android-16 or higher or it won't build.

PR #62, opened 2017-09-23
fix loading libandroid on MIUI ROMs

chrox:
Xiaomi MIUI ROMs won't load libandroid.so to the global namespace. We have to load the library by ourselves.

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1105,10 +1105,14 @@ end
 
 -- Android specific
 
+-- Some Android roms won't load libandroid.so to the global namespace thus
+-- we load it by ourselves.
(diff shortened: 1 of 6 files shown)


assets/android.lua
```diff
@@ -1105,10 +1105,14 @@ end
 
 -- Android specific
 
+-- Some Android roms won't load libandroid.so to the global namespace thus
```
Frenzie: Android continues to confound me. :-)

chrox: Didn't know this until I got a Xiaomi Mix. :) And after this fix there will be much less crashes on Android. And the No. 1 issue will be ANR which has something to do with the JIT failure. Hope we can work out that.

PR #65, opened 2017-10-09
Android Wifi Manager interface

mwoz123:
Bridge/interface to enable,disable, check status Wifi on Android

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -8,6 +8,8 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.WRITE_SETTINGS" />
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE"></uses-permission>
+    <uses-permission android:name="android.permission.CHANGE_WIFI_STATE"></uses-permission>
     <application android:icon="@drawable/icon" android:label="@string/app_name"
         android:hasCode="true" android:debuggable="true"
         android:largeHeap="true" android:vmSafeMode="true">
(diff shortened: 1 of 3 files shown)


AndroidManifest.xml
```diff
@@ -8,6 +8,8 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.WRITE_SETTINGS" />
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE"></uses-permission>
```
Frenzie: Could you change the XML to self-close just like the rest?

`<uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />`

src/org/koreader/launcher/MainActivity.java
```diff
@@ -14,6 +14,8 @@
 import android.graphics.Rect;
 import android.view.Window;
 import android.util.DisplayMetrics;
+import android.net.wifi.WifiManager;
+import android.content.Context;
```
Frenzie: Is there a particular reason it's here rather than grouped with the other `android.content.*`?

src/org/koreader/launcher/MainActivity.java
```diff
@@ -152,6 +154,19 @@ public boolean isFullscreen() {
...
+    public void setWifiEnabled(final boolean enabled) {
+        this.getWifiManager().setWifiEnabled(enabled);
+    }
+
+    public boolean isWifiEnabled() {
+        return this.getWifiManager().isWifiEnabled();
+    }
+
+    private WifiManager getWifiManager() {
+	return (WifiManager) this.getSystemService(Context.WIFI_SERVICE);
```
Frenzie: Nitpick and sorry I missed it before, but this should be 8 spaces, not a tab. :-)

mwoz123: Did I already wrote how stupid is replacing Tabs with 4 spaces?
Guys, come on, we're in 21 century ! :P

Frenzie: I prefer tabs over spaces myself, but I should point out that even if we used spaces it should've been two instead of one. :-P

Frenzie:
Thanks!

PR #73, opened 2017-10-23
[w.i.p.] isWifiEnabled and isFullscreen use JNI:callBooleanMethod

mwoz123:
in response to https://github.com/koreader/koreader/pull/3396#pullrequestreview-71096493

Commit from phone. 
needs testing

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1360,13 +1360,13 @@ local function run(android_app_state)
     end
     android.isFullscreen = function()
         return JNI:context(android.app.activity.vm, function(JNI)
-            local fullscreen = JNI:callBooleanMethod(
+            local fullscreen = JNI:callIntMethod(
                 android.app.activity.clazz,
                 "isFullscreen",
-                "()Z"
+                "()I"
             )
             android.LOGI("is fullscreen =", fullscreen)
-            return fullscreen
+            return fullscreen == 1
         end)
     end
     android.setFullscreen = function(fullscreen)
@@ -1395,13 +1395,13 @@ local function run(android_app_state)
 
     android.isWifiEnabled = function()
         return JNI:context(android.app.activity.vm, function(JNI)
-            local isWifiEnabled = JNI:callBooleanMethod(
+            local isWifiEnabled = JNI:callIntMethod(
                 android.app.activity.clazz,
                 "isWifiEnabled",
-                "()Z"
+                "()I"
             )
             android.LOGI("is WifiEnabled =", isWifiEnabled)
-            return isWifiEnabled
+            return isWifiEnabled == 1
         end)
     end
 
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1360,7 +1360,7 @@ local function run(android_app_state)
     end
     android.isFullscreen = function()
         return JNI:context(android.app.activity.vm, function(JNI)
-            local fullscreen = JNI:callIntMethod(
+            local fullscreen = JNI:callBooleanMethod(
                 android.app.activity.clazz,
                 "isFullscreen",
                 "()Z"
@@ -1395,7 +1395,7 @@ local function run(android_app_state)
 
     android.isWifiEnabled = function()
         return JNI:context(android.app.activity.vm, function(JNI)
-            local isWifiEnabled = JNI:callIntMethod(
+            local isWifiEnabled = JNI:callBooleanMethod(
                 android.app.activity.clazz,
                 "isWifiEnabled",
                 "()Z"
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1363,7 +1363,7 @@ local function run(android_app_state)
             local fullscreen = JNI:callBooleanMethod(
                 android.app.activity.clazz,
                 "isFullscreen",
-                "()Z"
+                "()C"
             )
             android.LOGI("is fullscreen =", fullscreen)
             return fullscreen
@@ -1398,7 +1398,7 @@ local function run(android_app_state)
             local isWifiEnabled = JNI:callBooleanMethod(
                 android.app.activity.clazz,
                 "isWifiEnabled",
-                "()Z"
+                "()C"
             )
             android.LOGI("is WifiEnabled =", isWifiEnabled)
             return isWifiEnabled
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1047,7 +1047,7 @@ function JNI:callBooleanMethod(object, method, signature, ...)
     local clazz = self.env[0].GetObjectClass(self.env, object)
     local methodID = self.env[0].GetMethodID(self.env, clazz, method, signature)
     self.env[0].DeleteLocalRef(self.env, clazz)
-    return self.env[0].callBooleanMethod(self.env, object, methodID, ...)
+    return self.env[0].CallBooleanMethod(self.env, object, methodID, ...)
 end
 
 function JNI:callStaticBooleanMethod(class, method, signature, ...)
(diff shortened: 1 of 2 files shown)


mwoz123:
That code change didn't help:
>10-23 21:07:19.608  4368  4386 E luajit-launcher: Failed to run script: [string "android.lua"]:1050: 'struct JNINativeInterface' has no member named 'callBooleanMethod'
10-23 21:07:19.625  4368  4386 V threaded_app: android_app_destroy!

 "android.lua" line 1050: 
>return self.env[0].callBooleanMethod(self.env, object, methodID, ...)

Is that the reason why nobody used callBooleanMethod before? :(

Does any one know how what's wrong here?

assets/android.lua
```diff
@@ -1360,7 +1360,7 @@ local function run(android_app_state)
     end
     android.isFullscreen = function()
         return JNI:context(android.app.activity.vm, function(JNI)
-            local fullscreen = JNI:callIntMethod(
+            local fullscreen = JNI:callBooleanMethod(
                 android.app.activity.clazz,
                 "isFullscreen",
                 "()Z"
```
Frenzie: Shouldn't this be something like B then?

mwoz123: You mean "()B" instead "()Z" ?
It's completely unreadable for me where it came from and what it does mean :(

If it caught my change the result is the same:
10-23 21:31:42.706  4577  4593 E luajit-launcher: Failed to run script: [string "android.lua"]:1050: 'struct JNINativeInterface' has no member named 'callBooleanMethod'
10-23 21:31:42.712  4577  4593 V threaded_app: android_app_destroy!

mwoz123: I must have overlooked doc for that. If there's such?

Frenzie: Pardon, B is byte and Z is boolean.

mwoz123: Just "Z" or "()Z" or "Z()"?

Frenzie: Oh, please link directly to the line. It's clearer. :-) You should change it to a capital C.

Frenzie: On line 1050.

Frenzie: > Just "Z" or "()Z" or "Z()"?

No it's good.

Frenzie: Sorry for my earlier brevity; I was on my phone in a slight hurry but I wanted to give you something useful to work with in case you were still looking at it.

Anyway, it should be fine if you fix it to an uppercase C on the line you mentioned.

https://github.com/koreader/android-luajit-launcher/blob/bd885618d058e96dbef35e744cbff1b6c2c4048b/assets/android.lua#L1050

Frenzie:
I apologize if I was confusing, but it's this line:
https://github.com/koreader/android-luajit-launcher/blob/bd885618d058e96dbef35e744cbff1b6c2c4048b/assets/android.lua#L1050


That has to be changed to 
`return self.env[0].CallBooleanMethod(self.env, object, methodID, ...)`

The other line is the line *I* mentioned, not the one you mentioned. ;-)

mwoz123:
done.

mwoz123:
I did a quick test and something still seems not ok.

mwoz123:
same result for isFullscreen:
>I/KOReader( 3137): is fullscreen =
I/KOReader( 3137): DEBUG screen_fullscreen_menu_table.lua:  Fullscreen swiching to:  false

Frenzie:
I suspect you'll need to cast this `jboolean` object into a C bool.

Presumably it'll give you these values
https://github.com/koreader/android-luajit-launcher/blob/bd885618d058e96dbef35e744cbff1b6c2c4048b/assets/android.lua#L772-L773

There's a to_string.
https://github.com/koreader/android-luajit-launcher/blob/bd885618d058e96dbef35e744cbff1b6c2c4048b/assets/android.lua#L1099-L1104

We'd probably have to add to_bool.

So you could fix it up properly if you want to. (I don't particularly feel like looking at that for the moment.)

But I think it's fine if you just keep the typo correction on CallBooleanMethod for the moment and use the int workaround for now. It doesn't look too horrible. :-)

mwoz123:
Changed to `int` + follow up changes in `MainActivity.java` but unfortunately on my x86 emulator result was exactly the same. 

This can mean there's an issue in Java implementation, but it was just done excalty as on some recipes on internet  (or issue with emulator?). Maybe it's again Java code has to be wrapped with something?

Any way if it's java issue then we would be able to uncommit ab96312

Frenzie:
Oh, I thought you meant you'd already tested the int method. For wifi I would only trust results from a real device, although I suppose in the emulator it should still always return false instead of nothing.

I don't have a working Android knowledge or anything. I just have a knowledge of various things that go into Android (such as XML and Java). But from browsing around the docs real quick it could be that you should use different methods for Android <4.4 and Android 4.4+.

### API 16
https://developer.android.com/training/system-ui/immersive.html
https://developer.android.com/reference/android/view/View.html#SYSTEM_UI_FLAG_FULLSCREEN

### API 1

Basically like this except get instead of set

https://github.com/koreader/android-luajit-launcher/blob/ab963127c3f7955e3918e647a991df7adf442ea0/src/org/koreader/launcher/MainActivity.java#L45-L61
https://developer.android.com/reference/android/view/WindowManager.LayoutParams.html#FLAG_FULLSCREEN

mwoz123:
I checked it using 5.1 emulator which is api 22 so should work. 

I checked setFullscreen and setWifiEnabled, it was working fine. Must have overlooked the isWifiEnabled and isFulscreen. Sorry. 

BTW as far as I remember the setFullscreenLayout was after my full screen commit. 

It seems to be the same issue. In android.lua or MainActivity.java. 
From Java perspective it seems ok unless it again needs some extra magic around that code but it's now constructed exactly like isCharging so should work.(unless isCharging is not working correctly, but is at least returns a value in logs).

 At least I can see now no difference between isCharging is Fullscreen and isWiFienabled...

Frenzie:
> I checked it using 5.1 emulator which is api 22 so should work.

Okay, but we want to keep things working on API 9 and up unless there's a really good reason not to. Doesn't your own Android reader come with Android API <16? ;-)

> BTW as far as I remember the setFullscreenLayout was after my full screen commit.

It was structured differently but it was there from the first time the Java layer was added in cb6deb890ce26e0cc53b0ce0521d45974d0261fa (with a slight mistake fixed in 655bf228e247cba543c15814cb6c06e9e1d022d4).

src/org/koreader/launcher/MainActivity.java
```diff
@@ -158,17 +158,17 @@ public void run() {
         });
     }
 
-    public boolean isFullscreen() {
-    	return (getWindow().getAttributes().flags & WindowManager.LayoutParams.FLAG_FULLSCREEN) != 0;
+    public int isFullscreen() {
+    	return ((getWindow().getAttributes().flags & WindowManager.LayoutParams.FLAG_FULLSCREEN) != 0) ? 1: 0;
```
Frenzie: Should this be `&&`?

mwoz123: & is a Java low level (bit)  comparison as far as I remember from couple years back, Java trainings. It's almost not used in plain Java. Never had a chance to see it in real code. 
But did seen it in android. 

Also if this was the issue it should return at least false. 

I'll double check if it's not the case..

Frenzie: I see, it's bitwise AND to extract the flag. Makes sense, although it seems more like a C kind of thing than something you'd expect to see in Java.

mwoz123:
This was a problem with x86 emulator.

Works on ARM.  Please merge. I'll push new PR for API 11 Fullscreen

Frenzie:
It's fine if it doesn't work but does it also not crash? :-P

mwoz123:
emulator x86 = doesn't crash, doesn't work :(

arm = doesn't crash, works;)

Frenzie:
I meant API-wise, sorry for the confusion.

PR #74, opened 2017-11-05
Android clipboard interface

mwoz123:
Android part of : https://github.com/koreader/koreader/issues/3441

Not tested. Please verify.

@Frenzie FYI

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1381,6 +1381,40 @@ local function run(android_app_state)
         end)
     end
 
+    android.getClipboardText = function()
+        return JNI:context(android.app.activity.vm, function(JNI)
+            JNI:callObjectMethod(
+                android.app.activity.clazz,
+                "getClipboardText",
+                "()Ljava/lang/String;"
+            )
+        end)
+    end
+
+    android.hasClipboardText = function()
+        return JNI:context(android.app.activity.vm, function(JNI)
+            JNI:callIntMethod(
+                android.app.activity.clazz,
+                "hasClipboardTextIntResultWrapper",
+                "()I"
+            )
+        end)
+    end
+
+    android.setClipboardText = function(text)
+        android.LOGI("setting clipboard text to: ", text)
+        local clipboard_text = JNI.env[0].NewStringUTF(JNI.env, text)
+        JNI:context(android.app.activity.vm, function(JNI)
+            JNI:callVoidMethod(
+                android.app.activity.clazz,
+                "setClipboardText",
+                "(Ljava/lang/String;)V",
+                clipboard_text
+            )
+            JNI.env[0].DeleteLocalRef(JNI.env, clipboard_text)
+        end)
+    end
+
     android.setKeepScreenOn = function(keepOn)
         android.LOGI("setting KeepScreenOn to ", keepOn)
         JNI:context(android.app.activity.vm, function(JNI)
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1383,26 +1383,31 @@ local function run(android_app_state)
 
     android.getClipboardText = function()
         return JNI:context(android.app.activity.vm, function(JNI)
-            JNI:callObjectMethod(
+            local text = JNI:callObjectMethod(
                 android.app.activity.clazz,
                 "getClipboardText",
                 "()Ljava/lang/String;"
             )
+            text = JNI:to_string(text)
+            android.LOGI("clipboard text copied: " .. text)
+            return text
         end)
     end
 
     android.hasClipboardText = function()
         return JNI:context(android.app.activity.vm, function(JNI)
-            JNI:callIntMethod(
+            local hasClipboardText = JNI:callIntMethod(
                 android.app.activity.clazz,
                 "hasClipboardTextIntResultWrapper",
                 "()I"
             )
+            android.LOGI("Has text in clipboard: ", hasClipboardText)
+            return hasClipboardText == 1
         end)
     end
 
     android.setClipboardText = function(text)
-        android.LOGI("setting clipboard text to: ", text)
+        android.LOGI("setting clipboard text to: " .. text)
         JNI:context(android.app.activity.vm, function(JNI)
             local clipboard_text = JNI.env[0].NewStringUTF(JNI.env, text)
             JNI:callVoidMethod(
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1376,7 +1376,7 @@ local function run(android_app_state)
                 android.app.activity.clazz,
                 "setFullscreen",
                 "(Z)V",
-                ffi.new('bool', fullscreen)
+                ffi.new("bool", fullscreen)
             )
         end)
     end
@@ -1422,7 +1422,7 @@ local function run(android_app_state)
                 android.app.activity.clazz,
                 "setKeepScreenOn",
                 "(Z)V",
-                ffi.new('bool', keepOn)
+                ffi.new("bool", keepOn)
             )
         end)
     end
@@ -1446,7 +1446,7 @@ local function run(android_app_state)
                 android.app.activity.clazz,
                 "setWifiEnabled",
                 "(Z)V",
-                ffi.new('bool', wifiEnabled)
+                ffi.new("bool", wifiEnabled)
             )
         end)
     end
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1403,8 +1403,8 @@ local function run(android_app_state)
 
     android.setClipboardText = function(text)
         android.LOGI("setting clipboard text to: ", text)
-        local clipboard_text = JNI.env[0].NewStringUTF(JNI.env, text)
         JNI:context(android.app.activity.vm, function(JNI)
+            local clipboard_text = JNI.env[0].NewStringUTF(JNI.env, text)
             JNI:callVoidMethod(
                 android.app.activity.clazz,
                 "setClipboardText",
--- a/src/org/koreader/launcher/MainActivity.java
+++ b/src/org/koreader/launcher/MainActivity.java
@@ -232,20 +232,44 @@ public class MainActivity extends NativeActivity {
     }
 
     public String getClipboardText() {
-        if (hasClipboardText()) {
-            if (SDK_INT < android.os.Build.VERSION_CODES.HONEYCOMB) {
-                android.text.ClipboardManager clipboard = (android.text.ClipboardManager) getSystemService(CLIPBOARD_SERVICE);
-                return clipboard.getText().toString();
-            } else {
-                ClipboardManager clipboard = (ClipboardManager) getSystemService(Context.CLIPBOARD_SERVICE);
-                ClipData.Item item = clipboard.getPrimaryClip().getItemAt(0);
-                return item.getText().toString();
-            }
-        }
-        return "";
+      final Box<String> result = new Box<String>();
+      final CountDownLatch cd = new CountDownLatch(1);
+      runOnUiThread(new Runnable() {
+          @Override
+          public void run() {
+              try {
+                  if (hasClipboardText()) {
+                      if (SDK_INT < android.os.Build.VERSION_CODES.HONEYCOMB) {
+                          android.text.ClipboardManager clipboard = (android.text.ClipboardManager) getSystemService(CLIPBOARD_SERVICE);
+                          result.value = clipboard.getText().toString();
+                      } else {
+                          ClipboardManager clipboard = (ClipboardManager) getSystemService(Context.CLIPBOARD_SERVICE);
+                          ClipData.Item item = clipboard.getPrimaryClip().getItemAt(0);
+                          result.value = item.getText().toString();
+                      }
+                  }
+                  result.value = "";
+              } catch (Exception e) {
+                  Log.v(LOGGER_NAME, e.toString());
+                  result.value = "";
+              }
+              cd.countDown();
+          }
+      });
+      try {
+          cd.await();
+      } catch (InterruptedException ex) {
+          return "";
+      }
+
+      if (result.value == null) {
+          return "";
+      }
+
+      return result.value;
     }
 
-    public void setClipboardText(String text) {
+    public void setClipboardText(final String text) {
         runOnUiThread(new Runnable() {
             @Override
             public void run() {
@@ -276,7 +300,31 @@ public class MainActivity extends NativeActivity {
     }
 
     public int hasClipboardTextIntResultWrapper() {
-        return hasClipboardText() ? 1 : 0;
+      final Box<Integer> result = new Box<Integer>();
+      final CountDownLatch cd = new CountDownLatch(1);
+      runOnUiThread(new Runnable() {
+          @Override
+          public void run() {
+              try {
+                  result.value = hasClipboardText() ? 1 : 0;
+              } catch (Exception e) {
+                  Log.v(LOGGER_NAME, e.toString());
+                  result.value = new Integer(0);
+              }
+              cd.countDown();
+          }
+      });
+      try {
+          cd.await();
+      } catch (InterruptedException ex) {
+          return 0;
+      }
+
+      if (result.value == null) {
+          return 0;
+      }
+
+      return result.value.intValue();
     }
 
     public int getStatusBarHeight() {
--- a/src/org/koreader/launcher/MainActivity.java
+++ b/src/org/koreader/launcher/MainActivity.java
@@ -26,21 +26,16 @@ import java.util.concurrent.CountDownLatch;
 public class MainActivity extends NativeActivity {
 
     private final static int SDK_INT = android.os.Build.VERSION.SDK_INT;
-
-    private class Box<T> {
-        public T value;
-    }
+    private final static String LOGGER_NAME = "luajit-launcher";
+    private FramelessProgressDialog dialog;
 
     static {
         System.loadLibrary("luajit");
     }
 
-    private static String TAG = "luajit-launcher";
-    private FramelessProgressDialog dialog;
-
     public MainActivity() {
         super();
-        Log.v(TAG, "Creating luajit launcher main activity");
+        Log.v(LOGGER_NAME, "Creating luajit launcher main activity");
     }
 
     @Override
@@ -48,23 +43,7 @@ public class MainActivity extends NativeActivity {
         super.onCreate(savedInstanceState);
     }
 
-    private void setFullscreenLayout() {
-        if(SDK_INT >= 11 && SDK_INT < 16) {
-            getWindow().getDecorView().setSystemUiVisibility(View.STATUS_BAR_HIDDEN);
-        } else if (SDK_INT >= 16 && SDK_INT < 19) {
-            getWindow().getDecorView().setSystemUiVisibility(
-                    View.SYSTEM_UI_FLAG_FULLSCREEN | View.SYSTEM_UI_FLAG_LOW_PROFILE);
-        } else if (SDK_INT >= 19) {
-            getWindow().getDecorView().setSystemUiVisibility(
-                    View.SYSTEM_UI_FLAG_LAYOUT_STABLE |
-                            View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION |
-                            View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN |
-                            View.SYSTEM_UI_FLAG_HIDE_NAVIGATION |
-                            View.SYSTEM_UI_FLAG_FULLSCREEN |
-                            View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY);
-        }
-    }
-
+    @Override
     protected void onResume() {
         super.onResume();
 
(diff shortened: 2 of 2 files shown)


src/org/koreader/launcher/MainActivity.java
```diff
@@ -224,40 +225,83 @@ public void run() {
...
+            }
+        }
+        return "";
+    }
+
+    public boolean hasClipboardText() {
+        if (SDK_INT < android.os.Build.VERSION_CODES.HONEYCOMB) {
+            android.text.ClipboardManager clipboard = (android.text.ClipboardManager) getSystemService(CLIPBOARD_SERVICE);
+            return clipboard.hasText();
+        }else{
```
Frenzie: nitpick

src/org/koreader/launcher/MainActivity.java
```diff
@@ -224,40 +225,83 @@ public void run() {
         }
     }
 
+    public void setClipboardText(String text) {
+        if (SDK_INT < android.os.Build.VERSION_CODES.HONEYCOMB) {
```
Frenzie: Perhaps it's just me, but I find the reverse logic easier to read because then I don't have to make a mental translation.

if (SDK_INT >= HONEYCOMB)
//new stuff
else
//fallback old stuff

Frenzie:
I'm not sure when I'd be able to test. Think in the order of 16 hours from now at a minimum. ;-)

mwoz123:
>Don't we also need some kind of READ_CLIPBOARD permission?

Neither documentation nor stackoverflow.com posts says anything about new permissions :
https://developer.android.com/guide/topics/text/copy-paste.html

src/org/koreader/launcher/MainActivity.java
mwoz123: setFullscreenLayout() also starts from lowest version. 
It might be useful is they again change something in newer version. 
If you still want reverse you'd have to wait unit I can reach pc, can't do that from phone.

src/org/koreader/launcher/MainActivity.java
mwoz123: As above

mwoz123:
>I'm not sure when I'd be able to test. Think in the order of 16 hours from now at a minimum. ;-)

Not sure if we need tests (at least from Java , not Android developer perspective). 
They say Java tests should be Blackbox public API Tests for logic, and we don't have to much logic.

Anyway for mocking I recommend Mockito :
http://site.mockito.org
It's one of the simplest, and also powerful and nowadays one of the most popular.
(of course on top of Junit which seems to be a standard these days)

Frenzie:
> Not sure if we need tests

I was merely referring to the fact that these are 129 lines of "Not tested. Please verify."

Unit tests would be nice to have but that's an entirely different can of worms.

src/org/koreader/launcher/MainActivity.java
Frenzie: The nitpick is about the lack of spaces. ;-)

Frenzie:
With a quick test this doesn't seem to be working.
```
11-06 13:34:45.764  2747  2764 I KOReader: INFO  initializing for device Android
11-06 13:34:45.764  2747  2764 I KOReader: INFO  framebuffer resolution: {
11-06 13:34:45.764  2747  2764 I KOReader:     ["h"] = 1920,
11-06 13:34:45.764  2747  2764 I KOReader:     ["w"] = 1080
11-06 13:34:45.764  2747  2764 I KOReader: }
11-06 13:34:45.764  2747  2764 I KOReader: DEBUG ASDFASDFASDFSDFASDFSDFASDFSDF
--------- beginning of crash
11-06 13:34:45.764  2747  2764 E AndroidRuntime: FATAL EXCEPTION: Thread-143
11-06 13:34:45.764  2747  2764 E AndroidRuntime: Process: org.koreader.launcher, PID: 2747
11-06 13:34:45.764  2747  2764 E AndroidRuntime: java.lang.RuntimeException: Can't create handler inside thread that has not called Looper.prepare()
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at android.os.Handler.<init>(Handler.java:200)
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at android.os.Handler.<init>(Handler.java:114)
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at android.content.ClipboardManager$2.<init>(ClipboardManager.java:69)
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at android.content.ClipboardManager.<init>(ClipboardManager.java:69)
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at android.app.SystemServiceRegistry$10.createService(SystemServiceRegistry.java:213)
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at android.app.SystemServiceRegistry$10.createService(SystemServiceRegistry.java:212)
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at android.app.SystemServiceRegistry$CachedServiceFetcher.getService(SystemServiceRegistry.java:768)
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at android.app.SystemServiceRegistry.getSystemService(SystemServiceRegistry.java:721)
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at android.app.ContextImpl.getSystemService(ContextImpl.java:1365)
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at android.view.ContextThemeWrapper.getSystemService(ContextThemeWrapper.java:123)
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at android.app.Activity.getSystemService(Activity.java:5263)
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at org.koreader.launcher.MainActivity.hasClipboardText(MainActivity.java:258)
11-06 13:34:45.764  2747  2764 E AndroidRuntime: 	at org.koreader.launcher.MainActivity.hasClipboardTextIntResultWrapper(MainActivity.java:264)
11-06 13:34:45.765  1624  1998 W ActivityManager:   Force finishing activity org.koreader.launcher/.MainActivity
```

Tested using this after Generic.init in android/device.lua using the same procedure .
```
logger.dbg("ASDFASDFASDFSDFASDFSDFASDFSDF")
logger.dbg(android.hasClipboardText())
logger.dbg(android.getClipboardText())
logger.dbg(android.setClipboardText("blabla"))
logger.dbg(android.getClipboardText())
```

There are also some issues with the other methods.

```
11-06 13:44:06.886  2848  2865 I KOReader: setting clipboard text to: 
11-06 13:44:06.886  2848  2865 E luajit-launcher: Failed to run script: [string "android.lua"]:1406: attempt to index field 'env' (a nil value)
11-06 13:44:06.886  2848  2865 V threaded_app: android_app_destroy!
```

Probably nothing that isn't easy to solve, but please do try to perform some basic testing.

src/org/koreader/launcher/MainActivity.java
mwoz123: Come on, you worse than lua check:p
This is default auto code formatting build in Android Studio. 
I can't imagine myself manually changing that

mwoz123:
>java.lang.RuntimeException: Can't create handler inside thread that has not called Looper.prepare()

Initially I thought it's something with `android.lua` as I'm child in dark in that file.
But stackoverflow says it has to be run on ui thread (whatever it means:P)
https://stackoverflow.com/a/9815528

I'll fix that later as I was about to also refactor this code to lambdas.

mwoz123:
btw can you check `android.lua` ?

Frenzie:
Except for that JNI.env thing in setClipboardText (see bottom of previous reply) it's presumably okay. As an aside, you should always try if the simplest method (i.e., a plain Lua string) works before doing FFI type casting or something even more complex. For a `const char*` a Lua string will do, for a `char*` you'd need to get the length.

assets/android.lua
```diff
@@ -1381,6 +1381,40 @@ local function run(android_app_state)
...
+                android.app.activity.clazz,
+                "hasClipboardTextIntResultWrapper",
+                "()I"
+            )
+        end)
+    end
+
+    android.setClipboardText = function(text)
+        android.LOGI("setting clipboard text to: ", text)
+        local clipboard_text = JNI.env[0].NewStringUTF(JNI.env, text)
```
Frenzie: There's no JNI.env out here because it's something passed into the function below. This needs to be inside the anonymous function.

src/org/koreader/launcher/MainActivity.java
mwoz123: Do you still want that in reverse order?

Frenzie: What's your own opinion on it from a legibility perspective?

Frenzie:
Since you're fixing up style issues I noticed a couple of 'bool' instead of "bool" (unless there's a functional difference).

src/org/koreader/launcher/MainActivity.java
mwoz123: I prefer the current way.

mwoz123:
you mean double quotes ? in MainActivity? in java single means char , double quote string.

Frenzie:
No, they were added by you in android.lua and I failed to spot them at the time. They're inconsistent style-wise but shouldn't make any functional difference.

mwoz123:
can you retest?

Frenzie:
It's a pity that you stuck the UI thread business in the same commit as that lambda stuff because that way I can't easily revert it for testing (although really I'm not entirely sure what I'm doing testing it in the first place >_>).

```
-compile:
    [javac] Compiling 1 source file to /home/frans/src/kobo/koreader/platform/android/luajit-launcher/bin/classes
    [javac] warning: [options] source value 1.5 is obsolete and will be removed in a future release
    [javac] warning: [options] target value 1.5 is obsolete and will be removed in a future release
    [javac] warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
    [javac] /home/frans/src/kobo/koreader/platform/android/luajit-launcher/src/org/koreader/launcher/MainActivity.java:51: error: lambda expressions are not supported in -source 1.5
    [javac]         handler.postDelayed(() -> setFullscreenLayout(), 500);
    [javac]                                ^
    [javac]   (use -source 8 or higher to enable lambda expressions)
    [javac] 1 error
    [javac] 3 warnings
```

Changing this might be okay since the "1.5 is obsolete" message also occurs in the nightly build, but it should be separated out so that it can be easily reverted if necessary.

src/org/koreader/launcher/MainActivity.java
```diff
@@ -21,92 +24,79 @@
...
                     View.SYSTEM_UI_FLAG_FULLSCREEN | View.SYSTEM_UI_FLAG_LOW_PROFILE);
         } else if (SDK_INT >= 19) {
             getWindow().getDecorView().setSystemUiVisibility(
                     View.SYSTEM_UI_FLAG_LAYOUT_STABLE |
-                    View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION |
-                    View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN |
-                    View.SYSTEM_UI_FLAG_HIDE_NAVIGATION |
-                    View.SYSTEM_UI_FLAG_FULLSCREEN |
-                    View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY);
+                            View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION |
```
Frenzie: What's the deal with the indentation here?

src/org/koreader/launcher/MainActivity.java
```diff
@@ -21,92 +24,79 @@
...
-                                                       Settings.System.SCREEN_BRIGHTNESS));
-                } catch (Exception e) {
-                    Log.v(TAG, e.toString());
-                    result.value = new Integer(0);
-                }
-                cd.countDown();
+        runOnUiThread(() -> {
+            try {
+                result.value = new Integer(Settings.System.getInt(
+                        getContentResolver(),
```
Frenzie: Also odd indentation. It's not aligned to the right in which case it would make sense to override regular one indentation.

src/org/koreader/launcher/MainActivity.java
```diff
@@ -122,75 +112,63 @@ public void run() {
...
+        return (level * 100) / scale;
     }
 
     public int isCharging() {
         Intent intent = this.registerReceiver(null,
                 new IntentFilter(Intent.ACTION_BATTERY_CHANGED));
         int plugged = intent.getIntExtra(BatteryManager.EXTRA_PLUGGED, -1);
         return plugged == BatteryManager.BATTERY_PLUGGED_AC
-            || plugged == BatteryManager.BATTERY_PLUGGED_USB ? 1 : 0;
+                || plugged == BatteryManager.BATTERY_PLUGGED_USB ? 1 : 0;
```
Frenzie: Same, should be one indentation.

src/org/koreader/launcher/MainActivity.java
```diff
@@ -201,63 +179,106 @@ public void run() {
...
+                } else {
+                    attrs.flags &= ~WindowManager.LayoutParams.FLAG_FULLSCREEN;
                 }
-                cd.countDown();
+                getWindow().setAttributes(attrs);
+            } catch (Exception e) {
+                Log.v(LOGGER_NAME, e.toString());
             }
+            cd.countDown();
+
```
Frenzie: Stray newline.

src/org/koreader/launcher/MainActivity.java
```diff
@@ -201,63 +179,106 @@ public void run() {
...
     }
 
+    public void setClipboardText(String text) {
+        runOnUiThread(() -> {
+            if (SDK_INT < android.os.Build.VERSION_CODES.HONEYCOMB) {
+                android.text.ClipboardManager clipboard = (android.text.ClipboardManager) getSystemService(CLIPBOARD_SERVICE);
+                clipboard.setText(text);
+            } else {
+                ClipboardManager clipboard = (ClipboardManager) getSystemService(CLIPBOARD_SERVICE);
+                ClipData clip = ClipData.newPlainText("Koreader_clipboard", text);
```
Frenzie: KOReader ;-)

src/org/koreader/launcher/MainActivity.java
mwoz123: Android Studio default code formating. There must be a max line lenght

mwoz123:
removed refactor to lambdas 161ef14472d090728e221deb47e11798fd8ae0e8 bundled with setters on UI. 
now only setters on UI thread in PR. 

Please retest.

Frenzie:
```
-compile:
    [javac] Compiling 2 source files to /home/frans/src/kobo/koreader/platform/android/luajit-launcher/bin/classes
    [javac] warning: [options] source value 1.5 is obsolete and will be removed in a future release
    [javac] warning: [options] target value 1.5 is obsolete and will be removed in a future release
    [javac] warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
    [javac] /home/frans/src/kobo/koreader/platform/android/luajit-launcher/src/org/koreader/launcher/MainActivity.java:255: error: local variable text is accessed from within inner class; needs to be declared final
    [javac]                         clipboard.setText(text);
    [javac]                                           ^
    [javac] /home/frans/src/kobo/koreader/platform/android/luajit-launcher/src/org/koreader/launcher/MainActivity.java:258: error: local variable text is accessed from within inner class; needs to be declared final
    [javac]                         ClipData clip = ClipData.newPlainText("KOReader_clipboard", text);
    [javac]                                                                                     ^
    [javac] Note: /home/frans/src/kobo/koreader/platform/android/luajit-launcher/src/org/koreader/launcher/MainActivity.java uses or overrides a deprecated API.
    [javac] Note: Recompile with -Xlint:deprecation for details.
    [javac] 2 errors
    [javac] 3 warnings
```

Fixing that problem with `public void setClipboardText(final String text) {` compiles but you didn't put `hasClipboardText` and `getClipboardText` on the UI thread so they still crash with the Looper error.

`setClipboardText` doesn't crash but due to the above I can't tell if it's working properly.

mwoz123:
Hmm Maybe fullscreen also doens't work because it's not on ui thread?
I'm not sure if we need all of that wrapping, but lets first see if it works correctly..

Frenzie:
Works without crashing now but I'm not getting any output. (Also does it really need to be quite that complex? lol)

I think fullscreen used to work; just something about the changed properties.

Sorry, out of time for today. Tomorrow's another day. ;-)

chrox:
@mwoz123 Is the Android building env not working for you? Or do you have no Android devices so that you cannot test before submitting a PR? Just curious. :)

mwoz123:
Problems with build. Probably Android Studio (Tools) upgrade did a mess. I need to re-setup.

Don't have a clue what might be wrong then. I suspect android.lua.

src/org/koreader/launcher/MainActivity.java
```diff
@@ -224,40 +231,140 @@ public void run() {
...
+                      if (SDK_INT < android.os.Build.VERSION_CODES.HONEYCOMB) {
+                          android.text.ClipboardManager clipboard = (android.text.ClipboardManager) getSystemService(CLIPBOARD_SERVICE);
+                          result.value = clipboard.getText().toString();
+                      } else {
+                          ClipboardManager clipboard = (ClipboardManager) getSystemService(Context.CLIPBOARD_SERVICE);
+                          ClipData.Item item = clipboard.getPrimaryClip().getItemAt(0);
+                          result.value = item.getText().toString();
+                      }
+                  }
+                  result.value = "";
```
mwoz123: What's that doing here?:O 
Isn't it destroying all the data we had in clipboard? 
I must have overlooked that.

I might not have soon time to take a look at that, but if someone would be able to check it can remove it?

Frenzie: Should be able to give it a quick try in an hour. Certainly sounds that way. :lol:

src/org/koreader/launcher/MainActivity.java
```diff
@@ -1,65 +1,50 @@
...
-import android.net.wifi.WifiManager;
+import android.view.WindowManager;
 
 import java.util.concurrent.CountDownLatch;
 
 public class MainActivity extends NativeActivity {
-    private class Box<T> {
-        public T value;
-    }
+
```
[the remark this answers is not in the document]
mwoz123: It's common it java for better readability.

src/org/koreader/launcher/MainActivity.java
mwoz123: removed. Can be tested.

assets/android.lua
```diff
@@ -1376,19 +1376,53 @@ local function run(android_app_state)
...
     end
 
+    android.getClipboardText = function()
+        return JNI:context(android.app.activity.vm, function(JNI)
+            JNI:callObjectMethod(
+                android.app.activity.clazz,
+                "getClipboardText",
+                "()Ljava/lang/String;"
+            )
+        end)
```
Frenzie: Needs JNI:to_string, see android.externalStorage for the best way to structure it

assets/android.lua
```diff
@@ -1376,19 +1376,53 @@ local function run(android_app_state)
...
+    end
+
+    android.hasClipboardText = function()
+        return JNI:context(android.app.activity.vm, function(JNI)
+            JNI:callIntMethod(
+                android.app.activity.clazz,
+                "hasClipboardTextIntResultWrapper",
+                "()I"
+            )
+        end)
```
Frenzie: See android.isFullscreen; you should change it to a boolean here by returning `== 1` but in either way you're not returning anything.

Frenzie:
Still not working unfortunately, but I see now what should be the problem. Not sure why I didn't spot it when it just jumped out at me today.

mwoz123:
> Not sure why I didn't spot it when it just jumped out at me today.

I had exactly the same with Java code today :D

assets/android.lua
```diff
@@ -1376,19 +1376,53 @@ local function run(android_app_state)
...
+    end
+
+    android.setClipboardText = function(text)
+        android.LOGI("setting clipboard text to: ", text)
```
Frenzie: Btw, this should be concatenated (`..` instead of `,`).

I suppose we could also change the `android.LOGI(message)` function to `android.LOGI(...)` with some kind of params loop so it'd take any number of arguments but that doesn't really seem worthwhile atm.

assets/android.lua
Frenzie: To be abundantly clear, in both cases you're effectively saying something like this:

```lua
return function()
  boop()
  return nil --default return value for any function
  -- we're dealing with a JNI datatype here, which complicates matters slightly
  -- so we don't even get a proper Lua nil but that's not really important
  -- to the principle
end
```

To do anything at all you'd need at least

```lua
return function()
  return boop()
end
```

We were probably misled by the fact that it's nested, so it looks like something is being returned when it actually isn't.

mwoz123:
Changed. Can you retest?

Frenzie:
Excellent, it's working now. :-)

mwoz123:
hooooooray!:)

PR #75, opened 2017-11-06
[w.i.p.] gradle migration init commit

mwoz123:
Init commit. 
It seems to be more complicated that I thought, especially because of C++ code.
**Can someone take over this?**

brief import summary:
https://github.com/mwoz123/android-luajit-launcher/blob/27b94ce45883dcdca8209e0c678dce9e5447bc0e/import-summary.txt
+ changed Java version to 8,
+ added task 'debug'
+ linked ndk code 
+ conf set to native SDK

Currently in PC fails with error:
> External Native Build Issues
Build command failed.
Error while executing process /home/marcin/Android/Sdk/ndk-bundle/ndk-build with arguments {NDK_PROJECT_PATH=null APP_BUILD_SCRIPT=/home/marcin/StudioProjects/android-luajit-launcher/app/src/main/jni/Android.mk NDK_APPLICATION_MK=/home/marcin/StudioProjects/android-luajit-launcher/app/src/main/jni/Application.mk APP_ABI=armeabi NDK_ALL_ABIS=armeabi NDK_DEBUG=0 APP_PLATFORM=android-9 NDK_OUT=/home/marcin/StudioProjects/android-luajit-launcher/app/build/intermediates/ndkBuild/release/obj NDK_LIBS_OUT=/home/marcin/StudioProjects/android-luajit-launcher/app/build/intermediates/ndkBuild/release/lib APP_SHORT_COMMANDS=false LOCAL_SHORT_COMMANDS=false -B -n}
Android NDK: android-9 is unsupported. Using minimum supported version android-14.    
Android NDK: ERROR:/home/marcin/StudioProjects/android-luajit-launcher/app/src/main/jni/Android.mk:luajit-prebuilt: LOCAL_SRC_FILES points to a missing file    
Android NDK: Check that /home/marcin/StudioProjects/android-luajit-launcher/app/src/main/jni/luajit-build/armeabi/lib/libluajit-5.1.a exists  or that its path is correct   
Build command failed.

MainActivity & android.lua based from:
https://github.com/koreader/android-luajit-launcher/pull/74


Frenzie:
Thank you for looking into this! Unfortunately I don't really have any time to look into this myself, plus I don't know anything about the subject at this point in time.

As to the error, it looks like you didn't change the Makefile, so what exactly is this `build command` you're using? You'd either have to replace the Makefile with something equivalent or change the Makefile where relevant:

https://github.com/koreader/android-luajit-launcher/blob/55418468e026da7ced95e3266093eadc9c22a21c/Makefile#L20-L37

Note that luajit builds *in source*, which is pretty much the only thing I consider truly problematic (i.e., a bug). Can Gradle take care of that somehow or should we do something with, e.g., CMake for that? In any case, although I consider that to be the greatest priority it wouldn't necessarily need to be part of this PR.

gradlew
```diff
@@ -0,0 +1,160 @@
...
+##
+##  Gradle start up script for UN*X
+##
+##############################################################################
+
+# Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
+DEFAULT_JVM_OPTS=""
+
+APP_NAME="Gradle"
+APP_BASE_NAME=`basename "$0"`
```
Frenzie: I assume you didn't write this script? Because it has some issues. For instance, `$()` should be used for executing commands.

import-summary.txt
```diff
@@ -0,0 +1,44 @@
+ECLIPSE ANDROID PROJECT IMPORT SUMMARY
```
Frenzie: This file should be part of the commit.

build.gradle
```diff
@@ -0,0 +1,21 @@
...
+    }
+    dependencies {
+        classpath 'com.android.tools.build:gradle:3.0.0'
+    }
+}
+
+allprojects {
+    repositories {
+        jcenter()
+        maven { url 'https://maven.google.com' }
```
Frenzie: What's Maven doing here? Is that really necessary?

mk-luajit.sh
```diff
@@ -25,6 +31,7 @@ function check_NDK() {
         exit 1
     fi
 }
+cd $NDK_PROJECT_PATH
```
Frenzie: You should always `cd $NDK_PROJECT_PATH || exit` in case `cd` fails.

But in our case, our Bash style is:

```bash
pushd $NDK_PROJECT_PATH && {
    # commands
} && popd || exit

Frenzie:
The correct way to move a Git submodule is `git mv old new`.

build.gradle
mwoz123: Right

mwoz123: Not needed

gradlew
mwoz123: Not mine. Auto generated

import-summary.txt
mwoz123: Just to let know for someone who 'd take over task

gradlew
Frenzie: Just out of curiosity, what tool generates legacy code from decades ago? See http://mywiki.wooledge.org/BashFAQ/082

NB This is not a Bashism.

import-summary.txt
Frenzie: Fair enough.

Frenzie:
Like I said I don't think I have time to look into this right now, but could you post a few relevant documentation links to get started if possible so someone else wouldn't have to start entirely from scratch?

gradlew
mwoz123: Android Studio;)

mwoz123:
Links:
https://guides.gradle.org/creating-new-gradle-builds/
and couple more links from the same source:
https://gradle.org/guides/#getting-started

but I think it's now more about tunning up paths, etc than gradle.

I finished on this problem:
>[armeabi-v7a] Gdbserver      : [arm-linux-androideabi] libs/armeabi-v7a/gdbserver
[armeabi-v7a] Gdbsetup       : libs/armeabi-v7a/gdb.setup
[armeabi-v7a] Compile thumb  : luajit <= android-main.c
app/src/main/jni/android-main.c:28:10: fatal error: 'luajit-2.1/lua.h' file not found
#include "luajit-2.1/lua.h"
         ^~~~~~~~~~~~~~~~~~
1 error generated.                                                                                                         
make[1]: *** [app/src/main/obj/local/armeabi-v7a/objs-debug/luajit/android-main.o] Error 1

Requirements:
- add to path `android` executable (so it can run `android update project --path . -t android-19`, btw `android` is deprecated and we should take a look if it can be replaced with `tools/bin/sdkmanager or tools/bin/avdmanager`)
- export $NDK=<your path to NDK>
- if you're using NDK 15 you have to workaroud:
https://github.com/koreader/android-luajit-launcher/pull/75/files#diff-b1d063e5e7900470cafeb272d808d0fcR9
- export PATH=$PATH:$NDK (just in case - it's already in code https://github.com/koreader/android-luajit-launcher/pull/75/files#diff-6cc961e6325fa0390c03b831fa6c5595R24)

mk-luajit.sh
mwoz123: good to know. thanks

Frenzie:
That's not a problem in principle (e.g., https://github.com/koreader/koreader-base/pull/510#discussion_r139290793) but in practice that either means supporting both (meh, complexity, etc.) or waiting a while longer. Not sure the time is right yet.

> export $NDK=//your path to NDK

We already do that automatically.

pazos:
@Frenzie @mwoz123: I'm going to close this PR. It has code already in master (clipboard stuff, for example). 

A month ago I never touched Android, but today I'm starting to understand why it sucks for these kind of projects. Sadly importing an old project in Android Studio will not work for us and will break our setup in many many ways.

The only solution is writting gradle tasks by hand. First a task to build luajit, then a task to build the activity and finally port our current ant tasks.

Frenzie:
> I'm starting to understand why it sucks <del>for these kind of projects</del>.

Ftfy. :angel:

PR #76, opened 2017-11-08
Change screenOrientation to sensorPortrait

poire-z:
Allows for flipping device orientation: holding it upside-down may be more confortable than the default portrait orientation the maker decided for.

References:
https://stackoverflow.com/questions/582185/android-disable-landscape-mode/582585#582585
https://stackoverflow.com/questions/9801580/screen-orientation-and-values-in-manifest-xml/9801725#9801725
https://developer.android.com/guide/topics/manifest/activity-element.html#screen

Got hold of a cheap tablet on which I tried koreader: koreader was stuck to a single portrait orientation (all the other apps were following 180° device rotation). This default orientation was not confortable (thin side, volume keys, warm area are then at the bottom right corner in my right hand).
I first solved it by using in patch.lua some tricks with:
```
self.bb:setRotation(2) -- in overidden framebuffer_android:init()
Device.input:registerEventAdjustHook(Device.input.adjustTouchMirrorX, Device.screen:getWidth())
Device.input:registerEventAdjustHook(Device.input.adjustTouchMirrorY, Device.screen:getHeight())
```
which was fine enough (except that swiping from the top shows the bottom menu, and swiping from the bottom shows the status bar, inverted :)

This is probably the most proper way to solve my problem (tested with apktool d / edit / apktoll b / sign apk)
However, it may not please trapezists, fakirs and gymnasts if they read while practicing :|

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -15,7 +15,7 @@
         android:largeHeap="true" android:vmSafeMode="true">
         <activity android:name="org.koreader.launcher.MainActivity"
                 android:label="@string/app_name"
-                android:screenOrientation="portrait"
+                android:screenOrientation="sensorPortrait"
                 android:launchMode="singleInstance"
                 android:theme="@style/Fullscreen">
             <meta-data android:name="android.app.lib_name"


Frenzie:
What happens in KOReader when you flip the device around 180°?

poire-z:
Koreader just doesn't notice anything: it is rotated by android so it follows gravity and stays readable if I stay vertical :)

poire-z:
Also, I have to say that I didn't notice the problem on my phone, which prevents "reverse portrait". But this tablet does not.
(I haven't tested this modification on my phone, but I suppose that if the phone does not allow reverse portrait, this patch will not make it work).

Frenzie:
I just realized I should've been slightly more specific in what I was asking so I'm not entirely sure if my question was answered or not. I didn't mean what happens when you launch it this way or the other, but rather what happens if you rotate it while it's running.

poire-z:
That's how I understood it, and what I answered :)
While it is running, if I flip upside down my tablet, koreader is rotated by Android to follow gravity (and it does nothing by itself, as it probably doesn't get any signal about this rotation happening).

Tested this morning's nightly that it works as intended on my tablet, and that it it works as before/have no side effect on my phone (where "reverse portrait" is probably not allowed/disabled : so, no matter how much i flip/turn my phone, koreader does not move and stays in the normal portraite mode).

Have you noticed some problem with this commit?

Frenzie:
I haven't been able to test it yet. :-)

PR #77, opened 2017-11-25
Fixed for darwin build

chrox:


mk-luajit.sh
```diff
@@ -6,11 +6,12 @@
 # NDKABI=19  # Android 4.4+
 #NDKABI=${NDKABI:-14} # Android 4.0+
 NDKABI=${NDKABI:-9} # Android 2.3+
-BUILD_ARCH=linux-$(uname -m)
 DEST=$(cd "$(dirname "$0")" && pwd)/jni/luajit-build/$1
+# might be linux-x86_64 or darwin-x86-64
+HOST_ARCH=*
```
Frenzie: This should be quoted (`HOST_ARCH="*"`).

chrox: True. Shellcheck is handy for that. :)

PR #89, opened 2019-01-06
Add logic to launch new files from intent + misc fixes

pazos:
Related to https://github.com/koreader/koreader/issues/4441

1. Add a function to get file path from intent (working from 2.3 to 7.1) - **moved from llapp_main.lua**
2. override current intent with newIntent
3. move a bunch of strings to debug logger
4. fix logging of boolean values

A function to get the file path from 8.0 onwards is missing (intent scheme changed from file to content). I'm not interested in doing this myself since JNI stuff is like Chinese to me.


Frenzie:
Thanks!

PR #90, opened 2019-01-09
Add wakelock helper,dont report isCharging when battery is full and fix some warnings.

pazos:
I was fixing warnings when I found https://github.com/koreader/koreader/issues/4430.

This PR adds an automatic wakelock handler during luajit-launcher app lifecycle and provides functions to modify its standard behaviour from lua.

The standard behaviour (acquire wakelock when the application is resumed, and thus focused, and release the wakelock when the application is paused will grant that both cpu, touchscreen and display are ready as long as the application runs on the foreground).

~~I tried to move some methods outside MainActivity to fix the mess we have here. For both powermanager, wifimanager and batterymanager we just need to get the MainActivity context but don't need to run on Ui thread.~~

Also added a minimum gradle wrapper for ant, which works fine, and can be extended later.


pazos:
Pinging @chrox for review.

src/org/koreader/launcher/WirelessHelper.java
```diff
@@ -0,0 +1,22 @@
...
+    public static void setState(Context context, final boolean enabled) {
+        getWifiManager(context).setWifiEnabled(enabled);
+    }
+
+    public static int getState(Context context) {
+        return getWifiManager(context).isWifiEnabled() ? 1 : 0;
+    }
+}
+
+
```
Frenzie: /whistles

(Yes, I nitpick.)

src/org/koreader/launcher/WirelessHelper.java
```diff
@@ -0,0 +1,22 @@
+package org.koreader.launcher;
+
+import android.content.Context;
+import android.net.wifi.WifiManager;
+
+/* helper class to get battery state */
```
Frenzie: /whistles

src/org/koreader/launcher/BatteryHelper.java
```diff
@@ -0,0 +1,35 @@
...
+
+        return (level * 100) / scale;
+    }
+
+    public static int isCharging() {
+        return isCharging;
+    }
+}
+
+
```
Frenzie: nit

pazos:
> Looks okay in principle; which versions did you test on? In principle we're going for API level 9 and up.
9 might mostly academic, but 14/15 compatibility is a must due to all the Android 4 ereaders.

Except for building with gradle, which is disabled, everything else is using old methods, so it should work like before.

Tested on emulator(4.2), and devices(7.1 and 8.1)

pazos:
> Alright, if you could also test Android 2.3 please do, if not let me know and I'll do it tonight. :-)

If you have an emulator please do it yourself. I'm fairly low on disk space on my Ubuntu VM.

Frenzie:
Never mind, we'll have to go without Android 2.3 testing. I forgot that the SD card on the Android 2.3 emulator is broken.

pazos:
> I forgot that the SD card on the Android 2.3 emulator is broken.

I don't understand. SD card shouldn't be needed for testing this PR. You can install the app with `adb install koreader.apk`

Frenzie:
Yes, you can test some things, but Android always has an "SD card" in, e.g., /mnt/sdcard. In newer Androids they call it "internal storage" for display purposes, while an optional real SD card will be mounted someplace like /storage/the-uuid-or-something.

tl;dr The Android 2.3 emulator is broken in such a way that many apps won't run properly, including KOReader. Anyway, the PR crashes where it's supposed to:

```
E/luajit-launcher( 1599): Failed to run script: frontend/cache.lua:46: cannot open /mnt/sdcard/koreader/cache/: Permission denied
```

Before #87 it would crash without ever getting around to loading the main program. ;-)

Of course theoretically I could still test pretty much everything with some simple code but, well, I don't really care. I'm 90% sure I'm the only person who even notices when things break on Android 2.3.

Some people claim you can fix it by setting the SD card to more than 512 MB in the settings, but I tried various values (including gigabytes) without success.

screenshot_2019-01-09_19-09-13

Frenzie:
Btw, since you're at it you could consider updating 7-zip to 18.06 which came out about a week ago. It claims yet another 3-10% speed increase.

Is the only real change in the "add support for wakelocks commit" or should I look at the whole thing again?

pazos:
> Btw, since you're at it you could consider updating 7-zip to 18.06 which came out about a week ago. It claims yet another 3-10% speed increase.

Sure

> Is the only real change in the "add support for wakelocks commit" or should I look at the whole thing again?

Yeah, is the only major change. The other little change is a slightly refactor of battery code to be able to report isCharging = false when the battery is full. It should hurt ;)

Frenzie:
Alright, thanks a lot!

Frenzie:
Glancing through there suggests r15c is required for that? We currently use r12b because r15c balks at API levels lower than 14, although in practice that doesn't seem to be a problem for running the program on Android 2.3.

Frenzie:
Btw, I did a quick experiment on GitLab with r15c and it's also failing. Unfortunately GitLab is stupid and they don't show you the part of the log with the actual failure, which is rather inconvenient.

PR #91, opened 2019-01-12
bump minimum target to API14, minimum NDK version to 15 and minimum SDK version to 19.

pazos:
It should be possible to build for API 11 (since both clipboard and screenSize work) changing to `<uses-sdk android:minSdkVersion="11" />.`

related to https://github.com/koreader/koreader/issues/4471

added a function to get sdk version from frontend.
setFullscreen should be now easier to understand.
slightly more verbose if requested.

#### For the future:

- handle configurationChanges
- provide a method to dump logs to sdcard: we need a permission for that. read logs. It is easy to implement. Just add the permission in AndroidManifest.xml and use `android.execute("logcat KOReader:V *:I -d /sdcard/koreader/koreader.log")

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,5 @@
-NDK_VER=$(shell grep -E 'NDKABI=[0-9]+' ./mk-luajit.sh | cut -d= -f2)
+# Supported Android ABIs: armeabi-v7a, x86 and x86_64
 
-# Android is kind of bizarre. Sometimes you need specific stuff.
-# Other times you need just `arm` or else…
-# For `x86` and `x86_64` it's all good.
 ifdef ANDROID_ARCH
 	ifeq ($(ANDROID_ARCH), arm)
 		ANDROID_FULL_ARCH?=armeabi-v7a
@@ -10,12 +7,12 @@ ifdef ANDROID_ARCH
 		ANDROID_FULL_ARCH?=$(ANDROID_ARCH)
 	endif
 endif
+
+# Default is build for arm
 ANDROID_FULL_ARCH?=armeabi-v7a
 
-# at least 16 is required to create a build with View.SYSTEM_UI_FLAG_FULLSCREEN
-# and View.SYSTEM_UI_FLAG_LOW_PROFILE
-# however, default to 19 because that's what the nightly build server uses
-NDKABI_MIN_16=$(shell [ ${NDKABI} -ge 16 ] && echo -n ${NDKABI} || echo -n 19)
+# Minimum SDK API is 19, required to use View.VERSION_CODES.KITKAT
+SDKAPI_MIN=$(shell [ ${NDKABI} -ge 19 ] && echo -n ${NDKABI} || echo -n 19)
 
 apk: local.properties project.properties
 	git submodule init
@@ -27,7 +24,7 @@ apk: local.properties project.properties
         #gradle debug
 
 local.properties project.properties:
-	android update project --path . -t android-$(NDKABI_MIN_16)
+	android update project --path . -t android-$(SDKAPI_MIN}
 
 clean:
 	-ndk-build clean
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1254,31 +1254,41 @@ local function run(android_app_state)
         end)
     end
 
+    android.getProduct = function()
+        return JNI:context(android.app.activity.vm, function(JNI)
+            local product = JNI:callObjectMethod(
+                android.app.activity.clazz,
+                "getProduct",
+                "()Ljava/lang/String;"
+            )
+            return JNI:to_string(product) or "unknown"
+        end)
+    end
+
     android.getScreenWidth = function()
         return JNI:context(android.app.activity.vm, function(JNI)
-            local width = JNI:callIntMethod(
+            return JNI:callIntMethod(
                 android.app.activity.clazz,
                 "getScreenWidth",
                 "()I"
             )
-            android.LOGV("get screen width  " .. width)
-            return width
         end)
     end
+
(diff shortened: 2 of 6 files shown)


assets/android.lua
```diff
@@ -1575,22 +1589,21 @@ local function run(android_app_state)
...
     os.execute = function(command) -- luacheck: ignore 122
         if command == nil then return -1 end
         local argv = {}
         command:gsub("([^ ]+)", function(arg) table.insert(argv, arg) end)
         return android.execute(unpack(argv))
     end
+
     android.LOGI("Application data directory "..android.dir)
     android.LOGI("Application library directory "..android.nativeLibraryDir)
-    android.LOGI("Screen size "..android.screen.width.."x"..android.screen.height)
```
Frenzie: How come you think this line is useless?

pazos: It is reported by KOReader

```logcat
I/KOReader( 3715):  initializing for device sdk_x86
I/KOReader( 3715):  framebuffer resolution: {
I/KOReader( 3715):     ["h"] = 800,
I/KOReader( 3715):     ["w"] = 480
I/KOReader( 3715): }
```

But, yeah screen.height and screen.width are required since framebuffer_android.lua relies on them.

BTW, i just found that `android.app.activity.sdkVersion` can be used instead of `android.getSdk()`

pazos:
er, I think I just pushed a new branch on the repo...

assets/android.lua
Frenzie: But that's much higher up. Basically I'm thinking a bit like the added logs in https://github.com/koreader/koreader-base/pull/788

It's just that every time something Android comes around it seems like there's too little in the way of logs. :-)

pazos: A proper log for KOReader on android would be `adb logcat KOReader:V *:I`. That means log verbose (and higher) logs for KOReader and info (and higher) logs for everything else, including luajit-launcher. The launcher throw errors if a lua file isn't well formatted or if it can't load assets, so it is almost impossible to have a sane luajit-launcher and don't reach device initialization.

I'm fine logging that on luajit-launcher too. Or logging that on https://github.com/koreader/koreader-base/blob/master/ffi/framebuffer_android.lua#L8, where screen.size is used for the first time.

Frenzie: I should probably fully explain my angle first, since I didn't express it very well (if at all) by talking about higher and lower level.

From KOReader's perspective it should be fine to remove, but keep in mind this isn't KOReader. It's a purposefully separate, generic way of using LuaJIT on Android. Obviously development is every so slightly KOReader-centric ;-) but this is the kind of line I'd probably feel grateful for not having to include myself even it wouldn't take very long and it seems slightly redundant.

In that sense it's a bit like @NiLuJe's KFMon or that [KPVBooklet](https://github.com/koreader/kpvbooklet) thing.

pazos: @Frenzie: yes, I understand your point, but as a **luajit launcher** it doesn't need to log all functions that are exposed to lua. We don't log if we have a clipboard or the wifi state at launch.

Frenzie: Those aren't comparable imo since they're much more tangential, but anyway, the latest commit looks nice.

src/org/koreader/launcher/MainActivity.java
```diff
@@ -349,37 +348,38 @@ public int getStatusBarHeight() {
...
-                size.set(display.getWidth(), display.getHeight());
-            }
-        } catch (Exception e) {
-            Log.v(LOGGER_NAME, e.toString());
+            // JellyBean 4.2 (API 17) and higher
+            DisplayMetrics metrics = new DisplayMetrics();
+            display.getRealMetrics(metrics);
+            size.set(metrics.widthPixels, metrics.heightPixels);
+        } catch (NoSuchMethodError e) {
+            // Honeycomb 3.0 (API 11) adn higher
```
Frenzie: Typo ;)
```suggestion
            // Honeycomb 3.0 (API 11) and higher
```

pazos:
@Frenzie: please review new commits. Nothing huge, just a few bits modified.

So, in short:
    - We support the following android ABIs: armeabi-v7a, x86 and x86-64
    - We require at least NDK version 15 to build native libraries.
    - We require at least SDK API version 19 to build Java code.
    - minimum target is API 14(ICS, 4.0), but things should work as is on API11 (Honeycomb 3.0), except that the new minSdktarget won't allow installation on devices w/ API < API 14.

Makefile
```diff
@@ -1,21 +1,18 @@
-NDK_VER=$(shell grep -E 'NDKABI=[0-9]+' ./mk-luajit.sh | cut -d= -f2)
+# Supported Android ABIs: armeabi-v7a, x86 and x86_64
 
-# Android is kind of bizarre. Sometimes you need specific stuff.
-# Other times you need just `arm` or else…
-# For `x86` and `x86_64` it's all good.
```
Frenzie: I wouldn't remove this comment explaining the mad code that follows.

pazos: Umm, It doesnt look mad to me. Android abis for arm are armeabi, armeabi-v7a... . Since we don't support armeabi and don't support aarch64 yet we just set armeabi-v7a when ANDROID_ARCH=arm.

And then build luajit with specified ANDROID_FULL_ARCH.

Frenzie: In isolation I suppose that's true, but I found/find not being able to just use, e.g., `armeabi-v7a` everywhere (i.e., having various parts of Android resolve it to just `arm` internally when appropriate) one of the many inconveniences of Android.

Command A goes this way, command B goes that way, if you're lucky command C follows at least a or B, etc.

PR #94, opened 2019-01-20
[wip] refactor java code + disable wakelocks by default

pazos:
#### DONE
- screen code goes to ScreenHelper
- power code goes to PowerHelper
- clipboard code goes to ClipboardHelper
- disable wakelocks by default
- logs on destroy from javaland
- removes the dumb logActivity function :D
- adds a couple of comments

#### TODO
- add timeout setter/getter to ScreenHelper
- explore if we can invalidate or postInvalidate current activity surface to force the epd driver to redraw the screen (for e-ink devices only)
- test the clipboard, yup.


pazos:
Uo! the diff looks really bad for MainActivity.java. If you look at the file itself it is easy to understand :+1:

Frenzie:
The only problem is that it's a bit hard to separate the changes from the copy/paste refactoring in the diff. Because except for wakelock that's basically all it is, right? :-)

pazos:
Yes, a bunch of copy&paste, the isWakelock=false  and a few other things named above.

I'm new to Java (and to Android), but it seems fairly easy to understand at a basic level. I went the C++ way of allocating new stuff on the constructor. Since in Java there is no destructor I just set each class object to null and rely on java GC to free unused allocations. The reference to these objects is hold during the entire life of the activity (from onCreate to onDestroy).

I'm not sure if this is a good implementation but I kept my phone running KOReader the last two days and I didn't found any memory leaks or invalid references or ...

The trick to runOnUiThread from a non Activity class came from https://stackoverflow.com/a/13086422

pazos:
@Frenzie: I will move the wakelock stuff to another PR to be merged ASAP and leave this PR opened for a while.

pazos:
> Btw, the refactoring and trying things with the drawing surface can also be separate perfectly fine? Imo they should ideally be separate commits anyway, regardless whether that's as separate PRs or separate commits within a PR. ;-)

Yeah, sure. But I was wrong, it is possible to force e-ink refreshes without touching surfaces at all. OFC invalidate doesn't work but as long as we use the epd directly we can get the root view with `getWindow().getDecorView().findViewById(android.R.id.content)`.

Now I'm not sure how to implement things on KOReader, Here is an example video (remove zip extension), I hooked the eink full refresh in setWakeLock to be able to quickly test. 

[eink.webm.zip](https://github.com/koreader/android-luajit-launcher/files/2793922/eink.webm.zip)

Just want the full refresh each X pages for now, no fancy stuff.

Any ideas @Frenzie @NiLuJe

pazos:
@Frenzie: thanks for the hints, I will check later today. BTW, I just found awesome EPD device abstraction for rockchip to use instead of onyx sdks (leaner and easily extensible). Since the code doesn't have an explicit license I've opened an issue on https://github.com/unwmun/refreshU/issues/1 and will wait for the author before porting/studying the code :+1:

pazos:
Did a few tests and they worked on various epd modes, but now I'm trying to follow your suggestion:

> It's probably easiest to simply rename that last one to something like function framebuffer:_updateSurface()

```diff
--- a/ffi/framebuffer_android.lua
+++ b/ffi/framebuffer_android.lua
@@ -18,7 +18,7 @@ function framebuffer:init()
     framebuffer.parent.init(self)
 end
 
-function framebuffer:refreshFullImp()
+function framebuffer:_updateWindow()
     if android.app.window == nil then
         android.LOGW("cannot blit: no window")
         return
@@ -60,4 +60,43 @@ function framebuffer:refreshFullImp()
     android.lib.ANativeWindow_unlockAndPost(android.app.window);
 end
 
+-- eink update modes
+local EPD_FULL, EPD_PART, EPD_A2, EPD_AUTO = 0, 1, 2, 3
+
+local function update_eink(mode)
+    if android.isEink() then
+        android.einkUpdate(mode)
+    end
+end
+
+function framebuffer:refreshFullImp()
+    framebuffer:_updateWindow()
+    update_eink(EPD_FULL)
+end
+
+function framebuffer:refreshPartialImp()
+    framebuffer:_updateWindow()
+    -- let the driver to do its job.
+end
+
+function framebuffer:refreshFlashPartialImp()
+    framebuffer:_updateWindow()
+    update_eink(EPD_PART)
+end
+
+function framebuffer:refreshUIImp()
+    framebuffer:_updateWindow()
+    -- let the driver to do its job.
+end
+
+function framebuffer:refreshFlashUIImp()
+    framebuffer:_updateWindow()
+    update_eink(EPD_AUTO)
+end
+
+function framebuffer:refreshFastImp()
+    framebuffer:_updateWindow()
+    update_eink(EPD_A2)
+end
+
 return require("ffi/framebuffer"):extend(framebuffer)
```

and got this error

 ```logcat
E/luajit-launcher( 1769): Failed to run script: ffi/framebuffer_android.lua:47: attempt to index local 'ext_bb' (a nil value)
```

Do you see something obvious?, thanks!

Frenzie:
Should probably be self:etc()?

PR #97, opened 2019-01-26
use java target/source 1.6

pazos:
It removes some warnings building against JDK8. In theory it supports building with JDK11/10 too, but I didn't test it.

Related to https://github.com/koreader/koreader/issues/4511#issue-403240699 and https://github.com/koreader/koreader/issues/4511#issuecomment-457757843

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -8,6 +8,8 @@
 
     <property name="java.target" value="1.6" />
     <property name="java.source" value="1.6" />
+    <property name="java.compilerargs" value="-Xlint:deprecation" />
+
 
     <!-- The ant.properties file can be created by you. It is only edited by the
          'android' tool to add properties to it.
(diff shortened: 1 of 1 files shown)


build.xml
```diff
@@ -6,6 +6,9 @@
          Version Control Systems. -->
     <property file="local.properties" />
 
+    <property name="java.target" value="1.6" />
+    <property name="java.source" value="1.6" />
```
Frenzie: So when they say "source option 6" they mean you should set it using `1.6`?

pazos: yep, all java versions are 1.x, so source 6 means 1.6.

Default values used by android are in ${SDK}/tools/ant/build.xml. 

I will add verbose messages on deprecated APIs too.

PR #98, opened 2019-01-30
add a eink-test activity

pazos:
this isn't built as part of android-luajit-launcher. Is a different apk which comes handy to test current
implemented EPD controllers. It provides info about device and lets the user to quickly test if the device
can be added to KOReader.

It is safe to run this activity on devices without e-ink displays, `adb logcat eink-test:V *:S` reports:

```adb logcat
01-30 16:20:32.932 20165 20165 I eink-test: fake update
01-30 16:20:33.885 20165 20165 I eink-test: rockchip forced update
01-30 16:20:33.886 20165 20165 E eink-test: java.lang.NullPointerException: enumType == null
01-30 16:22:41.853 20590 20590 I eink-test: fake update
01-30 16:22:42.643 20590 20590 I eink-test: rockchip normal update
01-30 16:22:42.646 20590 20590 E eink-test: java.lang.NullPointerException: enumType == null
```

EPD Controller abstract classes are duplicated, but I didn't find a way to build them without relying to hackish tricks, like making a symbolic link

@Frenzie: can this apk (12kb) be released as just another assets on the next release?. A brief description would be nice, and maybe you or somebody who undertand what is this and can explain to others in english easily can write one. :+1:


Frenzie:
> hackish tricks, like making a symbolic link

I don't think that's too hackish. It probably wouldn't work on Windows though. :-)

eink-test/src/org/koreader/einkTest/MainActivity.java
```diff
@@ -0,0 +1,99 @@
...
+        rk30xx_normal_button = (Button) findViewById(R.id.rockchipNormalButton);
+        rk30xx_forced_button = (Button) findViewById(R.id.rockchipForcedButton);
+
+        overview.setText("Manufacturer: " + MANUFACTURER);
+        overview.append("\n Brand: " + BRAND);
+        overview.append("\n Model: " + MODEL);
+        overview.append("\n Product: " + PRODUCT);
+        overview.append("\n Hardware: " + HARDWARE);
+
+        fakeDesc.setText("This button does nothing! is it just an example of an empty action. ");
```
Frenzie:
```suggestion
        fakeDesc.setText("This button does nothing! It's just an example of an empty action. ");
```

eink-test/src/org/koreader/einkTest/MainActivity.java
```diff
@@ -0,0 +1,99 @@
...
+        fakeDesc.append("If you don't see any difference between fake and real tests, your device ");
+        fakeDesc.append("is not supported by any implemented epd controller");
```
Frenzie:
```suggestion
        fakeDesc.append("is not supported by any implemented epd controller.");
```

PR #99, opened 2019-01-31
Release builds

pazos:
Related to https://github.com/koreader/koreader/issues/2040

It removes the debuggable property on release builds, among other little things.


Makefile
```diff
@@ -14,23 +14,36 @@ ANDROID_FULL_ARCH?=armeabi-v7a
...
+	# update local.properties and project.properties with sdk/ndk paths for current user
+	# this works with sdk tools <= 25.3.0
+	android update project --path . -t android-$(SDKAPI_MIN)
+
+	# update sources
 	git submodule init
 	git submodule sync
 	git submodule update
+
+build-native:
```
Frenzie: Should this perhaps also run something like `rm NativeActivity*`?

Makefile
```diff
@@ -14,23 +14,36 @@ ANDROID_FULL_ARCH?=armeabi-v7a
 # Minimum SDK API is 19, required to use View.VERSION_CODES.KITKAT
 SDKAPI_MIN=$(shell [ ${NDKABI} -ge 19 ] && echo -n ${NDKABI} || echo -n 19)
 
-apk: local.properties project.properties
+update:
+	# update local.properties and project.properties with sdk/ndk paths for current user
+	# this works with sdk tools <= 25.3.0
```
Frenzie: I was going to comment on this line but I see you already fixed it. ;-)

Makefile
```diff
@@ -14,23 +14,36 @@ ANDROID_FULL_ARCH?=armeabi-v7a
...
+
+build-native:
+	# build luajit, lzma and native activity for desired arch (armeabi-v7a, x86, x86_64)
 	./mk-luajit.sh $(ANDROID_FULL_ARCH)
 	ndk-build ANDROID_FULL_ARCH=$(ANDROID_FULL_ARCH)
+
+debug: update build-native
 	ant debug
-        #gradle debug
+	cp -pv bin/NativeActivity-debug.apk NativeActivity.apk
```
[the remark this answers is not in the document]
pazos: I'm not sure about what "duplicating this in base" means. I was thinking about just mv NativeActivity.apk to koreader-abi-version.apk. But yeah, I'm open to whatever thing you see as the better option .

Makefile
pazos: I would call `make clean` as the first step in the build, then build koreader, copy assets to assets/module and build the native activity. Calling make clean will take care of removing old stuff, like the apk, and is needed if you want to build for two or more archs on the same machine.

Makefile
pazos: I'm not good with symbols :P

Makefile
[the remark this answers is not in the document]
Frenzie: As written here, [this line](https://github.com/koreader/koreader/blob/4874751020968e62bd9506859a0ae9c539111481/Makefile#L369) would lose the `-debug`.

The alternative would be to put the logic of `-debug` or not in koreader-front, which could potentially be more likely to fail (as a feature, since you wouldn't accidentally stat the previous build). My view on these lines as written is basically that it hides what's going on in `luajit-launcher/bin/` and you have to read the Makefile to understand what happened, which is something I'm never a big fan of.

Makefile
Frenzie: I meant that in conjunction with [that comment](https://github.com/koreader/android-luajit-launcher/pull/99#pullrequestreview-198590584), in the sense of preventing accidentally copying a previous build from the other alternative. So what I actually meant was just `NativeActivity.apk`.

Makefile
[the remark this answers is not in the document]
pazos: Please rework as you like. I found the messages quite verbose:

release:
```bash
BUILD SUCCESSFUL
Total time: 1 second
cp -pv bin/NativeActivity-release-unsigned.apk NativeActivity.apk
'bin/NativeActivity-release-unsigned.apk' -> 'NativeActivity.apk'
application was built, type: release (unsigned)
You'll need to sign this application to be able to install it
```

debug
```bash
BUILD SUCCESSFUL
Total time: 1 second
cp -pv bin/NativeActivity-debug.apk NativeActivity.apk
'bin/NativeActivity-debug.apk' -> 'NativeActivity.apk'
application was built, type: debug (signed)
```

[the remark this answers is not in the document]
Frenzie: Fair enough. I suppose we can just leave it as is here, but let's just wait with merging until that `apk` target can be deleted.  It should be easy enough to adjust in front combined with a luajit-launcher bump, but I don't have time tonight and I should spend the remaining couple of hours of the afternoon differently. :-)

Frenzie:
Hm, this is weird, I swear this worked locally. >_>

https://gitlab.com/koreader/nightly-builds/-/jobs/154362968

```
BUILD SUCCESSFUL
Total time: 8 seconds
cp -pv bin/NativeActivity-release-unsigned.apk NativeActivity.apk
'bin/NativeActivity-release-unsigned.apk' -> 'NativeActivity.apk'
application was built, type: release (unsigned)
You'll need to sign this application to be able to install it
make[2]: Leaving directory `/builds/koreader/nightly-builds/koreader/platform/android/luajit-launcher'
cp platform/android/luajit-launcher/bin/NativeActivity.apk \
		koreader-android-arm-linux-androideabi-v2019.01.1-40-gb68cdc2_2019-01-31.apk
cp: cannot stat 'platform/android/luajit-launcher/bin/NativeActivity.apk': No such file or directory
make[1]: *** [androidupdate] Error 1
make[1]: Leaving directory `/builds/koreader/nightly-builds/koreader'
Makefile:445: recipe for target 'update' failed
make: *** [update] Error 2
```

(It's expected to fail to sign due to me not having updated the nightswatcher Docker image yet but I must've overlooked something else.)

Anyway, I'll be out tonight, so I'll fix it tomorrow. :-)

PR #101, opened 2019-01-31
suppress 'safe' deprecations:

pazos:
STATUSBAR_HIDDEN is used on API<16 only
WAKELOCKS are all deprecated except PARTIAL, as the time of writting they're available on lastest Android version (9).

### DIFF
--- a/src/org/koreader/launcher/MainActivity.java
+++ b/src/org/koreader/launcher/MainActivity.java
@@ -107,6 +107,7 @@ public class MainActivity extends NativeActivity {
         Log.v(LOGGER_NAME, String.format("App %s", state));
     }
 
+    @SuppressWarnings("deprecation")
     private void setFullscreenLayout() {
         if (SDK_INT < Build.VERSION_CODES.JELLY_BEAN) {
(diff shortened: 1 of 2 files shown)


Frenzie:
You mean they'll be dropped in the future?

pazos:
> You mean they'll be dropped in the future?

I don't think so, but they mark them as deprecated to force people to use KEEP_SCREEN_ON instead. The only one which is not deprecated is PARTIAL, which is used to wake the device once it is sleeping but has nothing to do with keeping the screen on (ie: device can dim, turn off the screen, but the application holds the right to wake the device whenever it wants). It is used for alarms, notifications and so on.

BTW, humblebundle has a nice bundle (including Android Cookbook) for less than 1€.

Frenzie:
> BTW, humblebundle has a nice bundle (including Android Cookbook) for less than 1€.

Thanks, apparently I decided to pick it up for $1. Apparently I now have 384 ebooks total. x.x

```
document.querySelectorAll('.subproduct-selector').length
384
```

(The vast majority for $1, of course.)

PR #103, opened 2019-02-01
ant misc

pazos:
move ant build properties to ant.properties
get android target api from project.properties
adapt Makefile & build.xml


Frenzie:
> get android target api from project.properties

`$NDKABI` is dynamic on purpose. That's a feature, so that you can easily test, say, `NDKABI=26 ANDROID_ARCH=x86 ./kodev run android`.

The whole ensuring a minimum nonsense is just because the program as a whole can run on 9 (or now 14) but the APK won't compile unless the number is at least 19.

Also I just realized you removed the comment warning that 19 is specifically used by the build Docker image, which imo is a bad idea because in the future one might just update it to a higher value without realizing the Docker image would also have to be updated.

pazos:
> $NDKABI is dynamic on purpose. That's a feature, so that you can easily test, say, NDKABI=26 ANDROID_ARCH=x86 ./kodev run android.

Ok, If is useful as is then I'ld suggest to change the name (because is not related to NDK and not related to ABI), and do not track local.properties because it will be generated each time `android update --path . -t android-xx`  is executed.

> The whole ensuring a minimum nonsense is just because the program as a whole can run on 9 (or now 14) but the APK won't compile unless the number is at least 19.

I think this can be done with @TargetApi(19) on the relevant parts of the code (I think we just need this for SetFullscreenLayout but I need to check).

> Also I just realized you removed the comment warning that 19 is specifically used by the build Docker image, which imo is a bad idea because in the future one might just update it to a higher value without realizing the Docker image would also have to be updated.

Yeah, good point!

Frenzie:
I definitely think it's very useful — I didn't make it work that way did for nothing! ;-)

The variable name was introduced by @houqp in https://github.com/koreader/koreader-base/pull/459 and https://github.com/koreader/android-luajit-launcher/pull/45 I'm not really sure what the story behind the variable name is. You can change it to NDK_API_LEVEL (or something similar) but be sure to catch 'em all. ;-)

> I think this can be done with @TargetApi(19) on the relevant parts of the code (I think we just need this for SetFullscreenLayout but I need to check).

No, that itself requires API level 16. Otherwise I would've done that.

Frenzie:
PS I made some possibly confusing edits to my post here, in case you only read the e-mail version.

Frenzie:
> do not track local.properties because it will be generated each time android update --path . -t android-xx is executed.

I'm not really sure why it's tracked at all tbh, which strikes me as unnecessary at best. I figured I might be missing something. Android's biggest problem in that regard is that I just don't think it's much *fun*, even though I have nothing against Java per se. So I might eye a file like that suspiciously, but if it ain't broke…

pazos:
@Frenzie: thanks for all the info.

about the android file mess:

1. common

- local.properties: path to sdk & ndk. We just use the sdk since we use a Makefile to build native libraries instead of an ant task. *do not track*
- project.properties: here we can add project specific tasks instead of using the Makefile, or configure how to use proguard. We don't add any task, so it is only used for the android target. Google recommends to put this in git but in our case it doesn't apply, we keep the Makefile on git instead.
- proguard-project.txt: how to optimize the release builds, currently not used and probably never needed for this program.

2. ant specific
- and.properties: properties only used to build with ant. Like javac flags, source and dest dirs, and signing.
- build.xml: this overrides specific tasks used in google ant build.xml. It loads stuff from the other files, including sdk/tools/ant/build.xml.

So, in the short term we can avoid tracking project.properties if we specify a target when updating the project (as we are doing). The file will be generated by the android tool, like local.properties and proguard-project.txt are.

In the long term we might want to migrate the Makefile to specific gradle tasks (since gradle doesn't play nice with current build structure).

Frenzie:
You confused me a little earlier because we already don't track that file regardless?

pazos:
yeah, sorry, I was talking about project.properties, not local.properties :p

pazos:
> Automatically generated, changes will be erased… but must be checked in? Maybe it makes sense, but it sounds contradictory.

Mmm, the android tool will parse the file and update target accordly. New lines (including comments) are preserved. If the file isn't found it will be recreated by `android update --path . -t android-xx`.  If no target is specified then the android tool will rely on projecf.properties for setting the target and will fail if the file is not present.

> The variable name was introduced by @houqp in koreader/koreader-base#459 and #45. I'm not really sure what the story behind the variable name is. You can change it to NDK_API_LEVEL (or something similar) but be sure to catch 'em all. ;-)

Yeah NDKABI makes sense on base to set sysroot and to be able to build C code. Which does not makes sense to me is use NDKABI as the same as minimum target API.

```
MINIMUM_SDK?=19

ifdef TARGET_API
    SDK=$(shell TARGET_API > MINIMUM_SDK && echo TARGET_API || echo MINIMUM_SDK)
endif

SDK?=$(MINIMUM_SDK)
```

I will update this PR with a few changes.

Frenzie:
> Yeah NDKABI makes sense on base to set sysroot and to be able to build C code. Which does not makes sense to me is use NDKABI as the same as minimum target API.

I don't understand what you mean. The NDKABI variable is the target API. That makes no more or less sense in base than it does here. There's a different platform/sysroot dir for every target level. The ABI wasn't made dynamic until I called it `ANDROID_ARCH`. Perhaps I should've called it ABI and renamed the other one.

> MINIMUM_SDK?=19

That's a misleading name. We can (probably) build on SDK 11-15, but not on 19.

pazos:
Mmm, I don't get it. NDK stands for native development kit and is related to build c/c++ with an android compatible toolchain.

SDK target is related to whatever java classes are part of a specific api. We can have a target of SDK 23 and still build an app for API14, just need to make sure to not use Methods that were not introduced in this API. This can be controled at runtime catching NoSuchMethodError exceptions, like we do on https://github.com/koreader/android-luajit-launcher/blob/master/src/org/koreader/launcher/MainActivity.java#L374

The only real reason for bumping minimum sdk api to 19 is to be able to use flags (Namely View flags  added in API16 and code versions to differenciate between pre19 and 19). It should be posible to replace those flags for the integers and build our java classes with a sdk14.

Frenzie:
> NDK stands for native development kit and is related to build c/c++ with an android compatible toolchain.

Which includes building against specific libraries and headers, all of which are part of platform/API X. ANativeWindow is a platform 9 feature.

screenshot_2019-02-02_16-09-41

We don't do anything that isn't supported by NDK API level 9, but what we do would also be impossible without it. You can match NDK supported API level 9 more or less freely with whichever SDK API level larger than 9 you want, of course.

> The only real reason for bumping minimum sdk api to 19 is to be able to use flags (Namely View flags added in API16 and code versions to differenciate between pre19 and 19). It should be posible to replace those flags for the integers and build our java classes with a sdk14.

Note that it used to just say 19 hardcoded. I wanted it dynamic, and that's probably when I found there to be a problem with <16. Using 14 (or 9) means putting in some legwork to change the Docker image, which for whatever reason is at 19.

It's not that much effort; I can do it if you think it's worthwhile. But in that case I'll definitely have to double test #93 as well.

pazos:
> Note that it used to just say 19 hardcoded. I wanted it dynamic, and that's probably when I found there to be a problem with <16. Using 14 (or 9) means putting in some legwork to change the Docker image, which for whatever reason is at 19.

> It's not that much effort; I can do it if you think it's worthwhile. But in that case I'll definitely have to double test #93 as well.

Nah, I modified the PR. I'm fine using NDKABI as a way to modify target API, it is just that I didn't understand why it was used.

PR #104, opened 2019-02-02
Refactor

pazos:
Basically the same as #94, but slightly better, and dvivided in specific commits.

- screen helper for screen functions
- power helper for power functions (battery, wifi and wakelocks)
- clipboard for clipboard stuff

- added a few comments on MainActivity, related to activity lifecycle and to public functions exposed to lua.
- set the screen layout before calling the superclass, so the handler shouldn't be needed.
- removed the dumb logActivity function which I added a few weeks ago.


Frenzie:
Should be good. But let's wait with merging (or at least bumping in front) until I upload a release tomorrow just in case.

pazos:
Sure, I'm going to add a function to get Network information (ssid and ip) since route isn't available on android. Does it make sense to include wireless on power, like it is, or shall I add a new class for that (WifiHelper, for example)?

Frenzie:
As a user it makes sense to me. From a code organization perspective I'm less sure but it sounds like it should be just a few lines?

pazos:
@Frenzie: yep, just two more lines.

But I moved network functions to NetworkHelper, just in case somebody wants to write the full network manager, like in other platforms.

In the process I managed to break my git history, so I will push the whole refactor in one big commit  (including the function to get the network info) and the other little tweaks (set layout before calling the superclass) as a separate commit. Things worked fine here, but you'll probably want to test the build yourself before bumping frontend.

src/org/koreader/launcher/MainActivity.java
```diff
@@ -205,6 +207,18 @@ public int isWifiEnabled() {
...
+    public String getNetworkInfo() {
+        WifiInfo wi = getWifiManager().getConnectionInfo();
+        int ip = wi.getIpAddress();
+        String ip_address;
+        if (ip > 0) {
+            ip_address = Formatter.formatIpAddress(ip);
+        } else {
+            ip_address = String.valueOf(ip);
+        }
+        return String.format("ssid: %s, ip: %s", wi.getSSID(), ip_address);
```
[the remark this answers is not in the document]
pazos: I will need some help formatting the string in lua, but yeah, we can return a string with a lot of info: https://developer.android.com/reference/android/net/wifi/WifiInfo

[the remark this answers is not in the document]
Frenzie: string.format in Lua is exactly the same thing. ;-)

assets/android.lua
```diff
@@ -1495,6 +1495,17 @@ local function run(android_app_state)
...
     end
 
+    android.getNetworkInfo = function()
+        return JNI:context(android.app.activity.vm, function(JNI)
+            local network_info = JNI:callObjectMethod(
+                android.app.activity.clazz,
+                "getNetworkInfo",
+                "()Ljava/lang/String;"
+            )
+            return JNI:to_string(network_info)
```
Frenzie: We could return multiple values here (e.g., SSID and IP) or perhaps a table with named indices for clarity.

pazos: Sure, just need help to split the Java String returned by getNetworkInfo into the values we want to return to lua.

Frenzie: Oh, if you want to use a string since it's just a one-off then just toss in some commas or semicolons as separators? I'm not entirely sure I follow. :-)

Simple example:
http://www.lua.org/manual/5.1/manual.html#pdf-string.gmatch

Complex example:
https://github.com/koreader/koreader/blob/3f415604b9b1ada6fcb478c5c961c08b469049e9/frontend/util.lua#L43-L68

pazos:
@Frenzie: done. I don't know exactly why but loosing/getting focus and menu switching is now faster on old emulator/phones(4.0.4/6.0), while I didn't notice any improvements in recent phones (8.0)

I want to fix the intent stuff in Android 8+ but I think I'll wait for this one.

Frenzie:
@pazos Alright, could you bump luajit-launcher as part of #4550 and merge when ready? I don't have the right git tools at hand right now but I have no further comments.

poire-z:
(Not sure it's related to this PR - I haven't upgraded since Feb 1th - but as I see screen layout in the comments:)

With today's nightly, on my Android 6.0 Phone, the bottom android software buttons (back, home, applist) are no more hidden, and impossible to make disappear - and they are drawn over koreader, hiding the footer and the last line(s) of book text.

poire-z:
Update to the above observation:
When I start koreader, that bottom buttons bar stick above the footer.
If i switch back to home or another app, and then get back to a (still alive) koreader, it's all good: that bottom bar is not displayed, but I can bring it up (translucide) with a swipe from bottom, and it disappears as soon as I interact with koreader.
Exit koreader (with the back button), restart it from home, and the bottom android bar is fully opaque and stick.

Frenzie:
Let's see, the stable doesn't display right either until you switch from and to it, but the behavior is different.

In the stable version you get a black bar at the bottom until you switch away from and back to KOReader.
In current master you get the buttons on top of KOReader until you switch away from and back to KOReader.

What happens if you play with the fullscreen switch? (It occurs to me that my setting may not be vanilla, but it was set to on.)

poire-z:
(Haven't tried current stable - and dunno if reinstalling - now needing removal and fresh install - will trash the koreader/ data dir, I hope not - but I guess you're not asking me to try again stable, and that you just did and gave what you observed :)
But I observe what you observe with current master.

I don't have a fullscreen checkbox in the menus anymore - I think I saw it was removed some weeks ago.

Frenzie:
Right, I guess I was looking at v2019.1.1 when I asked that. (Which basically seemed to work the same as v2019.2.)

The data dir is in /sdcard-or-something-like-that/koreader, so unlike most Android apps there's never any harm in uninstalling KOReader.

poire-z:
Confirming the issue with Android6 from 5 posts above is solved in this morning nightly :) Thanks!

PR #108, opened 2019-02-03
[chore, CI] Add shellcheck

Frenzie:
To prevent things like this:

```
In platform/android/luajit-launcher/eink-test/copyEPDControllers.sh line 10:
for EPD in "${ROCKCHIP_EPD}"; do
           ^-- SC2066: Since you double quoted this, it will not word split, and the loop will only run once.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,4 +14,4 @@ addons:
 # check for XML syntax errors
 script:
   - find . -type f -iname '*.xml' -exec xmllint {} + >/dev/null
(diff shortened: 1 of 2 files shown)


Frenzie:
PS The full shellcheck/shfmt code could be copied from https://github.com/koreader/koreader/blob/f2a9ed0b79b783d709c7b0c55abe165f12782bef/.ci/helper_shellchecks.sh but since we shouldn't use shellscripts in here much anyway, front will catch less serious coding style-related issues without requiring updating things in more places.

PR #109, opened 2019-02-03
exit after crashing, to avoid unknown states

pazos:
Related to https://github.com/koreader/koreader/issues/4543

Exit android_main if the application fails trying to load assets, libraries or trying to run a lua file.

This makes possible to start the activity again when requested without having to "force kill" the application manually.

### DIFF
--- a/jni/android-main.c
+++ b/jni/android-main.c
@@ -117,7 +117,7 @@ void android_main(struct android_app* state) {
     lua_close(L);
 
 quit:
-    LOGE("Destroying Native Activity due to previous errors");
+    LOGE("Destroying NativeActivity due to previous errors");
     ANativeActivity_finish(state->activity);
     exit(1);
 }
--- a/jni/android-main.c
+++ b/jni/android-main.c
@@ -117,5 +117,7 @@ void android_main(struct android_app* state) {
     lua_close(L);
 
 quit:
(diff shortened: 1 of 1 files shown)


jni/android-main.c
```diff
@@ -117,5 +117,7 @@ void android_main(struct android_app* state) {
     lua_close(L);
 
 quit:
+    LOGE("Destroying Native Activity due to previous errors");
```
Frenzie: Seems a bit odd to split it up? :-)

```suggestion
    LOGE("Destroying NativeActivity due to previous errors");
```

jni/android-main.c
```diff
@@ -117,5 +117,7 @@ void android_main(struct android_app* state) {
...
     ANativeActivity_finish(state->activity);
+    exit(1);
```
Frenzie: Is this code path only reached on a bad exit?

pazos: Yes.
It isn't called when configuration changes, which leads to onDestroy->onCreate (which is reached when orientation changes between portrait and reverse portrait). 
It isn't called when invoking "finish" from lua either, via Device:exit() from KOReader POV.

Frenzie: Excellent, thanks!

PR #113, opened 2019-02-12
Layout fixes

pazos:
Should fix https://github.com/koreader/koreader/issues/4584

I based this on ndk samples (teapots). But the sample is broken (it sets a listener on visibility changes and even then it calls setImmersiveSticky).

This also removes one of the deprecations (if google's code isnt' totally borked). I tested this on emulator 9/4.2. Can't test in 4.0.3 because google efforts to make our life harder (support for sdcard is broken).

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -15,7 +15,7 @@
     <uses-permission android:name="android.permission.WAKE_LOCK"/>
     <application android:icon="@drawable/icon" android:label="@string/app_name"
         android:largeHeap="true" android:vmSafeMode="true">
-        <meta-data android:name="android.max_aspect" android:value="2.1" />
+        <meta-data android:name="android.max_aspect" android:value="3.1" />
         <activity android:name="org.koreader.launcher.MainActivity"
                 android:label="@string/app_name"
                 android:screenOrientation="sensorPortrait"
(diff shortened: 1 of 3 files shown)


Frenzie:
> Can't test in 4.0.3 because google efforts to make our life harder (support for sdcard is broken).

Ah, just like on 2.3.3? But at least on that one they have the excuse that they put a big fat "deprecated" on it… Not like it worked a few years ago, but whatever…

pazos:
> Ah, just like on 2.3.3? But at least on that one they have the excuse that they put a big fat "deprecated" on it… Not like it worked a few years ago, but whatever…

They broke support on new android-studio-all-in-one-sdk thingy. It still works with ooold sdk tools (the standalone avdmanager). 

@Frenzie: I'm not sure if it still makes sense to support [fullscreen toggling](https://github.com/koreader/android-luajit-launcher/blob/master/src/org/koreader/launcher/ScreenHelper.java#L95-L122). Which still works in API14-16, doesn't work at all in 16-18 and is totally unneeded on API19+.

Frenzie:
@pazos That might depend a bit on how it all works out on some of those Android ereaders I suppose.

pazos:
Ok, since I can't test it I will leave things as is. Last time I tested it was working fine on API15 emulator.

Frenzie:
Thanks!

PR #114, opened 2019-02-15
[fix] Strip quotes from os.execute arguments

Frenzie:
Fixes https://github.com/koreader/koreader/issues/4586.

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1637,7 +1637,12 @@ local function run(android_app_state)
     os.execute = function(command) -- luacheck: ignore 122
         if command == nil then return -1 end
         local argv = {}
-        command:gsub("([^ ]+)", function(arg) table.insert(argv, arg) end)
+        command:gsub("([^ ]+)", function(arg)
+            -- strip quotes around argument, since they are not necessary here
+            arg = arg:gsub('"(.*)"', "%1") -- strip double quotes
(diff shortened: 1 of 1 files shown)


assets/android.lua
```diff
@@ -1637,7 +1637,12 @@ local function run(android_app_state)
     os.execute = function(command) -- luacheck: ignore 122
         if command == nil then return -1 end
         local argv = {}
-        command:gsub("([^ ]+)", function(arg) table.insert(argv, arg) end)
+        command:gsub("([^ ]+)", function(arg)
+            -- strip quotes around argument, since they are not necessary here
+            arg = arg:gsub('"(.*)"', "%1") -- strip double quotes
```
Frenzie: As written this has some issues in case there's a double-quoted argument with inner single quotes, but I'm not sure if that's worth worrying about atm.

Frenzie:
Of course I tested it, what an odd question? :-p

I don't quite see what you're getting at, however. Dofile is just dofile.

Frenzie:
I'll merge this now since it'll fix several failures which I thought must have been Android-related, but actually weren't. There is at least one very obvious issue remaining (i.e., one quotes arguments to enclose spaces) but it's not caused by this PR.

pazos:
hahaha, sorry :p 

I'm reading my question again and it's completely non-sense.

PR #121, opened 2019-02-19
fix codacy issues, part 2

pazos:
Remove static context, which is unsafe beecause it can't be garbage collected
Pass application context unless interaction with the screen is needed.
Log catched Java exceptions as errors. 

~~it seems to work, but I need a few days to test.~~

### DIFF
--- a/src/org/koreader/launcher/Clipboard.java
+++ b/src/org/koreader/launcher/Clipboard.java
@@ -10,14 +10,11 @@ import android.util.Log;
 import java.util.concurrent.CountDownLatch;
 
 public class Clipboard {
+    private Context context;
+    private ClipboardManager clipboard;
 
-    private static String TAG;
-    private static Context context;
-    private static ClipboardManager clipboard;
-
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -56,7 +56,7 @@ ant debug
 
 ## To-do
 
-* a concept to deal with native Lua modules.
-* a loader for native modules that have been put into the activity's library directory?
-* a loader for obb storage, maybe? We could put native Lua modules there, for example.
-* example code for framebuffer access and more
+-   a concept to deal with native Lua modules.
```
Frenzie: Please don't change this. That's ugly af and seriously decreases readability. :-/

(It's *their* default.)

pazos: yeah, it looks ugly. I don't know if we can suppress certain warnings with a configuration file.

Frenzie: If not we can just ignore the README file wholesale if desired. I don't find it very pressing either way. :-)

pazos: I love green A's and hate those orange E letters. :cry:

PR #124, opened 2019-02-26
[wip]: minimal support for rk33xx

pazos:
likebook mimas and probably others.

Related to https://github.com/koreader/koreader/issues/4595


pazos:
Errrr. It seems that device recognition is wrong. We need to wait :P

Frenzie:
@pazos Same issue in crengine earlier. I don't know, it's not that hard to set up your own stuff. Heck, Codacy even provides a Docker image.

Anyway, should be ready to merge? (Although you stick in an `@SuppressWarnings("PMD.UnusedFormalParameter")`.)

PR #126, opened 2019-03-02
add support for build "flavors"

pazos:
So we can check the flavor at runtime and have different behaviour based on that (ie: ota updates)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,15 @@ ifdef ANDROID_VERSION
 	VERSION?=$(ANDROID_VERSION)
 endif
 
+# support different flavors
+ifdef ANDROID_FLAVOR
+	FLAVOR?=$(ANDROID_FLAVOR)
+endif
+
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -31,9 +31,15 @@ ifdef ANDROID_VERSION
...
 
+# support different flavors
+ifdef ANDROID_FLAVOR
+	FLAVOR?=$(ANDROID_FLAVOR)
+endif
+
 # Defaults
 NAME?=1.5
 VERSION?=5
+FLAVOR?="rocks"
```
Frenzie: lol, what?

pazos: the strings.xml look cool now:

```xml
    <string name="app_name">KOReader</string>
    <string name="app_flavor">rocks</string>
```

I'm open to change default flavor :+1:

Frenzie: Works for me. ;-)

PR #127, opened 2019-03-03
show in the app picker for supported mimetypes **just** if scheme=file.

pazos:
# android bullshit starts:

for android 4.0 - 7.1 the application will be shown in the application picker for viewing supported documents
for android 8.0+ the application will be ignored even on supported documents. 

related to https://github.com/koreader/koreader/issues/4441

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -30,6 +30,7 @@
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
+                <data android:scheme="file" />
                 <data android:mimeType="application/pdf" />
                 <data android:mimeType="image/djvu" />
                 <data android:mimeType="image/vnd.djvu" />


AndroidManifest.xml
```diff
@@ -30,6 +30,7 @@
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
+                <data android:scheme="file" />
```
Frenzie: This doesn't affect <8 at all?

pazos: Short anwser: no

Long answer: it depends on the program which starts our activity. If it provides a file scheme then, no. If it provides a content scheme then yes. But anyways, since we can't get a reliable file path from content, it shouldn't hurt. In fact it is already used in the Manifest for the *.*.* mimetype workaround.

On oreo apps won't provide file schemes, on versions < 4.4 apps won't provide content schemes. On android versions between 5 and 7.1 it depends on how the app is coded.

Frenzie: Hm, lovely.

PR #133, opened 2019-03-20
open urls in external browser

pazos:
Related to https://github.com/koreader/koreader/issues/4821 and https://github.com/koreader/koreader/pull/4822

Tested on a device opening "https://koreader.rocks".

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1554,6 +1554,20 @@ local function run(android_app_state)
         end)
     end
 
+    android.openLink = function(link)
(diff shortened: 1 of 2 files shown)


pazos:
@Frenzie: will you take care of the frontend code? It should be almost the same, but replacing

`return os.execute("xdg-open '"..link.."'") == 0`  with 
`return android.openLink(link) == 0`

Frenzie:
Alright, I'll do it quickly before I leave.

PR #137, opened 2019-04-02
add logd to log luajit-launcher debug strings

pazos:
Reason: while adding new stuff to luajit-launcher I need to read a lot of logs. KOReader debug statements (logged with the verbose level) are not relevant here, so having a separate log level comes handy.

This is my proposal: debug statements are not printed in release builds. If you find the need to include (some of) them in release builds I'm ok with that as long as I can still log the relevant launcher stuff with `KOReader:D` without all the noise logged by the frontend.

Pinging @NiLuJe and @poire-z

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1121,6 +1121,9 @@ end
 function android.LOGV(message)
     android.LOG(ffi.C.ANDROID_LOG_VERBOSE, message)
 end
+function android.LOGD(message)
+    android.LOG(ffi.C.ANDROID_LOG_DEBUG, message)
+end
 function android.LOGI(message)
     android.LOG(ffi.C.ANDROID_LOG_INFO, message)
 end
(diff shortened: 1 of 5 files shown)


poire-z:
(Not really familiar with logcat and debugging on android, but your proposal and the code look allright to me.)

assets/android.lua
```diff
@@ -1301,6 +1304,29 @@ local function run(android_app_state)
     -- update logger name
     android.log_name = android.getName()
 
+
```
Frenzie: nit ;-)

assets/dl.lua
```diff
@@ -32,7 +32,7 @@ local dl = {
 }
 
 local function sys_dlopen(library)
-    A.LOGV(string.format("dl.lua - sys_dlopen - loading library %s", library))
+    A.DEBUG(string.format("dl.lua - sys_dlopen - loading library %s", library))
```
Frenzie: I realize this file is noisy, but note my rationale in https://github.com/koreader/android-luajit-launcher/pull/72.

pazos: We can leave the whole dl.lua file (or specific statements) logging to LOGV,

Frenzie: Sounds good to me. :-)

pazos:
Done! The next PR will be insane, so it is better to have this merged before :)

Frenzie:
Alright, should be good. :-D

PR #144, opened 2019-05-07
[wip]: add support for generic epd driver and a stub for known device bugs

pazos:
Untested. Waiting feedback in https://github.com/koreader/koreader/issues/4881#issuecomment-489374278

to-do: 

- disable wakelock switching if device needs wakelocks.
- update EPDController interface to support region/full updates based on current device/driver

thinking about:

- update android.screen.width and android.screen.height each time the surface was changed (needed for rotation, split screen, 3rd party keyboards...)


pazos:
closing for now!

PR #145, opened 2019-06-05
[wip] generic workaround for eink devices and full support for new tolinos

pazos:
It is https://github.com/koreader/android-luajit-launcher/pull/144 v2, without the wakelocks workaround for now.

I added android emulator for x86 in order to test myself and a couple of devices waiting for feedback:

nook glowlight 3 in https://github.com/koreader/koreader/issues/5056
sony dpt-r1 in https://github.com/koreader/koreader/issues/4881

It will probably won't work as is (~~maybe I need to use a SurfaceView instead of a View??~~)


pazos:
Closing this, focus on #146

PR #146, opened 2019-06-08
Add support for refreshing eink on some Freescale devices

pazos:
Tested on Tolino Vision 2, Tolino Shine 3 and Nook Glowlight 3. It should work on other Freescale/NTX devices too.

Updated the einkTest accordly and simplify its interface a little bit (~~don't display device info~~, add a share button instead)

### DIFF
--- a/eink-test/src/org/koreader/test/eink/MainActivity.java
+++ b/eink-test/src/org/koreader/test/eink/MainActivity.java
@@ -1,18 +1,19 @@
 package org.koreader.test.eink;
 
+import android.content.Intent;
+import android.graphics.Point;
 import android.util.Log;
+import android.view.Display;
 import android.view.View;
 import android.widget.Button;
 import android.widget.TextView;
 
 
 public class MainActivity extends android.app.Activity {
-    private static final String TAG = "eink-test";
 
-    private static final int RK30xxNORMAL = 1;
-    private static final int RK30xxFORCED = 2;
-    private static final int RK33xxNORMAL = 3;
-    private static final int FAKE = 999;
+    private static final int RK30xx = 1;
+    private static final int RK33xx = 2;
+    private static final int NTX_NEW= 3;
 
     private static final String MANUFACTURER = android.os.Build.MANUFACTURER;
     private static final String BRAND = android.os.Build.BRAND;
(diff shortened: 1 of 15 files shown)


pazos:
Will fix codacy new issues on einkTest in a few minutes!

eink-test/src/org/koreader/test/eink/MainActivity.java
```diff
@@ -1,119 +1,142 @@
...
+        /** current device info */
+        info.append("Manufacturer: " + MANUFACTURER + "\n");
+        info.append("Brand: " + BRAND + "\n");
+        info.append("Model: " + MODEL + "\n");
+        info.append("Product: " + PRODUCT + "\n");
+        info.append("Hardware: " + HARDWARE + "\n");
+
+        /** add platform if available */
+        String platform = getBuildProperty("ro.board.platform");
+        if (!("".equals(platform) && (platform == null))) {
```
Frenzie: Should that be an or instead?

pazos: the reason behind that shit is codacy issues suggesting .equals instead of ==. My code was meant to check that platform is not empty and not null. But to avoid codacy issues it ends being check that not (platform is empty and platform is null)

pazos:
> lgtm, but what about display + share instead of one or the other.

It takes space on the UI and allows people to report invalid id strings. In general I don't like to spend much time working on the test activity so I'm open to suggestions

pazos:
@Frenzie 

Yeah, both log and show :+1: 
undtitled

eink-test/src/org/koreader/test/eink/MainActivity.java
Frenzie: Sure, I'm not talking about the memory reference thing but about the check logic.

If `platform.equals("")` is `true`, `platform == null` must be `false`, and vice versa.

Therefore the outcome of the check should always be the same, unless you write `!("".equals(platform) || (platform == null))`.

```java
// I think it would be more legible this way, but that aside
platform != null && !platform.isEmpty()
```

hiworld3:
can you release this new eink test apk for us to test? i can't find the link
thanks!!

mrhrzg:
I can test on tolino shine hd 2. It's there a apk for this yet. Won't be able to build; I'm away from my tools...

Frenzie:
@mrhrzg The einkTest.apk as well as the regular release are available at https://github.com/koreader/koreader/releases/tag/v2019.06

mrhrzg:
I am stoked to report that with v2019.06 KOReader works on **tolino shine 2 hd*z. Amazing work and thank you so much!

mrhrzg:
I am stoked to report that with v2019.06 KOReader works on **tolino shine 2 hd**. Amazing work and thank you so much!

Issue #131, opened 2019-03-11
Get notch height on API >= 28


pazos:
Related to https://github.com/koreader/koreader/issues/4769

Mainly we need to add support for https://developer.android.com/reference/android/view/WindowInsets#getDisplayCutout().

and return screen.height = screen.height - notch_height

PR #149, opened 2019-06-22
gradle migration

### LINKED ISSUE #60: Switch to Gradle
It's the future… for now. One hopes it won't be discarded a few years down the line. in any case, SDK 26 doesn't work well with the current Ant-based system.

### LINKED ISSUE #129: Request permissions on api > 23
we can now build with sdk tools < 25.3.0 and sdk api 19 - 22. Newer sdk won't work.

we need to request permissions at runtime: https://developer.android.com/guide/topics/permissions/overview#dangerous-permission-prompt

 sdcard read/write is a "insecure" permission.

write settings is special: https://developer.android.com/guide/topics/permissions/overview#special_permissions

pazos:
- nativeActivity migration to gradle.
- einkTestActivity migration to gradle.
- migrate ant properties  (png crunching, application name, build flavors ...)
- update Makefile to use gradle instead of ant and rename apks and move them to bin folder.
- add support for runtime permissions
- fixed some potential memory leaks (using application context and calling DeviceInfo class directly)
- update the README

Fixes #60 
Fixes #129 
Unblocks #131


Frenzie:
Great work so far!

pazos:
@Frenzie: thanks!

I found a way to manage assets per abi, it is a work in progress.

Right now app loads fine (on both archs) and crash because it can't write on cache file (because we are building with api 23+ and not asking for permisssions yet)

pazos:
About luajit build, It works nice building manually. The benefits of writting a CMake script is repurposing the NDK from android-studio without the need to specify the path to NDK.

Pinging our cross-compiler master @NiLuJe: do you see a easy way of migrating mk-luajit.sh to cmake?.  The CMakelists.txt that I find online are not meant for cross-compilation and include checks we don't need here, like readline.

NiLuJe:
IIRC, The NDK ships a CMake ToolChain file. It might extra-hairy, and with a shitload of stuff we don't technically need, but Andorid being what it is, I *think* that's the recommended way to go by now.

It's (hopefully) as simple as passing that to cmake, like we [https://github.com/koreader/koreader-base/blob/3b9c5c1ca3d231fd399d5b9fe0d82fcd6fa6ebfd/Makefile.defs#L513](already do) for eInk platforms.

pazos:
Okay, switching luajit to CMake is beyond my interest. I've updated the Makefile and updated instructions on the first post.

I will try to fix the request permission bug later today

pazos:
fixed runtime permissions!

pazos:
Ok, did a few more tests. Everything is working.

What's left is finding a way to replace the old `android project update` thingy which is useful to detect ndk and sdk paths.

Android Studio will figure out paths for us but the IDE has a newer ndk not compatible with API14 (minimum is API17).

So repurposing our toolchain would be as simple as:

```
sdk.dir=/path/to/ko/base/toolchain/android-sdk-linux
ndk.dir=/path/to/ko/base/toolchain/android-ndk-r15c
```
That way we ensure that both koreader-base, luajit and the launcher native code are built with the same toolchain and we have compabitility with apis 14-28.

I would need some help figuring out how to build KOReader with ./kodev for both arm and x86 in a single shot and copy assets module to their respective places, any hint @Frenzie ?

Frenzie:
I think it makes the most sense to let `kodev build android` do some equivalent of `kodev build android-arm && kodev build android-x86`, because just for basic development/testing most of the time you'll probably only use one or the other. Everything would need very few changes that way, other than adding the arch over here:

https://github.com/koreader/koreader/blob/fca51757af6dace42d72163816ca12b52a2ee3aa/Makefile#L341-L343

The `android-x86` target could just run `android-arm` with `ANDROID_ARCH=x86`.

The bigger problem is making sure it works in the CI on GitLab. I think it might time out if the build time exceeds an hour.

pazos:
> The bigger problem is making sure it works in the CI on GitLab. Iirc it times out if the build time exceeds an hour.

Uouo, I just found that is possible to do `./gradlew assembleX86Release` even when that specific task is not listed by gradle.

So, great!

PR #153, opened 2019-06-29
A few things after switching to gradle.

pazos:
Use androidX support libraries instead of legacy. (Requires targetApi 28)
Move network/device methods from MainActivity to their own classes
Fix version code and version name

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,32 @@ ifdef ANDROID_ARCH
 	endif
 endif
 
+
 # Default is build for arm
 ANDROID_FULL_ARCH?=armeabi-v7a
 GRADLE_TASK?=assembleArm
+
+# find the path where the SDK is installed
+ifdef SDK
+        ANDROID_SDK_FULLPATH?=$(SDK)
+else
+        ifdef ANDROID_HOME
+                ANDROID_SDK_FULLPATH?=$(ANDROID_HOME)
+        else
+                ANDROID_SDK_FULLPATH?=`realpath ../../../base/toolchain/android-sdk-linux`
+        endif
+endif
+
+# find the path where the NDK is installed
+ifdef NDK
+	ANDROID_NDK_FULLPATH?=$(NDK)
+else
+	ifdef ANDROID_NDK_HOME
+		ANDROID_NDK_FULLPATH?=$(ANDROID_NDK_HOME)
+	else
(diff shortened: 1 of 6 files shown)


launcher/src/org/koreader/launcher/Device.java
```diff
@@ -0,0 +1,83 @@
...
+
+public class Device {
+
+    private static final String PRODUCT = DeviceInfo.PRODUCT;
+    private static final boolean HAS_EINK_SUPPORT = DeviceInfo.EINK_SUPPORT;
+    private static final boolean HAS_FULL_EINK_SUPPORT = DeviceInfo.EINK_FULL_SUPPORT;
+    private static final boolean NEEDS_WAKELOCK_ENABLED = DeviceInfo.BUG_WAKELOCKS;
+
+    private String TAG;
+    private Context context;
```
Frenzie: Codacy is complaining about this one. I'm not sure if the line is necessary at a quick glance? The constructor seems to be on line 21?

pazos: done!

Makefile
```diff
@@ -5,10 +5,33 @@ ifdef ANDROID_ARCH
...
 GRADLE_TASK?=assembleArm
 
+# find the path where the SDK is installed
+ifdef SDK
+        ANDROID_SDK_FULLPATH?=$(SDK)
+else
+        ifdef ANDROID_HOME
+                ANDROID_SDK_FULLPATH?=$(ANDROID_HOME)
+        else
+                ANDROID_SDK_FULLPATH?=`realpath ../../../base/toolchain/android-sdk-linux`
```
Frenzie: Shouldn't that be something like:
```suggestion
                ANDROID_SDK_FULLPATH?=$(shell realpath ../../../base/toolchain/android-sdk-linux)
```

Makefile
```diff
@@ -5,10 +5,33 @@ ifdef ANDROID_ARCH
...
+endif
+
+# find the path where the NDK is installed
+ifdef NDK
+	ANDROID_NDK_FULLPATH?=$(NDK)
+else
+	ifdef ANDROID_NDK_HOME
+		ANDROID_NDK_FULLPATH?=$(ANDROID_NDK_HOME)
+	else
+		ANDROID_NDK_FULLPATH?=`realpath ../../../base/toolchain/android-ndk-r15c`
```
Frenzie: Same here.

PR #156, opened 2019-07-04
add support for 3rd party dict apps

pazos:
implemented a few custom intents for some 3rd party apps (Aard2, Quickdic and Colordict) and generic intents for share/send actions.

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1679,6 +1679,37 @@ local function run(android_app_state)
         end)
     end
 
+    android.isPackageEnabled = function(package)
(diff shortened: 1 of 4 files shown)


Frenzie:
I would suggest Kiwix as a major alternative as well.

pazos:
> I would suggest Kiwix as a major alternative as well.

It should work impelementing the process_text intent in https://github.com/kiwix/kiwix-android/blob/develop/app/src/main/AndroidManifest.xml#L125-L131

PR #158, opened 2019-07-05
add epd support  for crema 1010p eink device

pazos:
Based on https://github.com/koreader/koreader/issues/4551#issuecomment-507016109 and https://github.com/koreader/koreader/issues/4551#issuecomment-507177518


nemorami:
It fixed 2019.07 version, but next new version have same problem. only 2017.07 version works.

PR #160, opened 2019-07-24
Add options to modify screen timeout while the application is running

pazos:
Until now we relied on android wide settings or use wakelocks to disable the timeout.

This PR adds other timeout options, which are embedded on the activity lifecycle:

- when the app is resumed use a custom timeout , if any, or acquire a new wakelock.
- when the app is paused restore default timeout or release the wakelock.

Since users can modify the screen timeout from android settings we need to update system timeout on resume too.

Some logs to show how this is integrated into the lifecycle of the activity:

#### activity pause/resume with no custom option

```adb logcat
D/KOReader( 3758): timeout callback, focus: false -> using system settings
D/KOReader( 3758): timeout callback, focus: true -> using system settings
```

#### activity pause/resume with custom option

```adb logcat
D/KOReader( 3758): set timeout for KOReader: 600 seconds
D/KOReader( 3758): timeout callback, focus: false -> using custom settings
D/KOReader( 3758): restoring system timeout: 600 -> 30 seconds
D/KOReader( 3758): timeout callback, focus: true -> using custom settings
D/KOReader( 3758): restoring KOReader timeout: 30 -> 600 seconds
```

Please feel free to suggest more descriptive names for the new methods. I'm clueless about it :man_facepalming:

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1407,6 +1407,27 @@ local function run(android_app_state)
         end)
     end
 
+    android.getScreenTimeout = function()
+        return JNI:context(android.app.activity.vm, function(JNI)
+            return JNI:callIntMethod(
+                android.app.activity.clazz,
+                "getScreenTimeout",
+                "()I"
+            )
+        end)
+    end
+
+    android.setScreenTimeout = function(timeout)
+        JNI:context(android.app.activity.vm, function(JNI)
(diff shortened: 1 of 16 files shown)


pazos:
@Frenzie: can I merge this?

Frenzie:
@pazos Go ahead!

PR #161, opened 2019-07-26
refactor: java inheritance and fix some warnings/lints

pazos:
A friend of mine lent me "Thinking in Java", after a couple of hours reading it I became a hardcore inheritance monk. Things went crazy and now I have an all-in-one PR. Sorry :man_facepalming: 

## Native

- To improve compatibility with Android Studio I moved each lib in their own subfolder with their own Android.mk file. Moved git submodules too.

- Added a local copy of android_native_app_glue, so we can share logger name and apply the same logger behaviour.

- Changed  the name "main thread" to "NativeThread". The main thread is running on the VM and handled normally via android framework and the native thread is running outside the VM using app_glue(pthreads).

## Java

- **Improves logging**: we extend android.app.Application in order to obtain the application name. Logger is now easier because it will require only a message, as we log everything with the app name as the logger tag (as we do in C and lua).

- **Improves inheritance** and hopefully is now easier to read.

    - Java methods invoked from lua are declared in **ILuaJNI**.

    - **NativeSurfaceView** extends from SurfaceView. Implements a placeholder for our native window inside a view hierarchy and adds a few methods to update the view on e-ink devices.

    - Base implementation of lua/JNI methods is done in **BaseActivity**.

    - **MainActivity** extends from *BaseActivity*, takes ownership of the native window using a new *NativeSurfaceView* and implements support for runtime permissions. It also overrides some BaseActivity methods with implementations that are view-aware.

- **Fixes the underlying cause behind most warnings**: most of them were NullPointerException and were fixed by checking if the returned value was not null.

- Suppress the rest of warnings reported by firing up `android studio -> Analyze -> Inspect Code`: **Trivial warnings are disabled in gradle scripts** and **the rest are suppressed using annotations** because they affect a few classes, methods or fields.

- **Modifies the Makefile**:
  - Enable all gradle warnings
  - More informative messages
  - add `make test` to build all the tests (currently there is only one: einkTest).
  - `make release` and `make debug` are used to build the NativeActivity.apk.

- Minor gradle plugin bump
- renamed device package from org.koreader.device to org.koreader.launcher.device
- created a separate package for helpers.


launcher/src/org/koreader/launcher/IntentUtils.java
```diff
@@ -49,6 +49,28 @@ public static Intent getByAction(String text, String pkg, String action) {
...
+
+    @SuppressWarnings({"ConstantConditions", "unused"})
+    public static String intentToString(Intent intent) {
+        if (intent == null) return "";
+        StringBuilder stringBuilder = new StringBuilder("\naction: ")
+            .append(intent.getAction())
+            .append("\ndata: ")
+            .append(intent.getDataString())
+            .append("\nextras: ")
+            ;
```
Frenzie: Huh, interesting. Makes sense (similar to trailing comma) but I don't think I've ever seen it.

Frenzie:
Oh btw did you want to get this in the upcoming stable (tomorrow at the earliest; probably a few more days due to translations) or wait for the next one?

pazos:
> Oh btw did you want to get this in the upcoming stable (tomorrow at the earliest; probably a few more days due to translations) or wait for the next one?

I would say wait for the next release. I profiled the application using Android Studio and things are *almost* the same as they were before applying this PR. Just need time to (re)review the new classes and see if everything makes sense.

pazos:
> I trust you on the `@SuppressWarnings` stuff, besides that it's probably slightly  easier to understand this way.

Some of the (unused/unused return values) are here to stay, as they are reported for the launcher project but not for the einkTest project. The same happens on epd controller abstract classes, which are firing weaker access warnings in the context of a project but they need to be public as they are copied and used outside that project. Also all methods of the mainActivity called using JNI are marked as unused because the inspection tools don't really understand that they are used from lua.

pazos:
@Frenzie: can we disable https://pmd.github.io/latest/pmd_rules_java_codestyle.html#defaultpackage in codacy?.

Encapsulation is cool, but to get rid of these warnings we're declaring methods/fields as public and suppressing WeakerAccess warnings.

Frenzie:
We'll have to make a custom ruleset.xml. Pity that the Codacy docs are rather vague even when existent.

pazos:
> We'll have to make a custom ruleset.xml. Pity that the Codacy docs are rather vague even when existent.

I just disabled it from Codacy webpage (I logged in with github).

Frenzie:
That specific rule you mean?

The Codacy web GUI is alright but then it's not very useful for local checks.

pazos:
> The Codacy web GUI is alright but then it's not very useful for local checks.

Sure. We want our own ruleset in the long run. I started to make one bindly (before logging in codacy): 
[ruleset.xml.zip](https://github.com/koreader/android-luajit-launcher/files/3483424/ruleset.xml.zip) - remove extension.

Codacy does not provide ready available rulesets.xml, so we need to do it manually (and there are more than 200 rules for PMD/Java)

pazos:
I added a local copy of android_native_app_glue to our source tree. It makes it easier for us to customize  it (ie: for logging with "tag: [subtag]: message" scheme).

IMO it makes a bit easier to understand what is happening because the (verbose) comments in the sources.

So, instead of requiring `adb logcat KOReader:* threaded_app:* *:S` to log the interaction between our program and the native glue we just type `adb logcat KOReader:* *:S`. All the logs that app_glue shows will be ignored on release builds.

Frenzie:
I hope it won't be annoying to update with NDK/SDK updates. :-)

pazos:
Nah, it never changed since API9.

pazos:
@Frenzie: please give it a quick review when you have some time (and re-read the first post). I think I'm not going to do more big changes, just little fixes.

Makefile
```diff
@@ -71,40 +71,45 @@ update:
...
+	(cd jni/luajit && ./mk-luajit.sh $(ANDROID_FULL_ARCH))
 
 prepare: update
-	# for Android Studio users. Build luajit for all supported abis
-	./mk-luajit.sh clean
-	./mk-luajit.sh x86
-	./mk-luajit.sh clean
-	./mk-luajit.sh armeabi-v7a
+	# for Android Studio users: build luajit for all supported abis.
+	(cd jni/luajit && ./mk-luajit.sh clean)
```
Frenzie: Might be easier to parse like this?
```
cd jni && mk-lijist && \
mk-luajit x86 && \
etc.
```
Or in any case it'd align with how we do it elsewhere; doesn't really matter as far as the result goes I suppose.

https://github.com/koreader/koreader-base/blob/76a0ba186b7268449301da1cf444d29dac21cfa2/Makefile.third#L18-L22

app/src/org/koreader/launcher/device/DeviceInfo.java
```diff
@@ -35,68 +30,89 @@
...
+
+        // we use the standard android build properties for device identification
         MANUFACTURER = getBuildField("MANUFACTURER");
         BRAND = getBuildField("BRAND");
         MODEL = getBuildField("MODEL");
         DEVICE = getBuildField("DEVICE");
         PRODUCT = getBuildField("PRODUCT");
 
-        HashMap<EinkDevice, Boolean> deviceMap = new HashMap<EinkDevice, Boolean>();
+        IS_BOYUE = MANUFACTURER.toLowerCase().contentEquals("boeye") ||
```
Frenzie: Btw, not related to the refactor but I assume this is not a typo?

app/src/org/koreader/launcher/device/EPDFactory.java
```diff
@@ -70,7 +69,7 @@ public static EPDController getEPDController(final String tag) {
     private static class FakeEPDController implements EPDController {
         @Override
         public void setEpdMode(android.view.View targetView, int mode, long delay, int x, int y, int width, int height, String epdMode) {
-            /* do nothing */
+            Logger.d("fake update");
```
Frenzie: What does this look like in the log? As a user I like log messages to be somewhat self-explanatory without reading the code, so preferably with a tag or ever so slightly more verbose with `fake EPD update`.

pazos: It is never reached on KOReader as we don't trigger eink updates on devices without eink support. It is useful on third party code where you can use getEPDFactory to receive an EPDController without caring if the device has in fact a support epd driver.

The reason I added the log statement is that lints do not like empty methods.

app/src/org/koreader/launcher/device/DeviceInfo.java
pazos: not, boyue and boeye based on FW version. Since I'm not sure which device has one or the other we check both.

pazos:
Now luajit-launcher logs are better: 

We log application info before starting any activity. 
We log activities, helpers and native_glue with a subtag.
We log available height and width from NativeSurfaceView

That doesn't affect the standard cross-platform KOReader logs in any way.

Here is what happens on early init when we launched KOReader in Android:

## release (info, warn, error)

```adb logcat
adb logcat KOReader:I *:S

I/KOReader(10884): Application started
```

## release (verbose, debug, info, warn, error)

```adb logcat
adb logcat KOReader:V *:S

I/KOReader(10884): Application started
V/KOReader(10884): Application info {
V/KOReader(10884):   Flags: user, debuggable
V/KOReader(10884):   Paths {
V/KOReader(10884):     Assets: /data/data/org.koreader.launcher/files
V/KOReader(10884):     Library: /data/app-lib/org.koreader.launcher-1
V/KOReader(10884):     Storage: /mnt/sdcard
V/KOReader(10884):   }
V/KOReader(10884): }
V/KOReader(10884): [NativeThread]: waiting for activity
V/KOReader(10884): [NativeSurfaceView] surface created
V/KOReader(10884): [NativeThread]: activity window ready.
V/KOReader(10884): [NativeSurfaceView] surface changed {
V/KOReader(10884):   width:  1080
V/KOReader(10884):   height: 1776
V/KOReader(10884): }
V/KOReader(10884): [NativeThread]: activity gained focus.
V/KOReader(10884): [NativeThread]: launching LuaJIT assets
```


 ## debug (verbose, debug, info, warn, error)

```adb logcat
adb logcat KOReader:V *:S

I/KOReader(10884): Application started
V/KOReader(10884): Application info {
V/KOReader(10884):   Flags: user, debuggable
V/KOReader(10884):   Paths {
V/KOReader(10884):     Assets: /data/data/org.koreader.launcher/files
V/KOReader(10884):     Library: /data/app-lib/org.koreader.launcher-1
V/KOReader(10884):     Storage: /mnt/sdcard
V/KOReader(10884):   }
V/KOReader(10884): }
D/KOReader(10884): [MainActivity] onCreate()
D/KOReader(10884): [BaseActivity] onCreate()
D/KOReader(10884): [NativeGlue]: Creating: 0xb8f44dc0
D/KOReader(10884): [NativeThread]: starting
V/KOReader(10884): [NativeThread]: waiting for activity
D/KOReader(10884): [ClipboardHelper] Starting
D/KOReader(10884): [NetworkHelper] Starting
D/KOReader(10884): [PowerHelper] Starting
D/KOReader(10884): [ScreenHelper] Starting
D/KOReader(10884): [NativeSurfaceView] Starting
D/KOReader(10884): [NativeGlue]: Start: 0xb8f44dc0
D/KOReader(10884): [NativeGlue]: activityState=10
D/KOReader(10884): [MainActivity] onResume()
D/KOReader(10884): [NativeGlue]: Resume: 0xb8f44dc0
D/KOReader(10884): [NativeGlue]: activityState=11
D/KOReader(10884): [NativeGlue]: InputQueueCreated: 0xb8f44dc0 -- 0xb8f7df30
D/KOReader(10884): [NativeGlue]: APP_CMD_INPUT_CHANGED
D/KOReader(10884): [NativeGlue]: Attaching input queue to looper
D/KOReader(10884): [BaseActivity] onAttachedToWindow()
V/KOReader(10884): [NativeSurfaceView] surface created
D/KOReader(10884): [NativeGlue]: NativeWindowCreated: 0xb8f44dc0 -- 0xb8fc08e8
D/KOReader(10884): [NativeGlue]: APP_CMD_INIT_WINDOW
V/KOReader(10884): [NativeThread]: activity window ready.
V/KOReader(10884): [NativeSurfaceView] surface changed {
V/KOReader(10884):   width:  1080
V/KOReader(10884):   height: 1776
V/KOReader(10884): }
D/KOReader(10884): [NativeGlue]: WindowFocusChanged: 0xb8f44dc0 -- 1
V/KOReader(10884): [NativeThread]: activity gained focus.
V/KOReader(10884): [NativeThread]: launching LuaJIT assets
```

PR #162, opened 2019-07-31
add an option to ignore volume keys or use them to turn pages

pazos:


assets/android.lua
```diff
@@ -1313,6 +1407,28 @@ local function run(android_app_state)
         end)
     end
 
+    android.areVolumeKeysIgnored = function()
+        return JNI:context(android.app.activity.vm, function(JNI)
+            return JNI:callIntMethod(
+                android.app.activity.clazz,
+                "areVolumeKeysIgnored",
```
Frenzie: Or maybe just something like `volumeKeysIgnored` or `getVolumeKeysIgnored`? Having it as a question seems a bit odd. ;-)

pazos: the getter/setter approach looks fine!

PR #164, opened 2019-08-03
avoid the JNI/Java overhead on input events

pazos:
JNI calls cost a little in terms of user experience (<=1ms) but introduce a delay when used in a motion event pool. Is the delay noticeable? YMMV. 

This fixed broken behaviour introduced in #162 and #163: choosing the JNI/Java thingy instead of keeping everything on lua land.


pazos:
Merge it now, for the next nightly

PR #171, opened 2019-08-21
helpers shouldn't hold a reference to the activity

pazos:
As they can be used outside activities for doing long term app things.
Switch to in-app brightness.

~~Barely tested,~~ can wait...


pazos:
The rule `Use explicit scoping instead of the default package private level` is checked again in Codacy. I dunno why. I'm going to disable it (again)

PR #172, opened 2019-08-23
workaround for Tolinos on FW10

pazos:
see https://www.mobileread.com/forums/showpost.php?p=3881771&postcount=22

### DIFF
--- a/app/src/org/koreader/launcher/BaseActivity.java
+++ b/app/src/org/koreader/launcher/BaseActivity.java
@@ -116,7 +116,11 @@ abstract class BaseActivity extends NativeActivity implements JNILuaInterface {
     /* device */
     public String getEinkPlatform() {
         if (DeviceInfo.EINK_FREESCALE) {
-            return "freescale";
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
+                return "freescale";
+            } else {
+                return "freescale-legacy";
+            }
         } else if (DeviceInfo.EINK_ROCKCHIP){
             return "rockchip";
         } else {


app/src/org/koreader/launcher/BaseActivity.java
```diff
@@ -116,7 +116,11 @@ public void setClipboardText(final String text) {
     /* device */
     public String getEinkPlatform() {
         if (DeviceInfo.EINK_FREESCALE) {
-            return "freescale";
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
+                return "freescale";
+            } else {
+                return "freescale-legacy";
```
Frenzie: How would this affect a non-Tolino on <Jelly Bean? Or is the problem wider than Tolino?

pazos: > How would this affect a non-Tolino on <Jelly Bean? 

This string is just used in the "report a bug" infomessage and now to check if we need to take control of the native window in MainActivity.java.

A device reporting "freescale-legacy" will report itself as `HAS_EINK`, `HAS_EINK_FULL` and `IS_FREESCALE` but it will fail when trying to update the screen because it has no control over the underlying surface.

> Or is the problem wider than Tolino?

Non sure about the Crema device supported but I think its api should be >= 16.

pazos: Also not sure about Nooks.

But, IMO, lgtm and hopefully somebody with an affected device will fill a new ticket.

Frenzie: Damn the torpedoes, full speed ahead! ;-)

PR #174, opened 2019-08-25
migrate from lzma to zip file. Add error handling.

pazos:
Just a first version. It seems to work fine, but **WIP**.


app/src/org/koreader/launcher/AssetsUtils.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private static boolean copyUncompressedAssets(Context context) {
+        AssetManager assetManager = context.getAssets();
+        String filesPath = context.getFilesDir().getAbsolutePath();
+        try {
+            // llapp_main.lua is the entry point for frontend code.
+            boolean entry_point;
+            String[] assets = assetManager.list("module");
+            for (String asset: assets) {
+                if ("llapp_main.lua".equals(asset) {
```
Frenzie: You'll notice this soon enough, but:
```suggestion
                if ("llapp_main.lua".equals(asset)) {
```

Frenzie: Err, actually the version you force pushed is a bit more broken. ;-)

pazos: Ready now

Frenzie: Cool! Trying to figure out how to refactor util due to the somewhat mysterious https://github.com/koreader/koreader/issues/5269 atm though. :-)

(I see where the circularity comes in, but not why I'm not getting it on desktop, Kobo or Android while this Kindle apparently is.)

Frenzie: @pazos I can't figure out how this native libs bundling stuff works (or at least I can't in a minute or two and I don't want to waste my time on a solved problem).

I added this but it's ignoring the libs while creating the APK.
```diff
diff --git a/app/AndroidManifest.xml b/app/AndroidManifest.xml
index eb39428..aa84d79 100644
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -19,6 +19,7 @@
         android:vmSafeMode="true"
         android:allowBackup="false"
         android:fullBackupContent="false"
+android:extractNativeLibs="true"
         android:usesCleartextTraffic="true">
         <meta-data android:name="android.max_aspect" android:value="3.1" />
         <activity
diff --git a/app/build.gradle b/app/build.gradle
index 77b2902..ce9c033 100644
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -34,6 +34,7 @@ android {
             java.srcDirs = ['src']
             res.srcDirs = ['res']
             assets.srcDirs = [ '../assets' ]
+            jniLibs.srcDirs = [ '../libs' ]
         }
     }
```

pazos: @Frenzie: now I just finished with this PR unless we catch some error. Shall I send another PR with the libraries branch? It should be easier for you to test (er, or at least we will share the same bugs :smile: )

app/src/org/koreader/launcher/AssetsUtils.java
```diff
@@ -0,0 +1,168 @@
...
+        } catch (IOException e) {
+            Logger.e(TAG, "error with raw assets:\n" + e.toString());
+            entry_point = false;
+        }
+        return entry_point;
+    }
+
+    /* unzip from stream */
+    private static void unzip(InputStream stream, String output) {
+        byte[] buffer = new byte[BASE_BUFFER_SIZE * 512];
```
Frenzie: What does this mean exactly?

assets/android.lua
```diff
@@ -2060,28 +2070,26 @@ local function run(android_app_state)
...
-        android.dismissProgress()
-        local elapsed_time = os.difftime(end_time, start_time)
-        if elapsed_time > 0 then
-            android.LOGV(string.format("update installed in %d seconds", elapsed_time))
-        end
-    else
-        error("error loading install.lua")
+    -- Do not extract assets if we don't have write permissions on external storage.
+    if not android.canWriteStorage() then
+        android.notification("Insuficient permissions", true)
```
Frenzie: Maybe something more like this?
```suggestion
        android.notification("Write permission required to extract resources.", true)
```

Or if you're purposefully being vague (which I'd prefer not to be), fix the typo ;-)
```suggestion
        android.notification("Insufficient permissions", true)
```

app/src/org/koreader/launcher/AssetsUtils.java
pazos: Use a long buffer (512k) to store the stream while uncompressing. It is a bit faster than using 1kb chunks.

1kb chunks are used while copying uncompressed assets as they are expected to have less size.

app/src/org/koreader/launcher/AssetsUtils.java
Frenzie: That sounds a lot easier, yes. If this is finished we can just go ahead and merge it I'd say, and go from there. :-)

assets/android.lua
pazos: this toast is shown when the user just rejected giving the app storage permissions, so the second suggestion should be sufficient for users. But the first one is good for the lua error that logcat throws. :+1:

app/src/org/koreader/launcher/AssetsUtils.java
pazos: cool! we can merge this as-is, work on libraries and just update KOReader once we're done with both things.

PR #177, opened 2019-09-01
avoid the progress dialog unless we're extracting resources from zipfile

pazos:
See https://github.com/koreader/koreader/pull/5264#issuecomment-526927402


pazos:
> Btw, does that mean the icon splash thing will show a bit longer?

I don't think so. It will run **only** if a zipfile with different revision is found and **while** the assets are being extracted. Previously it ran on every new launch before checking assets and was dismissed when the install function finished.

Frenzie:
Good, good. :-)

PR #183, opened 2019-09-13
migration to kotlin

### LINKED ISSUE #179: Get rid of WRITE_SETTINGS permission.
Now that is only used to set custom screen off timeout.

The alternative is:

1. after each resume update system timeout (already implemented)
2. calculate a specific wakelock time using desired time - system timeout
3. disable all menu items if wakelock time < system_timeout.

An example:

An user has its screen timeout set to  2 minutes but wants to set a custom timeout of 5 minutes for KOReader. She go to screen->timeout and see all options.

Those options with less timeout than her system timeout are disabled (30', 60'..)
The other options are available. If she sets the custom timeout to 5 minutes then we will apply a timeout of 5 - 2 = 3 minutes. The wakelock will be released after 3 minutes and then the normal system timeout will apply after 2 minutes.

pazos:
- [x] Migrate everything to Kotlin, except EPD calls for different platforms and EinkTest activity.
- [x] Migrate the helper classes to objects, and move them to the main package.
- [x] Security fix: logger, deviceinfo and epdfactory are now internal. This reduces the risk of malicious calls from 3rd party apps ~~using introspection~~, still vulnerable to introspection but meh!
- [x] Reimplement timeout without using WRITE_SETTINGS
 ~~Add rumble support~~

External PR:
- [x] Add detekt configuration for kotlin static analysis and match codacy default rules.
- [x] Modify config to fix this PR current issues.

Supersedes #180 , #181 
Fixes #179


pazos:
@Frenzie: just a quick review of the things I've done, not a line by line review :p

Frenzie:
Assuming the code comes down to the same thing it looks fine to me, not too dissimilar with a bit less boilerplate.

pazos:
> Assuming the code comes down to the same thing it looks fine to me.

It should behave exactly the same.

> not too dissimilar with a bit less boilerplate

I tried to write as much idiomatic kotlin code as possible.

See the following snippet as an example of idiomatic kotlin:

```kotlin
      val modeName = when (mode) {
            1 -> "EPD_FULL"
            2 -> "EPD_PART"
            3 -> "EPD_A2"
            4 -> "EPD_AUTO"
            else -> "invalid"
        }
```

This reduces boilerplate code greatly and it is, IMO, easier to read.

I'm in for merging this PR as is unless you find some issue on the new timeout logic (which I barely tested but works fine on my end).

I'm going to submit the frontend part...

PR #187, opened 2019-09-22
Show koreader icon while loading

ColinKinloch:
Displays the koreader icon on a white background while the MainActivity loads.
Renders the koreader icon on a white background to the MainActivity Surface once loaded.

### DIFF
--- a/app/src/org/koreader/launcher/MainActivity.kt
+++ b/app/src/org/koreader/launcher/MainActivity.kt
@@ -27,7 +27,7 @@ class MainActivity : BaseActivity() {
     private val epd = EPDFactory.epdController
     private var view: NativeSurfaceView? = null
     private var takesWindowOwnership: Boolean = false
-    private var drawSplashScreen: Boolean = true
+    private var showSplashScreen: Boolean = true
 
     companion object {
         private const val TAG = "MainActivity"
@@ -36,19 +36,8 @@ class MainActivity : BaseActivity() {
 
     override fun surfaceCreated(holder: SurfaceHolder) {
         super.surfaceCreated(holder)
-        if (drawSplashScreen == true) {
-            drawSplashScreen = false
-            holder.lockCanvas()?.let { canvas ->
-                try {
-                    ContextCompat.getDrawable(this, R.drawable.splash_icon)?.let { splashDrawable ->
-                        splashDrawable.setBounds(0, 0, canvas.getWidth(), canvas.getHeight())
-                        splashDrawable.draw(canvas)
-                    }
-                } catch (e: Exception) {
-                    Logger.w(TAG, "Failed to draw splash screen:\n$e")
-                }
-                holder.unlockCanvasAndPost(canvas)
-            }
+        if (showSplashScreen == true) {
+            drawSplashScreen(holder)
         }
     }
 
@@ -202,6 +191,10 @@ class MainActivity : BaseActivity() {
             1 else 0
     }
 
+    override fun setShowSplashScreen(show: Boolean) {
+        showSplashScreen = show
+    }
+
     /*---------------------------------------------------------------
      *                       private methods                        *
      *--------------------------------------------------------------*/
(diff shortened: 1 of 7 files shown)


app/src/org/koreader/launcher/FramelessProgressDialog.java
```diff
@@ -4,6 +4,7 @@
 import android.content.Context;
 import android.widget.ProgressBar;
 import android.view.ViewGroup.LayoutParams;
+import android.view.Gravity;
```
Frenzie: The imports are alphabetical. ;-)

pazos:
Untested, a few questions:

Is this intended as a replacement for the spinning circle? In that case is this only shown when uncompressing assets?

I'm not against this patch but I wonder if this change will affect the myriad of broken devices out there (that's the reason to not use NativeSurfaceView on all devices even when I wasn't able to reproduce the lag on any emulator/device I have).

If it is just a matter of aesthetics I would suggest to wait for the kotlin PR and submit on top of that, because I changed a few things there: i.e: FramelessProgressDialog is now a private class inside BaseActitivity.

I would like to remember that I had no previous knowledge about Android, Java or mobile development before starting to contribute here, so most of my ideas might: a) be stupid, b) be broken or c) be discarded.

ColinKinloch:
The "splash_icon" drawable is a white background with the koreader icon centred independent of the size of what it's being drawn on.

The "SplashScreen" theme displays the "splash_icon" fullscreen before the app has loaded, for any device with reasonable performance this should show up for less than a second.

Once MainActivity is loaded it sets it's theme back to "Fullscreen" and the window background to black (I think the koreader renderer assumes the buffer it's writing to is cleared to black).

Once the the OS gives SurfaceView a surface "splash_icon" is rendered to it once. This seems to happen before the asset extraction spinner shows up and isn't cleared until koreader runs.

I've also set the gravity on the dialog that holds the asset extractor spinner to "bottom", this moves it to the bottom of the window. so the spinner isn't rendered on top of the koreader logo.

I've tested it on:
* tolino shine2HD (android 4.4.2)
* Boox N96 (android 4.4.4)
* Boox NovaPro (android 6.0.1)

pazos:
Awesome, thanks!

Yeah, these borked devices are mostly Ice Cream Sandwitch.

app/src/org/koreader/launcher/MainActivity.kt
```diff
@@ -41,6 +41,17 @@ class MainActivity : BaseActivity() {
         override fun surfaceCreated(holder: SurfaceHolder) {
             Logger.v(tag, "surface created")
             setWillNotDraw(false)
+            holder.lockCanvas()?.let { canvas ->
```
pazos: This won't happen unless the device is tagged as "freescale", at least in theory. I guess you can override surfaceCreated on BaseActivity if we want the change in all devices.

Not sure if surfaceCreated is the best place to run something once. AFAICT onSurfaceCreated is run each time we return from pause and in that case we probably won't want to show the icon as the time until KO is usable is almost 0.

ColinKinloch: You're right in both cases.
I misunderstood where the SurfaceHolder.Callback was coming from.

pazos:
@ColinKinloch: found a minor glitch on API23+ - first install: the dots are shown over a black screen because the surface is recreated when we allow rw permissions on sdcard and thus the drawable is deleted.

untitled

Nothing important and can wait for the future :+1:

ColinKinloch:
Perhaps `MainActivity.drawSplashScreen` can be set to false just before the C code starts koreaders lua.

pazos:
Hey, I just noticed your last commit and tested it. I didn't found any difference.

I will merge this PR as is and will try to workaround in the future.

> Perhaps MainActivity.drawSplashScreen can be set to false just before the C code starts koreaders lua.

There is a problem calling stuff from lua during initialization. That's why we request permissions on create instead of adding the request to the lua JNI interface.

ColinKinloch:
Hmm, perhaps it would be better to draw in SurfaceHolder.Callback.surfaceChanged or SurfaceHolder.Callback2.surfaceRedrawNeeded

PR #189, opened 2019-09-29
move splash screen to BaseActivity

pazos:
Just because it looks simpler and applies to all android devices. This implementation avoids the jni/lua overhead.

Splash screen gets updated on every surface change as suggested in https://github.com/koreader/android-luajit-launcher/pull/187#issuecomment-536233247.
When the assets are extracted (or skipped if already installed) we disable the flag to avoid splash screen on each resume.


pazos:
@ColinKinloch: github doesn't let me request you a review, so I would ask informally :smile:

pazos:
untitled

PR #190, opened 2019-09-30
Add support for the android vibration engine & make some strings non nullable properties.

pazos:
Needs to be tested in Android 4.0 - 7.1. Can't be tested on the emulator AFAIK
Needs to be tweaked in Android Oreo and higher because default intensity is to much to be considered haptic feedback.

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -8,6 +8,7 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE"/>
     <uses-permission android:name="android.permission.CHANGE_WIFI_STATE"/>
+    <uses-permission android:name="android.permission.VIBRATE"/>
     <uses-permission android:name="android.permission.WAKE_LOCK"/>
 
     <application
--- a/app/src/org/koreader/launcher/BaseActivity.kt
+++ b/app/src/org/koreader/launcher/BaseActivity.kt
@@ -42,6 +42,7 @@ abstract class BaseActivity : NativeActivity(), JNILuaInterface,
 
     private var customBrightness: Boolean = false
     private var showSplashScreen: Boolean = true
+    private var hasVibrator: Boolean = false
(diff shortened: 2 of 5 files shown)


Frenzie:
We can't use the constants from <https://developer.android.com/reference/android/view/View.html#performHapticFeedback(int)>?

https://developer.android.com/reference/android/view/HapticFeedbackConstants.html is basically why I wrote it the way I did. :-)

pazos:
> We can't use the constants from https://developer.android.com/reference/android/view/View.html#performHapticFeedback(int)?

No with the vibration api. I'm not familiar with the haptic feedback api but I see it is part of the view and has some advantages to use in that context that we cannot do. And it seems to rely on VibrationEffects to do its job.

> https://developer.android.com/reference/android/view/HapticFeedbackConstants.html is basically why I wrote it the way I did. :-)

I could try but I'm not sure if the haptic feedback will propagate to the root view on most devices. Do you see any advantage using that?.

Frenzie:
> No with the vibration api.

Right, but using the vibration API is not "standardized" and requires permissions. Of course it uses the same methods underneath but it's an abstraction that takes care of system-wide consistency.

> I could try but I'm not sure if the haptic feedback will propagate to the root view on most devices. Do you see any advantage using that?.

The manufacturer will have made the constants right for the hardware. Or even if they didn't because they're stupid, they'll be odd in a manner consistent with the rest of the OS. Then again, the Zen UI keyboard has vibration settings possibly fully separate from the OS…

In theory, I think one shouldn't want to play around with the Vibration stuff directly except to do something special, like force feedback for a game. Of course it could be that just using a NativeSurfaceView is "special," which would be item 9999 on the list of things that are unfortunate about Android. ;-)

Also, your own quote is the strongest argument of all:
> Needs to be tweaked in Android Oreo and higher because default intensity is to much to be considered haptic feedback.

pazos:
@Frenzie: you were right, haptic feedback is the way to go. I leave this PR open until we merge the other one.

PR #192, opened 2019-09-30
haptic feedback support

pazos:
Improvement over #190 

I tested this on Moto G5S with great results. Probably needs to be tested in a device with a supported freescale e-ink screen and an haptic engine it something like that already exists.

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -323,6 +323,17 @@ enum {
     ATIMEOUT_SYSTEM = 0,
 };
 
+enum {
+    AHAPTIC_LONG_PRESS = 0,
+    AHAPTIC_VIRTUAL_KEY = 1,
+    AHAPTIC_KEYBOARD_TAP = 3,
+    AHAPTIC_CLOCK_TICK = 4,
+    AHAPTIC_CONTEXT_CLICK = 6,
+    AHAPTIC_KEYBOARD_RELEASE = 7,
+    AHAPTIC_VIRTUAL_KEY_RELEASE = 8,
+    AHAPTIC_TEXT_HANDLE_MOVE = 9,
+};
+
 int32_t AInputEvent_getType(const AInputEvent* event);
 int32_t AInputEvent_getDeviceId(const AInputEvent* event);
 int32_t AInputEvent_getSource(const AInputEvent* event);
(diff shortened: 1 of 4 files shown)


assets/android.lua
```diff
@@ -323,6 +323,12 @@ enum {
     ATIMEOUT_SYSTEM = 0,
 };
 
+enum {
+    AHAPTIC_LONG_PRESS = 0,
+    AHAPTIC_KEYBOARD_TAP = 3,
```
Frenzie: These are just the few I implemented as a kind of demo. It's probably simpler to just add the whole lot of them At the very `TEXT_HANDLE_MOVE` might potentially be useful in the future.

assets/android.lua
```diff
@@ -1475,6 +1481,17 @@ local function run(android_app_state)
...
     end
 
+    android.hapticFeedback = function(type)
+        android.DEBUG("haptic feedback: "..type)
+        JNI:context(android.app.activity.vm, function(JNI)
+            JNI:callVoidMethod(
+                android.app.activity.clazz,
+                "performHapticFeedback",
+                "(I)V",
+                ffi.new('int32_t', type)
```
Frenzie: This is always correct and not platform specific?

assets/android.lua
pazos: TEXT_HANDLE_MOVE was introduced in API 27. Most of the others were introduced in recent apis too.

I don't think using an unrecognized int will trigger an exception but I cannot test as the call is never triggered in the emulator and I currently don't have any working phone on any api below 28.

Shall I add the FLAG_IGNORE_GLOBAL_SETTING and FLAG_IGNORE_VIEW_SETTING?. These settings are useful to override (haptic) system settings.

assets/android.lua
pazos: Yeah, it calls https://github.com/koreader/android-luajit-launcher/pull/192/files#diff-272d98f2e8dcd4377fdca0ca2a53178aR181.

That function gets a view and calls https://github.com/koreader/android-luajit-launcher/pull/192/files#diff-272d98f2e8dcd4377fdca0ca2a53178aR194.

Finally view.performHapticFeedback(constant) will run only if the device has an haptic feedback engine and the setting wasn't disabled for that specific view or globally.

pazos:
> No comment on the other commits.

I will strip the last commit and submit it on another PR. I was testing it :p

assets/android.lua
Frenzie: I mean like 32-bit vs 64-bit. ;-)

pazos: oh! Based on this we're good: https://stackoverflow.com/a/15005226

assets/android.lua
Frenzie: > I don't think using an unrecognized int will trigger an exception but I cannot test as the call is never triggered in the emulator and I currently don't have any working phone on any api below 28.

Well, you could test something even higher. ;-) I could take a look some other time what happens on Android 6/23 in any case, but let's not let that hold up this PR.

> Shall I add

Sure, let's just add those and call it a day for now.

pazos: ok, We don't need FLAG_IGNORE_VIEW_SETTING as we're a single view activity.

PR #193, opened 2019-10-03
windowManager instead of wakelocks directly, restore WRITE_SETTINGS for custom timeouts

pazos:
~~:cry: do not merge yet.~~

WRITE_SETTINGS !? - feel free to improve it on the future using screenOn and some sort of the autosuspend plugin that disable screenOn after a certain time.

Safer way to keep the screen on, untested on problematic devices, like the Sony RP-1?

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -3,12 +3,12 @@
     xmlns:android="http://schemas.android.com/apk/res/android"
     package="org.koreader.launcher" >
 
-    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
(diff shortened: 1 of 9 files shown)


pazos:
Sorry @Frenzie but I restored the WRITE_SETTINGS permission and the old and trusty? way of setting  custom timeouts. I'm still thinking is not the best option but works fine, I use it and I don't want to struggle my head around it.

pazos:
> If it's just the old method then fine by me, or should I have to look at it in closer detail? Looks fine skimming.

It is almost the same, but I moved everything related to system settings to a new object "SystemSettings.kt".

Also, instead of auto-magically request WRITE_SETTINGS the new behaviour relies on a frontend implementation which should give users more context:

writeSettings

Frenzie:
The message looks alright. What kind of warning should it give?

PR #194, opened 2019-10-09
fix clipboard

pazos:
see https://github.com/koreader/koreader/pull/5454#issuecomment-539870395 and https://github.com/koreader/koreader/pull/5454#issuecomment-540055850


app/src/org/koreader/launcher/BaseActivity.kt
```diff
@@ -177,59 +181,49 @@ abstract class BaseActivity : NativeActivity(), JNILuaInterface,
...
-                        if (text != null) {
-                            result.value = text.toString()
-                        } else {
-                            result.value = ""
-                        }
+        runOnUiThread {
+            try {
+                val data = clipboard.primaryClip
+                if (data != null && data.itemCount > 0) {
+                    val text = data.getItemAt(0).coerceToText(
```
Frenzie: What a function name. :-P (Not new, I just forgot about it.)

PR #196, opened 2019-10-18
add aarch64 support

pazos:
Add a new abi, called arm64-v8a by Google, based on ARMv8a instruction set architecture. This is 64bits, FPU and SIMD/NEON or whatever the hell that means.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/196)

### DIFF
--- a/jni/luajit/mk-luajit.sh
+++ b/jni/luajit/mk-luajit.sh
@@ -2,10 +2,11 @@
 # see http://luajit.org/install.html for details
 # there, a call like one of the following is recommended
 
-# NDKABI=21  # Android 5.0+
-# NDKABI=19  # Android 4.4+
-NDKABI=${NDKABI:-14} # Android 4.0+
-#NDKABI=${NDKABI:-9} # Android 2.3+
+# We use NDK r15c for all architectures.
+
+# For 32 bits (armeabi-v7a and x86) we built against platform-14 (ICS)
+# For 64 bits (arm64-v8a and x86_64) we built against platform-21 (Lollipop)
+
 DEST=$(cd "$(dirname "$0")" && pwd)/build/$1
 # might be linux-x86_64 or darwin-x86-64
 HOST_ARCH="*"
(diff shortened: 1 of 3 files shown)


jni/luajit/mk-luajit.sh
```diff
@@ -62,6 +62,15 @@ case "$1" in
         NDKARCH="-march=armv7-a -mfloat-abi=softfp -Wl,--fix-cortex-a8"
         make -C luajit-2.0 install HOST_CC="gcc -m32" CFLAGS="-O2 -pipe" HOST_CFLAGS="-O2 -pipe -mtune=generic" LDFLAGS="" HOST_LDFLAGS="" TARGET_CFLAGS="${CFLAGS}" TARGET_LDFLAGS="${LDFLAGS}" CROSS="$NDKP" TARGET_FLAGS="${NDKF} ${NDKARCH}" TARGET_SYS=Linux DESTDIR="$DEST" PREFIX=
         ;;
+    arm64-v8a)
+        # Android/ARM, arm64-v8a (ARM64 VFP4, NEON)
+        check_NDK
+        TCVER=("${NDK}"/toolchains/aarch64-linux-android-4.*)
+        NDKP=${TCVER[0]}/prebuilt/${HOST_ARCH}/bin/aarch64-linux-android-
+        NDKF="--sysroot ${NDK}/platforms/android-21/arch-arm64"
+        NDKARCH="-DLJ_ABI_SOFTFP=0 -DLJ_ARCH_HASFPU=1 -DLUAJIT_ENABLE_GC64=1"
```
Frenzie: In principle this doesn't seem like something that should be here?

NiLuJe: Yup, it should roughly be set to the `ANDROID_ARM_ARCH` value computed in `Makefile.defs`.

c.f., https://github.com/koreader/koreader-base/issues/993#issuecomment-543777713

jni/luajit/mk-luajit.sh
```diff
@@ -62,6 +62,15 @@ case "$1" in
...
+        make -C luajit-2.0 install HOST_CC="gcc -m64" CFLAGS="-O2 -pipe" HOST_CFLAGS="-O2 -pipe -mtune=generic" LDFLAGS="" HOST_LDFLAGS="" TARGET_CFLAGS="${CFLAGS}" TARGET_LDFLAGS="${LDFLAGS}" CROSS="$NDKP" TARGET_FLAGS="${NDKF} ${NDKARCH}" TARGET_SYS=Linux DESTDIR="$DEST" PREFIX=
```
NiLuJe: In the same vein, we technically don't need the `-m64` here. We assume as a given everywhere else that our build host is `x86_64` ;).

jni/luajit/mk-luajit.sh
pazos: It seems that `-DLUAJIT_ENABLE_GC64=1` was needed until https://github.com/LuaJIT/LuaJIT/commit/bd00094c3b50e193fb32aad79b7ea8ea6b78ed25

Just found that lurking, didn't touch this PR since october.

NiLuJe: Ooh, new LuaJIT commits, shiny :D.

NiLuJe: Possibly relevant to this PR: https://github.com/LuaJIT/LuaJIT/commit/08272de59eedaa82726b8f671dccc7e0170addc0

pazos: Indeed, with new luajit these changes are not needed anymore.

pazos:
Updated and tested for both arm64-v8a and x86_64.

We can use the same NDK but we need to use platform-21, which is the first one to be supported by both ABIs. That also makes 21 the minimum Sdk where these APKs run.

Ofc, I tested this with `make example` for both ABIs. Didn't test anything KO related or loading shared libraries with elf loader, so these things might be missing.

pazos:
@Frenzie @NiLuJe: this one is ready to go too!

Frenzie (approved):
lgtm

PR #200, opened 2019-12-02
Generic intents

pazos:
ACTION_SEND, ACTION_SEARCH and ACTION_PROCESS_TEXT when called without an explicit package will be opened with the application picker if there is more than one alternative to open them.

Also fixed some warnings thrown by Android Studio linter.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/200)


pazos:
The rationale is in https://www.mobileread.com/forums/showpost.php?p=3922858&postcount=9 and it is motivated by some wonky apps, like the ones in https://www.mobileread.com/forums/showpost.php?p=3922499&postcount=6

PR #201, opened 2019-12-10
initial support for scoped storage

pazos:
It includes a minor fix to build ok against targetApi 29

Scoped storage won't affect android 10 unless we use targetApi 29. Current targetApi is still untouched, using level 28.

Basically it replaces storage_path on versions where [getExternalStorageDirectory()](https://developer.android.com/reference/android/os/Environment#getExternalStorageDirectory()) is deprecated, with a known path where RW permissions are granted.

From Android 11 onwards everything outside that directory needs to be accesed using SAF.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/201)

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -27,6 +27,12 @@
             android:resizeableActivity="false"
             android:launchMode="singleInstance"
             android:theme="@style/SplashScreen">
+            <!--add android:requestLegacyExternalStorage="true" when bumping targetApi to 29+
+
+                This will allow upgrades from previous versions keeping the legacy storage model.
+                New installations won't be affected and will use scoped storage instead on
+                compatible devices. -->
+
             <meta-data android:name="android.app.lib_name" android:value="luajit" />
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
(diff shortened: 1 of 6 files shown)


pazos:
After reading https://commonsware.com/blog/2019/06/07/death-external-storage-end-saga.html I think we can bump to api29, add the flag android:requestLegacyExternalStorage="true" and determine if the device has scoped storage using Environment.isExternalStorageLegacy().

That would make code nicer than the ugly API if-else check this PR implemented. Of course that would mean bump the buildsystem.

pazos:
> After reading https://commonsware.com/blog/2019/06/07/death-external-storage-end-saga.html I think we can bump to api29, add the flag android:requestLegacyExternalStorage="true" and determine if the device has scoped storage using Environment.isExternalStorageLegacy().
> 
> That would make code nicer than the ugly API if-else check this PR implemented. Of course that would mean bump the buildsystem.

Just tried. When building with targetApi 29 the flag android:requestLegacyExternalStorage="true" just affect updates, not new installations, and Environment.isExternalStorageLegacy() will return exactly that: false for new installations and true for upgrades.

So, instead of trying to guard the behaviour on android 11 I went with a single boolean to define what's the path. This will allow us to test scoped storage even on lower android versions, while keeping everything as they are for all the apis.

I added a few comments about the changes needed for bumping to api29. There's no reason to upgrade for now. We're actually targeting api28 just to support devices with screen notches/cutouts.

PR #203, opened 2019-12-13
import files from SAF

### LINKED ISSUE #191: add Storage Access Framework support in 4.4+
Specially for 6.0+ devices, that can allow RW permission permanently on the sdcard.

pazos:
Fixes #191.

**Edit:**

- Adds a function to import one or multiple files using Android SAF. This requires a minimum api of 19.
- Adds a "jailed" flavor which runs w/ targetApi 29 and minimumApi 19. Can be built with `JAILED=1` 
- Adds a number to the beginning of version code to differenciate between archs: 1 for arm and 2 for x86.
- Update supported mimetypes for both open files from intent and import files through SAF
- Relax a bit detekt long method warning, because it triggers with our simple safFilePicker function just because the "huge" array of mimetypes.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/203)


pazos:
Ok, with the updated PR is possible to request a build with scoped storage using `SCOPED=1 ./kodev release android`

I also changed version code to take arch into account, so for arm it will be 1 .. commits and for x86 2 .. commits. Google requires it for releasing in the app store and I think Fdroid uses that too. Maybe @Rudloff can confirm that.

pazos:
@Frenzie: please review when you have a bit of time. Specially the arch dependant version code. We can leave it out as is not strictly related to this PR.

Frenzie:
___
*[app/src/org/koreader/launcher/MainActivity.kt, line 263 at r2](https://reviewable.io/reviews/koreader/android-luajit-launcher/203#-LwPdFKb9UX-yHL0IawE:-LwPoQHBCsy8oNh0_weA:b-apdvfg) ([raw file](https://github.com/koreader/android-luajit-launcher/blob/1da9dcf324c182c4b2b680a3584e52d4aa694e78/app/src/org/koreader/launcher/MainActivity.kt#L263)):*
<details><summary><i>Previously, pazos (Martín Fernández) wrote…</i></summary><blockquote>

Yup, it is based on the intent-filter in AndroidManifest.xml. I will update both.

But "application/vnd.openxmlformats-officedocument.wordprocessingml.document", seriously? 
</blockquote></details>

The frontend code just works by extension really; those mimetypes are just window dressing.

Anyway, see here:
https://blogs.msdn.microsoft.com/vsofficedeveloper/2008/05/08/office-2007-file-format-mime-types-for-http-content-streaming-2/

app/src/org/koreader/launcher/MainActivity.kt
```diff
@@ -262,15 +262,27 @@ class MainActivity : BaseActivity() {
...
+                "application/fb3",
                 "application/msword",
                 "application/pdf",
+                "application/rtf",
+                "application/tcr",
+                "application/vnd.amazon.mobi8-ebook",
+                "application/vnd.ms-htmlhelp",
+                "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+                "application/vnd.palm",
                 "application/x-cbz",
```
Frenzie: I think you forgot cbt, and it might be worth to take another look at PdfDocument to see if there's more there. ;-)

Rudloff:
> I also changed version code to take arch into account, so for arm it will be 1 .. commits and for x86 2 .. commits. Google requires it for releasing in the app store and I think Fdroid uses that too. Maybe @Rudloff can confirm that.

Indeed, we require every APK to have a different version code.

pazos:
> Indeed, we require every APK to have a different version code.

With this PR the version code specified in koreader-android-fdroid-latest does not match the one in the app but still uses the number of commits.  Ie: for version code 7204 we're generating a version for arm -> 17204 and a version for x86 -> 27204.

pazos:
@Frenzie: I see a pending review on reviewable so I'm not sure if I should merge this yet.

pazos:
f**k reviewable!!!! :cry:

PR #207, opened 2019-12-28
merge EinkTest in the main app

pazos:
Can be called using `android.einkTest()` or from commandline with `adb shell am start -n org.koreader.launcher/.EPDTestActivity`

Migrated EPD abstract classes to kotlin, so we are 100% kotlin, 0% java. 

I barely tested that it gives the same exceptions as before, because I don't have an eink device. It must be checked before merged, specially the RK30xx class which was never touched.

Not sure where to place this option, for this example I went with developer options:

test (1)

As you can see it repurposes the name (and logo in older versions) from the main activity.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/207)


pazos:
Hey @Cellaris, @hugleo: could you please test this apk: https://www.dropbox.com/s/joh083miewf0v8s/koreader-arm-test.apk?dl=0

You can try to open the test activity as shown in the gif above but my main concern is if the full refresh still happens each x pages.

There's no need to test on different devices: just with an rk3026 and a rk3368 we're fine.
Thanks for the feedback!

Cellaris:
On the Onyx Boox C67 it works perfectly.

pazos:
Awesome, thank you very much :smile:

hugleo:
Also working for rk33xx

Frenzie:
Are we sure it's actually accessible this way? (On account of screen update issues.)

pazos:
> Are we sure it's actually accessible this way? (On account of screen update issues.)

From adb is always available. I will update instructions on https://github.com/koreader/koreader/issues/4551#issue-406883935 when this is merged.

PR #208, opened 2019-12-31
resolve file paths delivered as content without importing the file

pazos:
Tested with Simple Mobile Tools FM, Root Explorer and Solid Explorer.

It could lead to a false positive if the intent.VIEW contains data other than a file (a pipe, a socket). A false positive would start KOReader in the FM (or whatever the user configured) so it shouldn't hurt.

While this workaround still works on scoped storage devices is totally useless because we won't be able to read the file from the resolved path.

Some background in https://www.mobileread.com/forums/showpost.php?p=3932733&postcount=14 and following posts.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/208)

### DIFF
--- a/app/manifest/base.xml
+++ b/app/manifest/base.xml
@@ -35,6 +35,7 @@
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
                 <data android:scheme="file" />
+                <data android:scheme="content" />
                 <data android:mimeType="application/epub+zip" />
                 <data android:mimeType="application/fb2" />
                 <data android:mimeType="application/fb3" />
--- a/app/src/org/koreader/launcher/FileUtils.kt
+++ b/app/src/org/koreader/launcher/FileUtils.kt
@@ -9,7 +9,10 @@ import java.util.Locale
 import android.content.ContentResolver
 import android.content.Context
 import android.net.Uri
+import android.os.Build
+import android.os.ParcelFileDescriptor
 import android.provider.MediaStore
+import android.system.Os
 
 internal object FileUtils {
     const val TAG = "FileUtils"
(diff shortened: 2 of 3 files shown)


Frenzie:
___
*[app/src/org/koreader/launcher/FileUtils.kt, line 61 at r1](https://reviewable.io/reviews/koreader/android-luajit-launcher/208#-LxQ6-QyCJqK8rqXHf7N:-LxQ6-QyCJqK8rqXHf7O:b9dho83) ([raw file](https://github.com/koreader/android-luajit-launcher/blob/d9beb76466d466c1078026a68ebe96391b510710/app/src/org/koreader/launcher/FileUtils.kt#L61)):*
> ```Kotlin
>      * If the data in the uri is not a file this will fail
>      *
>      * @param pfd - parcelable file descriptor from contentResolver.openFileDescriptor
> ```

Is that supposed to say parsable? ;-)

pazos:
> Is that supposed to say parsable? ;-)

No no, it is supposed to say parcelable (or parcel?) because a [parcelFileDescriptor](https://developer.android.com/reference/android/os/ParcelFileDescriptor) is just an object that implements the [Parcelable interface](https://developer.android.com/reference/android/os/Parcelable.html).

A [parcel](https://developer.android.com/reference/android/os/Parcel.html) is used on android for IPC, via the android binder.

These low level api calls were never intended to deal directly with common files and the entire reason for some thing like a parcelFileDescriptor is to close the pipe or socket used by two processes for intercommunication.  It just happens that, in the context of a filtered action.VIEW of a few mimetypes, the fileDescriptor can be used to retrieve the real name of a file. And a file that is not a common file will throw an IOException.

So, in short: it is a hack. I'm using it since https://github.com/koreader/android-luajit-launcher/pull/127 (the legacy behaviour just didn't work) but I was following google/commonsware best practices, resumed in https://github.com/koreader/koreader/issues/4441#issuecomment-469054512.

In the end I figured out that MoonReader, Calibre Companion and other applications are using these kind of hackish approachs to circumvent some of google restrictions. We can do that too.

But, generally speaking, this workaround won't work for most SAF enabled work, so in that context we need to use [FileUtils.saveAsFile](https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/FileUtils.kt#L28-L30) instead.

Frenzie:
I see. (Or perhaps rather: whyyyyyyyyyy D:)

PR #210, opened 2020-01-02
wip: android tv support

pazos:
it works fine with the example app.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/210)


pazos:
Flavors can be called as we like but the "leanback" thingy and the banner are just for android tv.

The reason for different flavors is the lack of rotation support in KO. If we manage to add rotation support then we can use a merged manifest with both uses-features required false.

Frenzie:
Lack of rotation support? I guess you mean some Android-specific detail but it should be pretty simple to make it behave "properly." The only problem is that I have negative interest in making the experience what I consider much worse. ;-) Android rotation is unbelievably atrocious, except in some versions of Cyanogenmod. The tl;dr is that auto-rotation is an absolute disaster, and Android's choice between auto and portrait is just plain awful.

In Cyanogenmod you could quickly switch between auto & lock, as well as manual landscape/portrait. The obvious, sensible thing to do. Not some must-have feature for which you should install a freaking custom ROM.

/rant

pazos:
> Lack of rotation support? I guess you mean some Android-specific detail but it should be pretty simple to make it behave "properly." The only problem is that I have negative interest in making the experience what I consider much worse. ;-)

orientation on android means a lot of things. See how it works on a tablet with sensorPortrait. Surface is rotated, input coordinates too. 
Supporting orientation means react dynamically on surface changes, adapting the buffer to the new dimensions.

> Android rotation is unbelievably atrocious, except in some versions of Cyanogenmod. The tl;dr is that auto-rotation is an absolute disaster, and Android's choice between auto and portrait is just plain awful.

I agree.  Autorotation is a mess, even more on a document viewer!.

Frenzie:
> Supporting orientation means react dynamically on surface changes, adapting the buffer to the new dimensions.

It's not clear to me where you see an obstacle. :-)

As a big aside, adapting to new dimensions shouldn't be too hard. One could simply adapt the blueprint I wrote for SDL in https://github.com/koreader/koreader-base/pull/614. However, that's not relevant to orientation changes, only to plugging in a monitor using HDMI, or dynamically resizing a [window](https://developer.android.com/guide/topics/ui/multi-window). Resizing is certainly not handled perfectly, but I fixed the worst issues two years ago.

Back to the main point, we'd want to fully handle orientation changes ourselves along the lines described here:
https://developer.android.com/guide/topics/resources/runtime-changes#HandlingTheChange
Or in other words, do nothing at all except forwarding the new orientation to Lua.

The rest of the code should be quite simple, and better yet it @NiLuJe already wrote it out recently:
https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/device/input.lua#L668-L711 With some trickery we could even reuse that very function, but I don't know if that's a good idea. That's besides the point though. ;-)

> I agree. Autorotation is a mess, even more on a document viewer!.

Btw, I use [Tap 'n' Turn](https://gabm.github.io/TapAndTurn/). Not as good as CyanogenMod but it's about the closest you can get to sane behavior without custom ROMs.

pazos:
> It's not clear to me where you see an obstacle. :-)

Not an obstacle but I don't have knowledge about how KOReader handles the blitbuffer and trying to do things blindly on android is quite frustrating because you're running two different programs.

> As a big aside, adapting to new dimensions shouldn't be too hard. One could simply adapt the blueprint I wrote for SDL in koreader/koreader-base#614. However, that's not relevant to orientation changes, only to plugging in a monitor using HDMI, or dynamically resizing a window. Resizing is certainly not handled perfectly, but I fixed the worst issues two years ago.

Yeah, help with that would be awesome. Maybe a function to resize the screen.

> Back to the main point, we'd want to fully handle orientation changes ourselves along the lines described here:
https://developer.android.com/guide/topics/resources/runtime-changes#HandlingTheChange
Or in other words, do nothing at all except forwarding the new orientation to Lua.

We **must** use it because the activity is not intended to be stopped/resumed by configuration changes. 

I think this [activity callback](https://github.com/koreader/android-luajit-launcher/blob/master/jni/android_native_app_glue/android_native_app_glue.c#L438) will be triggered each time the surface changes its size. Is just a matter of resizing the blitbuffer there.

pazos:
https://github.com/koreader/koreader-base/pull/1128 makes possible this thing again!

Apparently we need to take control of the native window from the java side (like we do on some ntx boards). Otherwise CRE documents are render equally fine but the screen goes black after a few seconds without overlay (not in app overlay, but an android UI thingy, like volume control).

PR #217, opened 2020-01-06
Disable large heap for managed (java) memory and add some tools to debug builds to detect issues.

pazos:
VM large heap flag isn't needed AFAICT. In the days android was introduced it was hardcoded to 16MB for "normal" apps and a bit more for apps that use the large heap flag. Nowadays most modern devices report the same value with or without the flag: 384MB~.

the large heap flag is useful for activities that load a bunch of java libraries or that have big drawable objects . That's not our case. We run with about 2-3mb of -vm managed- heap usage and almost 0 allocations.

Anyways, I added a function to retrieve that info and log it.

The other changes: move dlopen to their own log tag.
Log policy violations on debug builds
Add leakCanary to debug builds

So, a catch-em-all log will be: `pidcat org.koreader.launcher -t KOReader -t dlopen -t LeakCanary -t StrictMode` while `adb logcat KOReader:* *:S` is related to the activity lifecycle and normal logs.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/217)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -5,9 +5,9 @@ apply plugin: 'kotlin-android-extensions'
 android {
     defaultConfig {
         applicationId "org.koreader.launcher"
-        minSdkVersion rootProject.ext.minSdk
-        targetSdkVersion rootProject.ext.targetSdk
-        compileSdkVersion rootProject.ext.targetSdk
+        minSdkVersion minSdk
+        targetSdkVersion targetSdk
+        compileSdkVersion targetSdk
         versionCode versCode as Integer
         versionName versName
     }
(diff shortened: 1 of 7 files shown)


Frenzie:
___
*[app/src/org/koreader/launcher/MainApp.kt, line 39 at r1](https://reviewable.io/reviews/koreader/android-luajit-launcher/217#-LxwH_-T7ymVZW_m9cBU:-LxwH_-UChUpAip374zq:b-vtbs7y) ([raw file](https://github.com/koreader/android-luajit-launcher/blob/cb0288fa2b9c07f898edc2642e8d1cf4b41138c2/app/src/org/koreader/launcher/MainApp.kt#L39)):*
> ```Kotlin
>         if (debuggable) {
>             Log.d(name, "StrictMode will detect all potential violations and log them")
>             val threadPolicy = StrictMode.ThreadPolicy.Builder().detectAll().penaltyLog().build()
> ```

Wow, that's quite the line.

PR #218, opened 2020-01-11
wip: send/share files

pazos:
Allow to send binary files via bluetooth, social apps or any other app that supports a given mimetype.

WIP: it uses a workaround, works fine **only** if the receiver has filesystem permissions to reach the file. A proper implementation would be written a custom fileProvider and share as content instead.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/218)


pazos:
Leaving this feature for myself, for the moment. Can be implemented on the future if somebody requests it :+1:

PR #221, opened 2020-01-30
replace WifiManager with ConnectivityManager

pazos:
Sample solution for https://github.com/koreader/koreader/issues/5787
Send to android wifi settings instead of dealing with wifi state. Update permissions.

Should work on 99'9999% of android devices, but should be tested against corner cases (aka 4.0.? e-ink devices)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/221)

### DIFF
--- a/app/manifest/base.xml
+++ b/app/manifest/base.xml
@@ -4,8 +4,7 @@
     package="org.koreader.launcher" >
 
     <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE"/>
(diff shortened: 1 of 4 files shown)


pazos:
ok, I was playing a bit with [gnirehtet](https://github.com/Genymobile/gnirehtet). When the traffic is routed via VPN it just reports that and makes isWifiOn false even when wifi is already enabled.

I think we might want to report isNetworkOn instead of isWifiOn. If no network available prompt for wifi connection.

Frenzie:
I thought we were already technically just going for connectivity? If not, checking specifically for wifi seems a bit odd. :-)

PR #235, opened 2020-07-05
Basic support for Android Tv and Chrome OS

pazos:
~~Also basic multiwindow support (used on android split view and chrome OS).~~
There's no need to build multiple APKs as platform detection happens at runtime.

Superseedes #210

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/235)


pazos:
BTW I installed Android Studio 4 and it doesn't suck. That's a lot for an IDE.

Frenzie:
Oh yeah, it's fairly speedy. Besides that it seems roughly the same? ^_^

pazos:
> Oh yeah, it's fairly speedy. Besides that it seems roughly the same? ^_^

The profiler is a bit revamped. It also has some layout goodies that we can't use :p

pazos:
BTW, I'm going to remove multiwindow support from this PR. It is harder than I thought and too much burder at this moment. I will open a separate issue to keep it tracked.

PR #236, opened 2020-07-07
Update LuaJIT

NiLuJe:
re https://github.com/koreader/koreader-base/pull/1133

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/236)

### DIFF
--- a/jni/luajit/mk-luajit.sh
+++ b/jni/luajit/mk-luajit.sh
@@ -40,6 +40,8 @@ function check_NDK() {
     fi
 }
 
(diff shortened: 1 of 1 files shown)


Frenzie:
@pazos Not necessarily, the patch made it so that the x86 build actually worked on my x86 ASUS Zenfone 2. But if you just get it into a nightly build I can check on it. (I could also build it myself but not right now anyway.)

Frenzie:
@pazos But shouldn't it just need a minor update of a few lines?

pazos:
> But shouldn't it just need a minor update of a few lines?

Most of things patched are no longer there. I didn't take a deep look. Just rebased #235 and saw the issue.

NiLuJe:
Ah. Forgot the patch ;p. It's updated in the base pr, I just forgot it hero;).

PR #239, opened 2020-07-15
light device abstraction and native frontlight dialog

pazos:
use a native alertDialog to change brightness (and warmth on supported devices).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/239)

### DIFF
--- a/app/src/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/org/koreader/launcher/device/DeviceInfo.kt
@@ -2,15 +2,15 @@
  * based on https://github.com/unwmun/refreshU
  *
  * Note: devices don't need to be declared here unless
- * they have known e-ink update routines and/or bug workarounds. */
+ * they have known e-ink update routines, custom light settings  and/or bug workarounds. */
 
 package org.koreader.launcher.device
 
 import android.os.Build
-import java.util.*
+import java.util.Locale
+import kotlin.collections.HashMap
 
 object DeviceInfo {
-
     val PRODUCT: String
     val EINK_FREESCALE: Boolean
     val EINK_ROCKCHIP: Boolean
@@ -18,6 +18,7 @@ object DeviceInfo {
     val EINK_FULL_SUPPORT: Boolean
     val BUG_WAKELOCKS: Boolean
     val BUG_SCREEN_ROTATION: Boolean
+    val NEEDS_VIEW: Boolean
 
     private val MANUFACTURER: String
     private val BRAND: String
@@ -38,6 +39,7 @@ object DeviceInfo {
     private val ENERGY: Boolean
     private val INKBOOK: Boolean
     private val TOLINO: Boolean
+    private val TOLINO_EPOS: Boolean
     private val NOOK_V520: Boolean
     private val SONY_RP1: Boolean
     private val EMULATOR_X86: Boolean
@@ -45,6 +47,7 @@ object DeviceInfo {
 
     // default values for generic devices.
     internal var EINK = EinkDevice.UNKNOWN
+    internal var LIGHTS = LightsDevice.NONE
     private var BUG = BugDevice.NONE
 
     enum class EinkDevice {
@@ -67,6 +70,11 @@ object DeviceInfo {
         NOOK_V520
     }
 
+    enum class LightsDevice {
+        NONE,
+        TOLINO_EPOS
+    }
+
     enum class BugDevice {
         NONE,
         SONY_RP1,
@@ -84,6 +92,7 @@ object DeviceInfo {
         // --------------- device probe --------------- //
         val deviceMap = HashMap<EinkDevice, Boolean>()
         val bugMap = HashMap<BugDevice, Boolean>()
+        val lightsMap = HashMap<LightsDevice, Boolean>()
 
         // Boyue T62, manufacturer uses both "boeye" and "boyue" ids.
         BOYUE_T62 = (IS_BOYUE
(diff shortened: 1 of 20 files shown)


pazos:
Oops, looks like a mess because it is a mess. Will fix that later. Now just pushing this to document the new interface for https://github.com/koreader/koreader/pull/6332

app/src/org/koreader/launcher/device/DeviceInfo.kt
```diff
@@ -38,15 +41,18 @@ object DeviceInfo {
     private val ENERGY: Boolean
     private val INKBOOK: Boolean
     private val TOLINO: Boolean
+    private val TOLINO_EPOS: Boolean
```
zwim, line 44: Shouldn't it be called TOLINO_EPOS_2, as there is also a "Tolino Epos" reader, which also has natural light.
The "Epos" is different from the "Epos 2".

pazos: Better yet. We can have both as true under a single name. After all setting that value to true just means that the device has a "TolinoWarmthController".

zwim:
When I add
```------------- frontend/ui/elements/common_settings_menu_table.lua -------------
index 4abaeec9..037c92ed 100644
@@ -201,6 +201,14 @@ if Device:isAndroid() then
     local isAndroid, android = pcall(require, "android")
     if not isAndroid then return end
 
+    -- on Android: overwrite generic frontlight with a native Dialog 
+    common_settings.frontlight = {
+        text = _("Frontlight"),
+        callback = function()
+            android.settings.dialog(_("Frontlight settings"), _("Intensity"), _("Warmth"), _("Ok"))
+            end,
+    }
+
     -- screen timeout options, disabled if device needs wakelocks.
     common_settings.screen_timeout = require("ui/elements/screen_android"):getTimeoutMenuTable()
```
I get the native android dialog. It is working on my mobile and on the Epos2.

Touch gestures for in/decreasing intensity works on the Epos2.

BUT on the phone: With the touch gestures I get the popups "Frontlight disabled"/"Frontligh enabled". (Even with the original dialog). No change in intensity.

app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
```diff
@@ -0,0 +1,95 @@
...
+/* Special controller for Tolino Epos/Epos2.
+ * see https://github.com/koreader/koreader/pull/6332
+ *
+ * Thanks to @zwim
+ */
+
+class TolinoWarmthController : LightInterface {
+    companion object {
+        private const val TAG = "lights"
+        private const val BRIGHTNESS_MAX = 100
```
zwim, line 21: It has to be 
BRIGHTNESS_MAX = 255
to get the full intensity.

pazos:
> When I add ...

@zwim: this is the backend part. The things you're talking about are frontend. Lets move conversation to https://github.com/koreader/koreader/pull/6332.

app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
pazos: Uh, ok. I based this one on your original report:

> The brightness and warmth values have to be in the range 0-100, otherwise the slider of the warmth is broken. I don't know why.

But that can be changed easily.

zwim:
So far the frontend and backend work together, very well. Thank you too.

zwim:
@pazos: You can find the patch with the changes for the blocking flag superseding the changes in your PR in 
https://www.dropbox.com/s/wwqlp1j72x5jq64/patch-for-239-with-blocking.patch?dl=0

This is a patch from my `work/koreader/platform/android/luajit-launcher`. I hope you can work with that. It was created with `git format-patch -1 ...`

pazos:
> I hope you can work with that.

Sure, will cherry pick this weekend. Thank you!

pazos:
@zwim: I added you as coauthor and now I can't request you a review using github :/

Please review anyways!

detekt.yml
```diff
@@ -115,7 +115,7 @@ empty-blocks:
   EmptyForBlock:
     active: true
   EmptyFunctionBlock:
-    active: true
+    active: false
```
Frenzie, line 118: Too bothersome? :-)

pazos: Yeah, it forces us to comment on emtpy seekbar methods  with `// it does nothing`

assets/android.lua
```diff
@@ -1575,6 +1575,35 @@ local function run(android_app_state)
...
+                "showFrontlightDialog",
+                "(Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;)I",
+                t, i, w, o
+            )
+            jni.env[0].DeleteLocalRef(jni.env, t)
+            jni.env[0].DeleteLocalRef(jni.env, i)
+            jni.env[0].DeleteLocalRef(jni.env, w)
+            jni.env[0].DeleteLocalRef(jni.env, o)
+        end)
+    end
```
pazos, line 1596: > Do you want an empty line here?

I certainly do!. Thanks for pointing it out!

I would also like to refactor the entire code in android.lua. Patch after patch it became in a mix of different naming conventions and coding styles. 

I think I will do it after this PR. I can ping you if you want to review it :)

zwim:
The changes look good. I have tested it on the Epos2. It works perfectly.

PR #242, opened 2020-07-22
handle all possible configuration changes

pazos:
To avoid `Engine thread destroy requested!`

I would say "handle" because we're doing nothing with them. Just telling android we don't need the application destroyed and recreated when the configuration changed.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/242)


Frenzie:
Weird. :-)

PR #244, opened 2020-08-17
add warmth for Tolino Vision 4

zwim:
The Vision 4 and the Epos 2 use the same DEVICE, BRAND and MODEL identifiers.
For accessing the warmth file, both use different files in /sys/class/backlight/

This PR allows changing warmth on both devices.

Internal: `setLastModified` does not work as expected (i.e. it does not throw an exception), therefore it is better to use `canWrite`

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/244)

### DIFF
--- a/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
+++ b/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
@@ -30,7 +30,7 @@ class TolinoWarmthController : LightInterface {
                 if (File(COLOR_FILE_VISION4HD).exists()) {
                     COLOR_FILE = COLOR_FILE_VISION4HD
                 }
-                else if (File(COLOR_FILE_EPOS2).exists()) {
+                else { // if (File(COLOR_FILE_EPOS2).exists())
                     COLOR_FILE = COLOR_FILE_EPOS2
                 }
                 Logger.v(TAG, "using $COLOR_FILE")
--- a/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
+++ b/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
@@ -22,10 +22,10 @@ class TolinoWarmthController : LightInterface {
         private const val MIN = 0
         private const val COLOR_FILE_EPOS2 = "/sys/class/backlight/tlc5947_bl/color"
         private const val COLOR_FILE_VISION4HD = "/sys/class/backlight/lm3630a_led/color"
-	private val COLOR_FILE = if (File(COLOR_FILE_VISION4HD).exists())
-	    COLOR_FILE_VISION4HD
-	else 
-	    COLOR_FILE_EPOS2
+        private val COLOR_FILE = if (File(COLOR_FILE_VISION4HD).exists())
+            COLOR_FILE_VISION4HD
+        else
+            COLOR_FILE_EPOS2
     }
 
     override fun hasFallback(): Boolean {
--- a/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
+++ b/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
@@ -22,7 +22,7 @@ class TolinoWarmthController : LightInterface {
         private const val MIN = 0
         private const val COLOR_FILE_EPOS2 = "/sys/class/backlight/tlc5947_bl/color"
         private const val COLOR_FILE_VISION4HD = "/sys/class/backlight/lm3630a_led/color"
-        private var COLOR_FILE = "" // gets initialized before first write access
+        private lateinit var COLOR_FILE : String // gets initialized before first write access
     }
 
     override fun hasFallback(): Boolean {
@@ -78,18 +78,18 @@ class TolinoWarmthController : LightInterface {
         }
 
         // on the first call, we check on which Tolino we are running
-        if (COLOR_FILE=="") { 
+        if (COLOR_FILE == "") {
             if (File(COLOR_FILE_VISION4HD).exists()) {
-                COLOR_FILE=COLOR_FILE_VISION4HD
+                COLOR_FILE = COLOR_FILE_VISION4HD
             }
             else if (File(COLOR_FILE_EPOS2).exists()) {
-                COLOR_FILE=COLOR_FILE_EPOS2
+                COLOR_FILE = COLOR_FILE_EPOS2
             }
             Logger.v(TAG, "using $COLOR_FILE")
         }
 
         val colorFile = File(COLOR_FILE)
-        Logger.v(TAG, "Setting warmth to $warmth")        
+        Logger.v(TAG, "Setting warmth to $warmth")
         try {
             if (!colorFile.canWrite()) {
                 Runtime.getRuntime().exec("su -c chmod 666 $COLOR_FILE")
--- a/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
+++ b/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
@@ -22,20 +22,10 @@ class TolinoWarmthController : LightInterface {
         private const val MIN = 0
         private const val COLOR_FILE_EPOS2 = "/sys/class/backlight/tlc5947_bl/color"
         private const val COLOR_FILE_VISION4HD = "/sys/class/backlight/lm3630a_led/color"
-        private lateinit var COLOR_FILE : String // gets initialized before first write access
-
-        fun initColorFile() {
-            // on the first call, we check on which Tolino we are running
-            if (!::COLOR_FILE.isInitialized) {
-                if (File(COLOR_FILE_VISION4HD).exists()) {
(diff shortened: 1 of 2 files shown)


app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
```diff
@@ -74,15 +76,24 @@ class TolinoWarmthController : LightInterface {
...
         }
-        val colorFile = File(COLOR_FILE)
-        Logger.v(TAG, "Setting warmth to $warmth")
-        try {
-            colorFile.setLastModified(System.currentTimeMillis())
-        } catch (e: Exception) {
-            Runtime.getRuntime().exec("su -c chmod 666 $COLOR_FILE")
+
+        // on the first call, we check on which Tolino we are running
+        if (COLOR_FILE=="") {
```
Frenzie, line 81: Codacy had a little complaint:
```suggestion
        if (COLOR_FILE == "") {
```

(also spaces around `=` on later lines)

Seems reasonable to me, pinging @pazos

app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
```diff
@@ -20,7 +20,9 @@ class TolinoWarmthController : LightInterface {
         private const val BRIGHTNESS_MAX = 255
         private const val WARMTH_MAX = 10
         private const val MIN = 0
-        private const val COLOR_FILE = "/sys/class/backlight/tlc5947_bl/color"
+        private const val COLOR_FILE_EPOS2 = "/sys/class/backlight/tlc5947_bl/color"
+        private const val COLOR_FILE_VISION4HD = "/sys/class/backlight/lm3630a_led/color"
+        private var COLOR_FILE = "" // gets initialized before first write access
```
pazos, line 25: `lateinit var COLOR_FILE : String` is more idomatic 👍

zwim:
@pazos: If I try to run the last commit https://github.com/koreader/android-luajit-launcher/pull/244/commits/837c7a900de49d403abdf3a6e366fdaa43ff36fb then the app crashes. 
`E/AndroidRuntime( 3058): kotlin.UninitializedPropertyAccessException: lateinit property COLOR_FILE has not been initialized`


What I am doing wrong?

zwim:
I have found the problem. Reading the logs helps a lot ;-)

zwim:
@pazos: The problem was that COLOR_FILE is also used at getWarmth. 
I corrected that, but somehow `./kodev release android` does not compile my corrected file. I crosschecked this, with a deliberate syntax error in TolinoWarmthController.kt. But `./kodev release android` does NOT fail!
What the hell could that be?
No I am doing a `./kodev clean android` and than a `./kodev release android`, let’s see what comes out.

Frenzie:
@zwim No need to do a full clean (hopefully you use ccache 'cause that'll take half an hour ;-) ), use a `make clean` in the luajit-launcher directory instead.

zwim:
@Frenzie : I did the make clean in luajit-launcher before, but it did not help. After the full clean it seems to work now.

This was tested on the Vision 4 and the Epos 2. It works now.

Frenzie:
@zwim There's always more to clean in the luajit-launcher subdirectory I'm sure; `make mrproper` should clean the gradle stuff. (Although your changes should be detected; not sure what to make of that.)

More generically, you could use something like `git stash`, `git reset --hard`, `git stash pop` — that'd remove untracked files, beware. But kodev clean will get rid of *everything*.

pazos:
> @pazos: If I try to run the last commit [837c7a9](https://github.com/koreader/android-luajit-launcher/commit/837c7a900de49d403abdf3a6e366fdaa43ff36fb) then the app crashes.
> `E/AndroidRuntime( 3058): kotlin.UninitializedPropertyAccessException: lateinit property COLOR_FILE has not been initialized`
> 
> What I am doing wrong?

Kotlin provides null safety. You declare COLOR_FILE as a string and you need to initialize it later.
Please don't do any fancy stuff. The original code worked fine. If you want to add new models then check for the specific files for these models first. If not set COLOR_FILE as the one used on your Tolino (like it was).

```kotlin
if (File(COLOR_FILE_VISION4HD).exists()) {
    COLOR_FILE = COLOR_FILE_VISION4HD
}
else // original COLOR_FILE for your epos 2
```

zwim:
> ```kotlin
> if (File(COLOR_FILE_VISION4HD).exists()) {
>     COLOR_FILE = COLOR_FILE_VISION4HD
> }
> else // original COLOR_FILE for your epos 2
> ```

Yes, that is better.

zwim:
This works fine and it is less complicated.

app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
```diff
@@ -22,10 +22,8 @@ class TolinoWarmthController : LightInterface {
         private const val MIN = 0
         private const val COLOR_FILE_EPOS2 = "/sys/class/backlight/tlc5947_bl/color"
         private const val COLOR_FILE_VISION4HD = "/sys/class/backlight/lm3630a_led/color"
-	private val COLOR_FILE = if (File(COLOR_FILE_VISION4HD).exists())
-	    COLOR_FILE_VISION4HD
-	else 
-	    COLOR_FILE_EPOS2
+        
+        private val COLOR_FILE = if (File(COLOR_FILE_VISION4HD).exists()) COLOR_FILE_VISION4HD else COLOR_FILE_EPOS2
```
zwim, line 26: Isn't the line to long?
https://github.com/koreader/koreader-base/wiki/Coding-style

pazos: I find it easier to understand in one line, but have no problem reverting back to more lines if you find it confusing :). The main intent for the change was identation.

Anyways, the codying style you mention doesn't match the rules we have enforced in this repo (see https://github.com/koreader/android-luajit-launcher/blob/master/detekt.yml#L209-L211). Not something that I would reject to change, but enforcing a maxLineLength of 80 characters will probably trigger a few issues on the codebase.

zwim: I would find it easier, if the if would be separated over more lines. So you can see on one view what is happening, instead of reading the whole thing. But you are more familiar with the code, so I obey your decisions and your taste. ;-)

For the rules: I messed up koreader/koreader and koreader/android-luajit-launcher, sorry.

Correcting the indent is clear.

pazos: Nah, I like to see you interested by android and reviewing things here. Feel free to suggest any improvement that comes to your mind :+1:

zwim: OK, for me the most readable variant would be your first proposal. (Although it might be a bit dangerous, cause of the missing curly braces, but therefore we have the indentaton):
```
        private val COLOR_FILE = if (File(COLOR_FILE_VISION4HD).exists()) 
            COLOR_FILE_VISION4HD
        else 
            COLOR_FILE_EPOS2
```
But either, your split or my proposal are fine.
Thank you.

pazos: Please do it.

Now I'm curious about https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/utils/FileUtils.kt#L31. Do you find that easy to read? :p

zwim: Now I will go and read something on my Epos2 with that awesome software named KOReader. Bye.

Frenzie: @pazos I *can* read it but easy's definitely not the right word. :-)

zwim: @pazos: Your example is short. If you have to code something very compact, it is readable, but not easily.
I would prefer a longer version. The good thing is the documentation above the function.

What is in a month, six months a year. When coding your thoughts, you are fully in the problem. But others are not, so making it clear would be a benefit. This are my six cent.

pazos: I agree that well formatted code is easier to read. But kotlin makes expresiveness next level. Take something like https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/BaseActivity.kt#L291-L300. It makes sense to split the function in several lines because its length, but if the same function would return 1 or 0 I wouldn't think that:

```kotlin
return first?.let { second?.let { 0 } ?: 1 } ?: 1
```
is harder to read than

```kotlin
return first?.let {
    second?.let {
        0
    } ?: 1
} ?: 1
```

the same happens with *some* return statements

```kotlin
return if (something) something else "nothing"
```

VS

```kotlin
return if (something)
    something
else
    "nothing"
```

Frenzie: > return if (something) something else "nothing"

Much easier to read imo. :-)

PR #247, opened 2020-08-22
Methods to check if frontlight is changeable

zwim:
Check if frontlight is changeable or locked by system settings, on Tolino.
On Generic Android assume frontlight is always changeable.

It was tested on Tolino Epos2, Vision4, a Phone and on the kodev emulator.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/247)

### DIFF
--- a/app/src/org/koreader/launcher/MainActivity.kt
+++ b/app/src/org/koreader/launcher/MainActivity.kt
@@ -222,8 +222,8 @@ class MainActivity : BaseActivity() {
      *             override methods used by lua/JNI                *
      *--------------------------------------------------------------*/
 
-    override fun canChangeFrontlight(): Int {
-        return lights.canChangeFrontlight(this@MainActivity)
+    override fun toggleFrontlightSwitchOn(): Int {
+        return lights.toggleFrontlightSwitchOn(this@MainActivity)
     }
 
     override fun getScreenBrightness(): Int {
--- a/app/src/org/koreader/launcher/MainActivity.kt
+++ b/app/src/org/koreader/launcher/MainActivity.kt
@@ -222,8 +222,8 @@ class MainActivity : BaseActivity() {
      *             override methods used by lua/JNI                *
      *--------------------------------------------------------------*/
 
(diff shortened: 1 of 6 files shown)


zwim:
> I think something like enableFrontlight (more KOReader) or setFrontlightOn (more Android) might be slightly more elegant.

We can use any name. But the method actually just toggles a Switch. If you set the intensity manually to zero and toggle the switch, the frontlight stays off. So calling the method setFrontlighOn() might be misleading.

The Epos2 has a SoftwareButton (in the Tolino app) to toggle frontlight. 
If that button is on, the intensity can be changed as usual.
If that button is off, you can set and read the intensity in Tolino/Android, but actually the frontlight is off.
Now if you toggle the button on, the last intensity of the Tolino/Android system is used.
e.g: 
1. button is **off** -> frontlight is **off**,
2. change frontlight to 50 -> frontlight is **off**, 
3. change frontlight to 40 -> frontlight is **off**,
4. toggle Button on -> frontlight is at 40,
5. toggle Button off -> frontlight is **off**,
6. toggle Buttion on -> frontlight is 40,
7. change frontlight to 0,
6. you can toggle the button on or off -> frontlight is **off**

Frenzie:
We might have to agree to disagree. To me, toggle implies setting it the opposite of whatever it currently is. The phrase "toggle on" is clear but possibly incorrect. So far no big deal, but where it starts to matter is if you're skimming sometime in the future.

You might think, `toggleFrontLight`, I know what that does, and you might overlook that sneaky `SwitchOn` at the end of it. That's not what `toggle` is expected to do.

> So calling the method setFrontlighOn() might be misleading.

I don't see how it's any more or less misleading though. Perhaps I'm focusing on `toggle` while you're focusing on `switch`? Which does seem somewhat redundant to me, but while I'm not a fan of its verbosity it affects clarity less.

zwim:
@Frenzie: You brought it to the point: I was focused on the switch. Would `enableFrontlightSwitch` be a feasible naming?

PR #249, opened 2020-08-28
add method to check frontlight switch state, correct return value of enableFrontlightState

zwim:
This method will be necessary to give a correct light status in the status line on startup.

The return value of enableFrontlightState was wrong (did not correspond to the comment above the method).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/249)

### DIFF
--- a/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
+++ b/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
@@ -41,8 +41,7 @@ class TolinoWarmthController : LightInterface {
         return false
     }
 
-    // try to toggle on frontlight switch on Tolinos, returns the former switch state
-    override fun enableFrontlightSwitch(activity: Activity): Int {
+    override fun getFrontlightSwitch(activity: Activity): Int {
         // ATTENTION: getBrightness, setBrightness use the Android range 0..255
         // in the brightness files the used range is 0..100
         val startBrightness = getBrightness(activity)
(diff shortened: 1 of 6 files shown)


app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
```diff
@@ -41,8 +41,7 @@ class TolinoWarmthController : LightInterface {
         return false
     }
 
-    // try to toggle on frontlight switch on Tolinos, returns the former switch state
-    override fun enableFrontlightSwitch(activity: Activity): Int {
+    override fun getFrontlightSwitch(activity: Activity): Int {
```
Frenzie, line 44: getFrontlightSwitchStatus?

Frenzie (approved):
Seems fine, not tested or anything of course ;-)

app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt
zwim: Why not. I'll change that tomorrow, if nothing else pops up.

Galunid:
@zwim While you're at it
https://github.com/koreader/android-luajit-launcher/blob/b8fec042c433276f0dc7ff352c36d2328f5daa83/app/src/org/koreader/launcher/device/lights/TolinoWarmthController.kt#L124-L125
I assume`BRIGHTNESS_MAX` should be `WARMTH_MAX`

zwim:
> I assume`BRIGHTNESS_MAX` should be `WARMTH_MAX`

Your are right. :+1:

pazos:
@zwim: closing this PR as it was intended to be merged with https://github.com/koreader/koreader/pull/6594. Please correct me if I'm wrong :)

zwim:
@pazos: You are right. This PR was mainly for  koreader/koreader#6594. 

But it would be good to apply "corrected range check in SetWarmth" (d894296) as this would be the correct range check.

PR #250, opened 2020-08-31
Implement EPD Controller for newer Onyx Boox devices (Qualcomm)

Galunid:
Currently onyx devices rely on system controller, which doesn't work as well as it could with KOReader.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/250)

### DIFF
--- a/app/src/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/org/koreader/launcher/device/DeviceInfo.kt
@@ -14,6 +14,7 @@ object DeviceInfo {
     val PRODUCT: String
     val EINK_FREESCALE: Boolean
     val EINK_ROCKCHIP: Boolean
+    val EINK_QCOM: Boolean
     val EINK_SUPPORT: Boolean
     val EINK_FULL_SUPPORT: Boolean
     val BUG_WAKELOCKS: Boolean
@@ -65,6 +66,7 @@ object DeviceInfo {
         CREMA,
         CREMA_0650L,
         ONYX_C67,
+        ONYX_NOVA2,
         ENERGY,
         INKBOOK,
         TOLINO,
@@ -149,6 +151,7 @@ object DeviceInfo {
                 && PRODUCT.contentEquals("nova2")
                 && DEVICE.contentEquals("nova2"))
         lightsMap[LightsDevice.ONYX_NOVA2] = ONYX_NOVA2
+        deviceMap[EinkDevice.ONYX_NOVA2] = ONYX_NOVA2
 
         // Onyx C67
         ONYX_C67 = (MANUFACTURER.contentEquals("onyx")
@@ -243,11 +246,13 @@ object DeviceInfo {
             INKBOOK ||
             ONYX_C67
 
+        EINK_QCOM = ONYX_NOVA2
+
         // basic eink support
-        EINK_SUPPORT = EINK_FREESCALE || EINK_ROCKCHIP
+        EINK_SUPPORT = EINK_FREESCALE || EINK_ROCKCHIP || EINK_QCOM
 
         // full eink support
-        EINK_FULL_SUPPORT = CREMA || TOLINO
+        EINK_FULL_SUPPORT = CREMA || TOLINO || ONYX_NOVA2
 
         // need wakelocks
         BUG_WAKELOCKS = BUG == BugDevice.SONY_RP1
@@ -256,7 +261,7 @@ object DeviceInfo {
         BUG_SCREEN_ROTATION = BUG == BugDevice.EMULATOR
 
         // needs a surfaceView to do epd updates
-        NEEDS_VIEW = EINK_FREESCALE
+        NEEDS_VIEW = EINK_FREESCALE || EINK_QCOM
     }
 
     private fun getBuildField(fieldName: String): String {
--- a/app/res/layout/main.xml
+++ b/app/res/layout/main.xml
@@ -77,6 +77,24 @@
 
         <TextView
             android:layout_width="fill_parent"
+            android:layout_height="12dp" />
+
+        <TextView
+            android:id="@+id/qualcommText"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:gravity="center_horizontal"
+            android:textSize="18sp" />
+
+        <Button
+            android:id="@+id/qualcommButton"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:text="qualcomm/onyx update"
+            android:textSize="28sp" />
+
+        <TextView
+            android:layout_width="fill_parent"
             android:layout_height="18dp" />
 
         <TextView
--- a/app/src/org/koreader/launcher/EPDTestActivity.kt
+++ b/app/src/org/koreader/launcher/EPDTestActivity.kt
@@ -7,6 +7,7 @@ import android.view.View
 import android.widget.Button
 import android.widget.TextView
 import org.koreader.launcher.device.epd.freescale.NTXEPDController
+import org.koreader.launcher.device.epd.qualcomm.QualcommEPDController
 import org.koreader.launcher.device.epd.rockchip.RK30xxEPDController
 import org.koreader.launcher.device.epd.rockchip.RK33xxEPDController
 import java.util.*
@@ -26,9 +27,11 @@ class EPDTestActivity : Activity() {
         val rk30xxDescription: TextView = findViewById(R.id.rk30xxText)
         val rk33xxDescription: TextView = findViewById(R.id.rk33xxText)
         val ntxNewDescription: TextView = findViewById(R.id.ntxNewText)
+        val qualcommDescription: TextView = findViewById(R.id.qualcommText)
         val rk30xxButton: Button = findViewById(R.id.rk30xxButton)
         val rk33xxButton: Button = findViewById(R.id.rk33xxButton)
         val ntxNewButton: Button = findViewById(R.id.ntxNewButton)
+        val qualcommButton: Button = findViewById(R.id.qualcommButton)
         val shareButton: Button = findViewById(R.id.shareButton)
 
         /* current device info */
@@ -66,6 +69,10 @@ class EPDTestActivity : Activity() {
         ntxNewDescription.text = "This button should work on modern Tolinos/Nooks and other ntx boards"
         ntxNewButton.setOnClickListener { runEinkTest(NTXTEST) }
 
+        /* qualcomm - At least Onyx Boox Nova 2 */
+        qualcommDescription.text = "This button should work on some Onyx Boox devices. At least qualcomm ones"
+        qualcommButton.setOnClickListener { runEinkTest(QUALCOMMTEST) }
+
         /* share button */
         shareButton.setOnClickListener { shareText(info.text.toString()) }
     }
@@ -82,15 +89,20 @@ class EPDTestActivity : Activity() {
             } else if (test == RK33xxTEST) {
                 info.append("rk33xx: ")
                 if (RK33xxEPDController.requestEpdMode("EPD_FULL")) success = true
-            } else if (test == NTXTEST) {
+            } else if (test == NTXTEST || test == QUALCOMMTEST) {
                 // get screen width and height
                 val display = windowManager.defaultDisplay
                 val size = Point()
                 display.getSize(size)
                 val width = size.x
                 val height = size.y
-                info.append("tolino: ")
-                if (NTXEPDController.requestEpdMode(v, 34, 50, 0, 0, width, height)) success = true
+                if (test == NTXTEST) {
+                    info.append("tolino: ")
+                    if (NTXEPDController.requestEpdMode(v, 34, 50, 0, 0, width, height)) success = true
+                } else if (test == QUALCOMMTEST) {
+                    info.append("qualcomm: ")
+                    if (QualcommEPDController.requestEpdMode(v, 98, 50, 0, 0, width, height)) success = true
+                }
             }
         } catch (e: Exception) {
             e.printStackTrace()
@@ -113,6 +125,7 @@ class EPDTestActivity : Activity() {
         private const val RK30xxTEST = 1
         private const val RK33xxTEST = 2
         private const val NTXTEST = 3
+        private const val QUALCOMMTEST = 4
         private val MANUFACTURER = android.os.Build.MANUFACTURER
         private val BRAND = android.os.Build.BRAND
         private val MODEL = android.os.Build.MODEL
--- a/app/src/org/koreader/launcher/device/EPDFactory.kt
+++ b/app/src/org/koreader/launcher/device/EPDFactory.kt
@@ -7,6 +7,7 @@ import org.koreader.launcher.Logger
 import org.koreader.launcher.device.epd.freescale.NTXNewEPDController
 import org.koreader.launcher.device.epd.rockchip.RK3026EPDController
 import org.koreader.launcher.device.epd.rockchip.RK3368EPDController
+import org.koreader.launcher.device.epd.qualcomm.QualcommOnyxEPDController
 import org.koreader.launcher.interfaces.EPDInterface
 import java.util.*
 
@@ -38,6 +39,10 @@ object EPDFactory {
                 DeviceInfo.EinkDevice.NOOK_V520 -> {
                     logController("Freescale NTX")
                     NTXNewEPDController()
+                }
+                DeviceInfo.EinkDevice.ONYX_NOVA2 -> {
+                    logController("QualcommEPDController")
+                    QualcommOnyxEPDController()
                 }
                 else -> {
                     FakeEPDController()
--- a/app/src/org/koreader/launcher/device/epd/qualcomm/QualcommEPDController.kt
+++ b/app/src/org/koreader/launcher/device/epd/qualcomm/QualcommEPDController.kt
@@ -1,0 +1,67 @@
+package org.koreader.launcher.device.epd.qualcomm
+
+import java.lang.Thread
+import java.util.Locale
+
+import android.view.View
+import android.util.Log
+
+import org.koreader.launcher.interfaces.EPDInterface
+
+// More information including epd mode values
+// https://github.com/koreader/android-luajit-launcher/pull/250#issuecomment-711443457
+abstract class QualcommEPDController : EPDInterface {
+    companion object  {
+        private const val TAG = "epd"
+
+        fun preventSystemRefresh() : Boolean{
+            // Sets UpdateMode and UpdateScheme to None
+            // this function is called EpdController.setSystemUpdateModeAndScheme in onyxsdk
+            return try{
+                Class.forName("android.view.View").getMethod("setWaveformAndScheme",
+                    Integer.TYPE,
+                    Integer.TYPE,
+                    Integer.TYPE).invoke(null, 5, 1, 0)
+                true
+            } catch (e: Exception) {
+                Log.e(TAG, e.toString())
+                false
+            }
+        }
+
+        fun requestEpdMode(targetView: android.view.View,
+                                mode: Int, delay: Long,
+                                x: Int, y: Int, width: Int, height: Int) : Boolean
+        {
+            return try{
+                // We need to always call this, not sure why, if it's not called before
+                // system will refresh after us, it'll refresh anyway if user set
+                // Normal mode, or Regal mode works flawlessly otherwise
+                preventSystemRefresh()
+                // EpdController.refreshScreenRegion in onyxsdk
+                var refreshScreen = Class.forName("android.view.View").getMethod("refreshScreen",
+                                        Integer.TYPE,
+                                        Integer.TYPE,
+                                        Integer.TYPE,
+                                        Integer.TYPE,
+                                        Integer.TYPE)
+                object: Thread(){
+                    override fun run(){
+                        Thread.sleep(delay)
+                        try {
+                            refreshScreen.invoke(targetView, x, y, width, height, mode)
+                            Log.i(TAG, String.format(Locale.US,
+                                "requested eink refresh, type: %d x:%d y:%d w:%d h:%d",
+                                mode, x, y, width, height))
+                        } catch(e: Exception) {
+                            Log.e(TAG, e.toString())
+                        }
+                    }
+                }.start()
+                true
+            } catch(e: Exception) {
+                false
+            }
+        }
+    }
+}
--- a/app/src/org/koreader/launcher/device/epd/qualcomm/QualcommOnyxEPDController.kt
+++ b/app/src/org/koreader/launcher/device/epd/qualcomm/QualcommOnyxEPDController.kt
@@ -1,0 +1,14 @@
+/* Tested on Onyx Boox Nova 2 */
+
+package org.koreader.launcher.device.epd.qualcomm
+
+import org.koreader.launcher.interfaces.EPDInterface
+
+class QualcommOnyxEPDController : QualcommEPDController(), EPDInterface {
+    override fun setEpdMode(targetView: android.view.View,
+                            mode: Int, delay: Long,
+                            x: Int, y: Int, width: Int, height: Int, epdMode: String?)
+    {
+        requestEpdMode(targetView, mode, delay, x, y, width, height)
+    }
+}
--- a/app/src/org/koreader/launcher/device/epd/qualcomm/QualcommOnyxEPDController.kt
+++ b/app/src/org/koreader/launcher/device/epd/qualcomm/QualcommOnyxEPDController.kt
@@ -5,6 +5,8 @@ package org.koreader.launcher.device.epd.qualcomm
 import org.koreader.launcher.interfaces.EPDInterface
 
 class QualcommOnyxEPDController : QualcommEPDController(), EPDInterface {
+    override fun resume() {}
+    override fun pause() {}
     override fun setEpdMode(targetView: android.view.View,
                             mode: Int, delay: Long,
                             x: Int, y: Int, width: Int, height: Int, epdMode: String?)


Galunid:
Any idea why it doesn't refresh screen when `OnyxEPDController.setEpdMode` is not called on lua side? Do I need to register it somewhere else, other than in `EDPFactory` and `DeviceInfo`? I get log that it uses `OnyxEpdController`, but it doesn't do anything. It works in `EPDTestActivity`, like it's supposed to.

app/build.gradle
```diff
@@ -50,6 +50,7 @@ android {
         }
         arm {
             ndk { abiFilters "armeabi-v7a" }
+            minSdkVersion safSdk
```
pazos, line 53: I'm afraid we cannot merge this directly. That would mean that the minimum android version is now Lollipop and that's a big step since most eink readers are still 4.x.

app/build.gradle
```diff
@@ -112,6 +113,9 @@ dependencies {
     implementation "androidx.legacy:legacy-support-v4:$androidx_supportv4_version"
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_plugin_version"
 
+    // OnyxSDK
+    implementation('com.onyx.android.sdk:onyxsdk-base:1.5.4')
```
pazos, line 117: You can try to implement what onyx is doing under the hood, since it will be based on reflection, like us are doing.

app/manifest/base.xml
```diff
@@ -16,7 +17,7 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.WRITE_SETTINGS" />
 
-    <application
+    <application tools:replace="android:allowBackup"
```
pazos, line 20: Is that needed?, why?

pazos:
> Any idea why it doesn't refresh screen when `OnyxEPDController.setEpdMode` is not called on lua side? Do I need to register it somewhere else, other than in `EDPFactory` and `DeviceInfo`? I get log that it uses `OnyxEpdController`, but it doesn't do anything. It works in `EPDTestActivity`, like it's supposed to.

EPDTestActivity is hooked on android view hierarchy. The native window is not. That's why we pass a view to `setEpd` things. 

I would suggest to look at `/proc/cpuinfo` and see if it has a similar SoC of one of these supported. I think it should work fine with Nook EPD Controller, if you want to perform just full screen refreshes with it.

pazos:
> Do I need to register it somewhere else, other than in `EDPFactory` and `DeviceInfo`? 

Also yeah. Give it a look at https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/device/DeviceInfo.kt#L225-L250

app/manifest/base.xml
Galunid: It's needed since Onyxsdk's manifest has `android:allowBackup = true`. This introduces conflict in our manifest and gradle screams at me for doing so and refuese to build ;-).

app/build.gradle
Galunid: Yup, it's a hack to get things testing quickly, it needs minSdkVersion >= 16, I'll probably add a new target for onyx. Same thing with modification in `EPDTestActivity`. I'll revert those when I'm preparing to merge.

app/build.gradle
pazos: For example, there's a method called [`refreshScreenRegion`](https://github.com/buggins/coolreader/blob/master/android/src/com/onyx/android/sdk/api/device/epd/EpdController.java#L106-L108) which just issues a screen update for the current device. Assuming the device is IMX7 is just does https://github.com/buggins/coolreader/blob/master/android/src/com/onyx/android/sdk/device/IMX7Device.java#L151

Transforming that to a driver that skips the need of onyx sdk is not hard, but needs some time for trial and error. See https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/device/epd/freescale/NTXEPDController.kt#L41-L58.

Galunid:
I want to have partial refreshes, I added full refresh, to make certain it's handled by my controller, instead of Onyx's

Galunid:
> I would suggest to look at `/proc/cpuinfo` and see if it has a similar SoC of one of these supported. I think it should work fine with Nook EPD Controller, if you want to perform just full screen refreshes with it.

SoC is Qualcomm MSM8953 (Snapdragon 625)
Nook's EPD Controller doesn't work here. It doesn't get any exception, but it doesn't refresh screen either. Onyx has different modes than `freescale`/`rockchip`.

Galunid:
Notes to self and anyone interested
Example refresh log by normal usage of the device:
`I SDM     : update_to_display[0/1] -- marker[2903] waveform_mode = 4, update_mode = 0, Rect[0 0 1872 1404], flags = 8000`
From what I've found there are 4 `waveform_mode` values that user can choose:
`Normal mode` - 255
`Regal mode` - 5
`A2 mode` - 1
`X mode` - 4
additionally there is a full update with mode 2
`update_mode` is either 1 or 0, I haven't encountered any other, but that doesn't mean there isn't. 1 is for full refresh, 0 is for other types I believe.
There's enum in OnyxSdk, but it's used internally by device's implementations, to convert to `UpdateMode`
```java
public enum EPDMode {
    FULL,
    AUTO,
    TEXT,
    AUTO_PART,
    AUTO_BLACK_WHITE,
    AUTO_A2,
    EPD_REGLA
}
```

```java
public enum UpdateMode {
    None,
    DU,
    GU,
    GU_FAST,
    GC,
    ANIMATION,
    ANIMATION_QUALITY,
    ANIMATION_MONO,
    ANIMATION_X,
    GC4,
    REGAL,
    REGAL_D,
    DU_QUALITY,
    HAND_WRITING_REPAINT_MODE
}
```
Other interesting things
```java
int staticIntFieldSafely = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_ONYX_AUTO_MASK");
        int staticIntFieldSafely2 = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_ONYX_GC_MASK");
        int staticIntFieldSafely3 = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_AUTO_MODE_REGIONAL");
        int staticIntFieldSafely4 = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_WAIT_MODE_NOWAIT");
        int staticIntFieldSafely5 = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_WAIT_MODE_WAIT");
        int staticIntFieldSafely6 = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_WAVEFORM_MODE_DU");
        int staticIntFieldSafely7 = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_WAVEFORM_MODE_ANIM");
        int staticIntFieldSafely8 = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_WAVEFORM_MODE_GC4");
        int staticIntFieldSafely9 = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_WAVEFORM_MODE_GC16");
        int staticIntFieldSafely10 = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_WAVEFORM_MODE_REAGL");
        int staticIntFieldSafely11 = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_REAGL_MODE_REAGLD");
        int staticIntFieldSafely12 = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_UPDATE_MODE_PARTIAL");
        int staticIntFieldSafely13 = ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_UPDATE_MODE_FULL");
        int staticIntFieldSafely14 = ReflectUtil.getStaticIntFieldSafely(classForName, "UI_DEFAULT_MODE");
        ReflectUtil.getStaticIntFieldSafely(classForName, "EINK_DITHER_MODE_DITHER");
        d = staticIntFieldSafely;
        e = staticIntFieldSafely2;
        f = staticIntFieldSafely3 | staticIntFieldSafely4 | staticIntFieldSafely6 | staticIntFieldSafely12;
        g = ReflectUtil.getStaticIntFieldSafely(classForName, "UI_DU_QUALITY_MODE");
        h = staticIntFieldSafely3 | staticIntFieldSafely4 | staticIntFieldSafely9 | staticIntFieldSafely12;
        i = staticIntFieldSafely3 | staticIntFieldSafely5 | staticIntFieldSafely9 | staticIntFieldSafely13;
        j = staticIntFieldSafely3 | staticIntFieldSafely4 | staticIntFieldSafely7 | staticIntFieldSafely12;
        k = ReflectUtil.getStaticIntFieldSafely(classForName, "UI_A2_QUALITY_MODE");
        l = ReflectUtil.getStaticIntFieldSafely(classForName, "UI_MONO_A2_MODE");
        m = ReflectUtil.getStaticIntFieldSafely(classForName, "UI_X_A2_MODE");
        n = staticIntFieldSafely3 | staticIntFieldSafely4 | staticIntFieldSafely8 | staticIntFieldSafely12;
        o = staticIntFieldSafely3 | staticIntFieldSafely4 | staticIntFieldSafely10 | staticIntFieldSafely12;
        p = staticIntFieldSafely3 | staticIntFieldSafely4 | staticIntFieldSafely11 | staticIntFieldSafely10 | staticIntFieldSafely12;
        q = staticIntFieldSafely14;
        r = ReflectUtil.getStaticIntFieldSafely(classForName, "HAND_WRITING_REPAINT_MODE");
        Class<?> classForName2 = ReflectUtil.classForName("android.onyx.optimization.Constant");
        // used by isInSystemRefreshModeDefault
        UPDATE_MODE_DEFAULT = ReflectUtil.getStaticIntFieldSafely(classForName2, "UPDATE_MODE_DEFAULT");
        // not used anywhere in code
        UPDATE_MODE_DU = ReflectUtil.getStaticIntFieldSafely(classForName2, "UPDATE_MODE_DU");
        // not used anywhere in code
        UPDATE_MODE_A2 = ReflectUtil.getStaticIntFieldSafely(classForName2, "UPDATE_MODE_A2");
        // not used anywhere in code
        UPDATE_MODE_REGAL = ReflectUtil.getStaticIntFieldSafely(classForName2, "UPDATE_MODE_REGAL");
        // used by isInSystemRefreshModeX
        UPDATE_MODE_X = ReflectUtil.getStaticIntFieldSafely(classForName2, "UPDATE_MODE_X");
```

pazos:
> SoC is Qualcomm MSM8953 (Snapdragon 625)

Good, new platform :)

> Nook's EPD Controller doesn't work here. It doesn't get any exception, but it doesn't refresh screen either. Onyx has different modes than freescale/rockchip.

Also good :). It is not expected to work as is. But all supported platforms follow the same scheme:

- there's a method in the view class that's in charge of requesting a screen update
- we invoke that method using reflection, that's useful as we just need to know the name of the class and the parameters it takes.
- if there's no sdk, which is usual, the stock app is re'd to know the name of the method and its parameters

In the case of Tolinos (and other full_eink devices) the system driver doesn't do any refresh on its own and we do them all.
In the rest of cases the system driver update the window on its own and we just flash on user request.

In the case of onyx, I think there's a way to prevent system updates (some sort of manual mode) and then act like in Tolinos. If you don't prevent system updates you'll get 2x screen refreshes.

Galunid:
It "works". The bad news is that system applications don't work, after going home, there's no screen refresh etc. And KOReader doesn't always update properly, but it's a start.

Galunid:
Sidenote, full screen refresh in KOReader on "normal" mode is a bit slower than "onyx A2" refresh, where A2 degrades quality quite a bit. The improvement is drastic.

Galunid:
When calling `refreshScreen`
```java
public enum UpdateMode {
    None, // 5
    DU, // 1
    GU, // 2
    GU_FAST, // 1
    GC, // 98
    ANIMATION, // 4
    ANIMATION_QUALITY, // 2308
    ANIMATION_MONO, // 33554436
    ANIMATION_X, // 16777220
    GC4, // 3
    REGAL, // 6
    REGAL_D, // 4102
    DU_QUALITY, // 2305
    HAND_WRITING_REPAINT_MODE // 524290
}
```

app/src/org/koreader/launcher/device/epd/onyx/OnyxEPDController.kt
```diff
@@ -0,0 +1,72 @@
...
+            2 -> UpdateMode.GC
+            3 -> UpdateMode.GC4
+            4 -> UpdateMode.GC
+            5 -> UpdateMode.ANIMATION_QUALITY
+            7 -> UpdateMode.ANIMATION_QUALITY
+            else -> UpdateMode.GC
+        }
+        object : Thread(){
+            override fun run(){
+                EpdController.enableScreenUpdate(targetView, true)
```
pazos, line 58: I think we can add `onResume` and `onPause` methods to EPDInterface, enable manual mode on resume and disable it on pause.

This way you can avoid the thread, the delay and just fire `EpdController.refreshScreenRegion(targetView, x, y, x + width, y + height, updateMode)`.

BTW, the final waveform depends on https://github.com/koreader/koreader-base/blob/master/ffi/framebuffer_android.lua, which depends on proper DeviceInfo ids, like HAS_EINK_SUPPORT and HAS_EINK_FULL_SUPPORT

Galunid: Well, this doesn't work anyway, the refresh is queued and after enabling refresh again, the queue gets emptied, regardless of delay. Ironically, `enableScreenUpdate` blocks our updates too... I asked in https://github.com/onyx-intl/OnyxAndroidDemo/issues/40 if there's a way to prevent updates, but I'm not too optimistic about getting a response ;)

Galunid: For now, I'd like to get it working without onyx sdk, since there are multiple reasons for not using it, we need like 2 functions.

Galunid:
This "works". The refresh mode is always set to `94` (full refresh). Mapping other modes shouldn't be a problem.
Unfortunately it's wonky (**refresh happens before page is changed**).
@pazos Any idea why that may happen?

app/src/org/koreader/launcher/device/epd/onyx/OnyxEPDController.kt
```diff
@@ -0,0 +1,41 @@
+package org.koreader.launcher.device.epd.onyx
+
+import java.lang.Thread
+
+import android.util.Log
+import android.view.View
+
+import com.onyx.android.sdk.api.device.epd.EpdController
```
pazos: these aren't needed now, right?

pazos:
> This "works". The refresh mode is always set to `98` (full refresh). Mapping other modes shouldn't be a problem.
> Unfortunately it's wonky (**refresh happens before page is changed**).
> @pazos Any idea why that may happen?

Emm, I see you are somehow using the same settings that the rest of freescale devices. Also you implemented a delay with a thread, that's sounds sensible. 

But, **again**, you need to tweak https://github.com/koreader/koreader-base/blob/master/ffi/framebuffer_android.lua to use the waveform modes and the delays you want.

For example, see how Freescale devices are now handling `refreshPartialImp(x, y, w, h)` https://github.com/koreader/koreader-base/blob/master/ffi/framebuffer_android.lua#L152. You need to add a delay other than 0 and write the specific waveform mode.

app/src/org/koreader/launcher/device/epd/onyx/OnyxEPDController.kt
Galunid: Yes, it was already removed in https://github.com/koreader/android-luajit-launcher/pull/250/commits/01ad467ba1daf6731ee700a3f8d5ced3680765a1

app/src/org/koreader/launcher/device/DeviceInfo.kt
```diff
@@ -259,7 +261,7 @@ object DeviceInfo {
         BUG_SCREEN_ROTATION = BUG == BugDevice.EMULATOR
 
         // needs a view
-        NEEDS_VIEW = ! EINK_SUPPORT || EINK_FREESCALE
+        NEEDS_VIEW = ! EINK_SUPPORT || EINK_FREESCALE || EINK_QCOM
```
pazos, line 264: Not sure if it was legible enough.

NEEDS_VIEW is used in MainActivity to create a surfaceView and take control of the native window from the java side. That's used to perform epd updates, because passing the ViewRootImpl() as is doesn't work.

The logic right now is: enable the surfaceView **unless** it is a eink device and not a freescale one. My advice would be check first if the device works without the surfaceView as the surfaceView is costly and might introduce some delay.

Galunid: I removed it and there's indeed quite a bit difference. It works way better now. Earlier I had issues with navigating menu I didn't know what was causing them. They're gone now. I guess it wasn't legible enough, I took it as "you want to get `view` as a parameter, then set this to true".

pazos: Probably doesn't make any sense now to have it on DeviceInfo. The surfaceView hack only makes sense on Nooks and Tolinos (EINK_FREESCALE). It is also needed on androidTV and ChromeOS because CRe documents went black if there's no input for a few seconds.

Something like the following could make sense

```kotlin

/* Whether this device needs a SurfaceView implementation or not. Most devices work well (and faster) without one.
Only enable it if your device needs it for EPD updates (like Tolinos) or somehow the system driver work badly without one
(certain android TV boxes, which dim the screen if they don't detect a "valid" view hierarchy) */

NEEDS_VIEW = EINK_FREESCALE
```

pazos (changes requested):
Looks way better now, congrats!!

Just 2cts to look even better:

- move the core logic to an abstract class called `QualcommEPDController` and wrap reflection methods in a try catch block. 
- rename the class that implements the EPDInterface to something less generic, like `QualcommOnyxEPDController` and extend from `QualcommEPDController`

That way we can add calls to the abstract class directly in `EPDTestActivity` without crashes, and you can add all the comments you want in the abstract class.

Please have a look at the difference between `NTXEPDController` and `NTXNewEPDController` to understand what I'm talking about.

Galunid:
It's working pretty well, at least better than the default in my opinion, some things can definitely be improved with fine-tuning stuff in base. What's left to do is implement `onResume`/`onPause`, since we need to disable updates only once, not on every refresh. And setting default mode to A2, and disabling "system" full refreshes, since those are now handled by KOReader. Other than that it needs a bit of refactoring and safeguards against reflection crashes

Galunid:
Just to confirm, adding `epd.onPause()` in `MainActivity.kt` shouldn't crash anything, if I created default method in interface with empty implementation?
```diff
@@ -126,6 +126,7 @@ class MainActivity : BaseActivity() {
     override fun onResume() {
         Logger.d(TAG_MAIN, "onResume()")
         super.onResume()
+        epd.onResume()
         applyCustomTimeout(true)
     }
 
@@ -133,6 +134,7 @@ class MainActivity : BaseActivity() {
     override fun onPause() {
         Logger.d(TAG_MAIN, "onPause()")
         super.onPause()
+        epd.onPause()

interface EPDInterface {
+    fun onPause(){}
+    fun onResume(){}
```

pazos:
@Galunid: an interface is like an abstract class but with all methods unimplemented. Classes that implement an interface need to be abstract or implement all methods from the interface.

So, if you add to EPDInterface

```kotlin
fun onPause()
fun onResume()
```

you need to implement these methods in all classes that implement the interface. Everywhere, except on onyx, the implementation is empty:

```kotlin
override fun onPause() {}
override fun onResume() {}
```

Pretty much like we're doing with [some lights methods](https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/device/lights/OnyxWarmthController.kt#L103-L105), but this time is simpler as both onResume and onPause are void methods

pazos:
^^ also applies, somehow, to the logic I requested.

For example:

1. [`NTXEPDController`]() is an abstract class, with a single static method `requestEpdMode`, which performs, via reflection, an EPD update. It is cool because we can repurpose the abstract class as is everywhere and will run **without crashes** even on incompatible devices (ie:  no unhandled exceptions if reflection fails)

2. [`NTXEPDNewController`](https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/device/epd/freescale/NTXNewEPDController.kt) is used as a bridge that implements the only method of [`EPDInterface`](https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/interfaces/EPDInterface.kt) inheriting the implementation from NTXEPDControler

So, this scheme applies to all supported devices. 1 is the implementation, as an abstract class with static methods. 2 is the bridge.

If you add methods to EPDInterface you'll need to implement them on every bridge (ie: all device/epd classes that are not abstract) but you don't need to touch their implementations.

I hope it is clear enough :)

Galunid:
Yes that's pretty reasonable, what I meant in my earlier question was `default method`. It was introduced in `Java 8` I believe.
```java
interface LightsManager{
    default void toggleLights(){
        System.out.println("Toggling lights, yayy!");
    }
}
class TestLightsManager implements LightsManager{

}

public class Main{
    public static void main(String[] args){
        LightsManager lm = new TestLightsManager();
        lm.toggleLights(); // this works
    }
}
```

This way you can add methods to the interfaces without breaking backwards compatibility (which perfectly fit this case), not sure how that works in Kotlin though. Anyway that's not needed, since I need to call `preventSystemRefresh` before our every refresh ;/

P.S. Really appreciate the help!

Galunid:
Sadly `NEEDS_VIEW` is needed, otherwise system "overrides" our update.
Note to self: request A2 mode in `preventSystemUpdate`
Tomorrow I'll open pull request in base with the rest of the changes, I'm think I'm done with this part ;)

pazos:
I tried to submit a PR to your branch but it is too old :(. Could you add that? (please review the **mode** passed. I took 98 from old conversation, it might be wrong.

I've tested and saw nothing breaks in EPDTestActivity, so it is good to go from my POV but I would like to wait for the base counterpart before merging this one.

```diff
diff --git a/app/res/layout/main.xml b/app/res/layout/main.xml
index e636053..c9d1024 100644
--- a/app/res/layout/main.xml
+++ b/app/res/layout/main.xml
@@ -75,6 +75,24 @@
             android:text="freescale/ntx update"
             android:textSize="28sp" />
 
+        <TextView
+            android:layout_width="fill_parent"
+            android:layout_height="12dp" />
+
+        <TextView
+            android:id="@+id/qualcommText"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:gravity="center_horizontal"
+            android:textSize="18sp" />
+
+        <Button
+            android:id="@+id/qualcommButton"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:text="qualcomm/onyx update"
+            android:textSize="28sp" />
+
         <TextView
             android:layout_width="fill_parent"
             android:layout_height="18dp" />
diff --git a/app/src/org/koreader/launcher/EPDTestActivity.kt b/app/src/org/koreader/launcher/EPDTestActivity.kt
index 6bf345c..2059a38 100644
--- a/app/src/org/koreader/launcher/EPDTestActivity.kt
+++ b/app/src/org/koreader/launcher/EPDTestActivity.kt
@@ -10,6 +10,7 @@ import android.widget.TextView
 import android.widget.Button
 
 import org.koreader.launcher.device.epd.freescale.NTXEPDController
+import org.koreader.launcher.device.epd.qualcomm.QualcommEPDController
 import org.koreader.launcher.device.epd.rockchip.RK30xxEPDController
 import org.koreader.launcher.device.epd.rockchip.RK33xxEPDController
 
@@ -28,9 +29,11 @@ class EPDTestActivity : Activity() {
         val rk30xxDescription: TextView = findViewById(R.id.rk30xxText)
         val rk33xxDescription: TextView = findViewById(R.id.rk33xxText)
         val ntxNewDescription: TextView = findViewById(R.id.ntxNewText)
+        val qualcommDescription: TextView = findViewById(R.id.qualcommText)
         val rk30xxButton: Button = findViewById(R.id.rk30xxButton)
         val rk33xxButton: Button = findViewById(R.id.rk33xxButton)
         val ntxNewButton: Button = findViewById(R.id.ntxNewButton)
+        val qualcommButton: Button = findViewById(R.id.qualcommButton)
         val shareButton: Button = findViewById(R.id.shareButton)
 
         /* current device info */
@@ -68,6 +71,10 @@ class EPDTestActivity : Activity() {
         ntxNewDescription.text = "This button should work on modern Tolinos/Nooks and other ntx boards"
         ntxNewButton.setOnClickListener { runEinkTest(NTXTEST) }
 
+        /* qualcomm - At least Onyx Boox Nova 2 */
+        qualcommDescription.text = "This button should work on some Onyx Boox devices. At least qualcomm ones"
+        qualcommButton.setOnClickListener { runEinkTest(QUALCOMMTEST) }
+
         /* share button */
         shareButton.setOnClickListener { shareText(info.text.toString()) }
     }
@@ -84,15 +91,20 @@ class EPDTestActivity : Activity() {
             } else if (test == RK33xxTEST) {
                 info.append("rk33xx: ")
                 if (RK33xxEPDController.requestEpdMode("EPD_FULL")) success = true
-            } else if (test == NTXTEST) {
+            } else if (test == NTXTEST || test == QUALCOMMTEST) {
                 // get screen width and height
                 val display = windowManager.defaultDisplay
                 val size = Point()
                 display.getSize(size)
                 val width = size.x
                 val height = size.y
-                info.append("tolino: ")
-                if (NTXEPDController.requestEpdMode(v, 34, 50, 0, 0, width, height)) success = true
+                if (test == NTXTEST) {
+                    info.append("tolino: ")
+                    if (NTXEPDController.requestEpdMode(v, 34, 50, 0, 0, width, height)) success = true
+                } else if (test == QUALCOMMTEST) {
+                    info.append("qualcomm: ")
+                    if (QualcommEPDController.requestEpdMode(v, 98, 50, 0, 0, width, height)) success = true
+                }
             }
         } catch (e: Exception) {
             e.printStackTrace()
@@ -115,6 +127,7 @@ class EPDTestActivity : Activity() {
         private const val RK30xxTEST = 1
         private const val RK33xxTEST = 2
         private const val NTXTEST = 3
+        private const val QUALCOMMTEST = 4
         private val MANUFACTURER = android.os.Build.MANUFACTURER
         private val BRAND = android.os.Build.BRAND
         private val MODEL = android.os.Build.MODEL
```

Galunid:
@pazos Your patch works as expected, I rebased and squashed the commits, in case you need to push anything. It's still missing one thing and doesn't work as well as the default yet (sometimes menu gets stuck), so I'd prefer to hold it for a while too, not to mention that without base it'll break, since it'll get `freescale` refresh modes.

pazos:
> Since we are dealing with View I had a vague recollection that views should be updated on a separate UI thread, does that apply here (should I use something like runOnUIThread?)

Nope. The view is just a hint for the framework that tries to perform an update on the root view of the activity. On rockchip devices we use [the native window directly as the rootView](https://github.com/koreader/android-luajit-launcher/blob/master/app/src/org/koreader/launcher/MainActivity.kt#L426) and on freescale we hook a surfaceView behind the native window and pass that instead.

Unless you find a crash saying "method x tried to touch the view hierarchy (...) only from UIThread" or "Looper ..." , in that case you'll need to pass the activity instead of a view and run with `activity.runOnUiThread`

pazos (approved):
LGTM :+1:

Galunid:
Other than that I need to force the device into A2 mode, to prevent updates without user interaction, please hold it for a while.

pazos:
> Other than that I need to force the device into A2 mode, to prevent updates without user interaction, please hold it for a while.

I plan to merge https://github.com/koreader/android-luajit-launcher/pull/257 later this week . I added the resume and pause methods to DeviceHelper, which are propagated to EPD controllers.

Please consider to work on top of that branch if you want to disable/enable things onResume/onPause. :+1:

Galunid:
@pazos This PR is ready to be merged after the one in base, users will have to select the mode manually I guess, since there's no api that works within onyxsdk.

pazos:
@Galunid: if you don't mind waiting a couple of days I would like to merge #257 first and then resolve conflicts on this one.

I see that in the end you won't need to run methods onResume() and onPause(). I guess I will revert that bit from #257 unless you want to experiment with them on the future.

Galunid:
@pazos I think it's fine to keep them, I don't think there are any conflicts here actually, I'll just have to add onPause/onResume, which isn't a big deal. This can wait as long as you need.

Galunid:
Rebased, resolved conflicts, untested yet, but it shouldn't break, unless I messed up resolving conflicts ^_^
I'll report once I test it.

pazos:
@Galunid: everything looks fine. Going to merge this one and submit https://github.com/koreader/koreader-base/pull/1221#issuecomment-717218405 on top of it.

If you find a bug or something you want to correct we still have time, since we're not bumping this right now :)

mergen3107:
@Galunid 
Sorry I lost track of this in the middle of this thread, so Onyx Boox devices with Qualcomm SoC will only work in A2 mode in KOReader? Do you know whether this degrades image quality in muPDF severely compared to Normal mode?

Galunid:
@mergen3107 Nope, they will work in normal mode, but require the "A2" mode enabled in onyx application settings. Otherwise you'll get extra flashes. So no, you'll effectively have everything rendered in normal mode

mergen3107:
@Galunid 
Ah, thanks! :)

mergen3107:
@Galunid 
Hi, sorry for bumping this old thread with a somewhat off topic question, but do you know by chance how to full refresh screen on an Onyx Boox via command line?

mergen3107:
I found the answer!

`adb shell am broadcast -a onyx.android.intent.action.REFRESH_SCREEN`

PR #256, opened 2020-10-15
[RFC]: check runtime permissions before starting LuaJIT

pazos:
1. Don't send [`APP_CMD_GAINED_FOCUS`](https://github.com/koreader/android-luajit-launcher/blob/master/jni/android-main.c#L59-L62) until the runtime checks are done and permissions granted.

2. Send `APP_CMD_GAINED_FOCUS` when the permissions are granted for the first time (to resume android-main.c in https://github.com/koreader/android-luajit-launcher/blob/master/jni/android-main.c#L95)

3. Perform a initial runtime permission check before calling the NativeActivity. This is mostly cosmetic and prevents the app blinking on normal startup.

Legacy code was ok 99% of the time (because requesting a new permission would prevent the activity from gaining focus on the first time, so no LuaJIT assets were run)

The new code makes possible more complex / multiple checks and even showing UI feedback.

Also makes possible to do a "clean exit" of the activity using `finish()` without caring about the native thread.

To-do:

- extract assets can be done completely in Java, there's no need to call it from our lua state.

- remove permission checks in loader lua code, as they're not relevant anymore. The LuaJIT VM won't run unless all mandatory permissions for a given API are met.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/256)


app/res/values/strings.xml
```diff
@@ -1,3 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
+    <string name="error">Error</string>
+    <string name="error_assets">Error extracting assets. Please reinstall the application</string>
```
Frenzie, line 4:
```suggestion
    <string name="error_assets">Error extracting assets. Please reinstall the application.</string>
```

app/res/values/strings.xml
```diff
@@ -1,3 +1,9 @@
...
+    <string name="error_manage_storage">Android 11: please allow the app to manage all files.</string>
+    <string name="error_storage_permission">Insufficient permissions</string>
+    <string name="warning">Warning</string>
+    <string name="warning_storage_permission">The app needs storage permissions to work</string>
```
Frenzie, line 8:
```suggestion
    <string name="warning_storage_permission">The app needs storage permissions to work.</string>
```

app/src/org/koreader/launcher/MainActivity.kt
```diff
@@ -697,4 +718,22 @@ class MainActivity : BaseActivity() {
     private fun toMin(milliseconds: Int): Int {
         return if (milliseconds > 0) milliseconds / (1000 * 60) else 0
     }
+
+    private fun showPermissionDialog(message: String, isError: Boolean) {
+        val builder = AlertDialog.Builder(this@MainActivity)
+        val title = resources.getString(if (isError) R.string.error else R.string.warning)
+        builder.setTitle(title).setMessage(message).setPositiveButton("Ok", { _, _ ->
```
Frenzie, line 725: I think my Android (and us too) says "OK".
```suggestion
        builder.setTitle(title).setMessage(message).setPositiveButton("OK", { _, _ ->
```

Frenzie (approved):
lgtm

pazos:
> Don't send APP_CMD_GAINED_FOCUS until the runtime checks are done and permissions granted.

That's a dead end, as the app will trigger an ANR is the windows doesn't get focused in 5seconds. So no fancy dialogs everywhere, I'm afraid. :(

PR #253, opened 2020-10-08
bump to API 30

pazos:
So Google changed his scoped storage enforcement for Android 11+. It is indeed cool because killing the File API was nosense: https://developer.android.com/training/data-storage/manage-all-files

It is a draft, as it requires changes in the build server and the toolchain:

```diff
diff --git a/toolchain/Makefile b/toolchain/Makefile
index 684b51e..3e12a83 100644
--- a/toolchain/Makefile
+++ b/toolchain/Makefile
@@ -19,7 +19,7 @@ android-sdk:
        pushd $(SDK_DIR)/tools/bin && \
                yes | ./sdkmanager --update && \
                yes | ./sdkmanager --licenses && \
-               ./sdkmanager "platform-tools" "build-tools;28.0.3" "platforms;android-28" "patcher;v4" && \
+               ./sdkmanager "platform-tools" "build-tools;30.0.2" "platforms;android-30" "patcher;v4" && \
                popd
        rm -f $(SDK_DIR)/$(SDK_TARBALL)
```

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/253)

PR #257, opened 2020-10-18
refactor MainActivity, backport android_native_glue changes from r22

pazos:
```
Fixed: timeout logic (system timeout wasn't restored on activity paused)
Fixed: code styling issues
Moved: clipboard, device and timeouts to helper classes
Added: utils to manage permissions
Added: manage battery optimizations (API23+)
Changed: non-mandatory permission dialogs are available from lua
Changed: non e-ink devices don't use a SurfaceView
Changed: theme is now material design on compatible android versions (API21+)
Changed: backported android_native_glue changes from NDK r22
Removed: useless warning in AssetsUtils
```

cf: https://github.com/koreader/android-luajit-launcher/pull/253#issuecomment-711154573

It doesn't involve **massive** changes in the code but uses modern "material design" dialogs if possible.

Pinging both @Galunid and @zwim: can you test that everything **still** works fine on your epd devices?

@zwim: lights still work?, epd updates fires fine?
@Galunid: does it build on docker container?, do you see the new light dialog with material style?, Is it legible?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/257)


Galunid:
@pazos I'm building it, I'll let you know what happens. Anything I should pay attention in particular?

pazos:
> Anything I should pay attention to in particular?

A screenshot of the light dialog would be great. Other than that just check that nothing crashes :)

Galunid:
There was some problem with mupdf, it seems I need to clean build (it happened on emulator too), I'll test it in ~1h

Galunid:
Doesn't seem to crash. I'll continue to use it in case something comes up, but all seems good.
### Screenshot:
Screenshot_20201018-170949

zwim:
@pazos: I will check it, but I will need until Friday, because I have more work due to Corona :-(

pazos:
Last commit on this branch contains changes into the `android_native_app_glue` that affect the frontend.

We were reacting to `APP_CMD_WINDOW_REDRAW_NEEDED` even when that callback was never fired. So, to prevent undefined behaviour in KOReader, the following change is needed:

```diff
diff --git a/frontend/device/android/device.lua b/frontend/device/android/device.lua
index c5530ffd..e9d93488 100644
--- a/frontend/device/android/device.lua
+++ b/frontend/device/android/device.lua
@@ -135,9 +135,14 @@ function Device:init()
             if ev.code == C.APP_CMD_SAVE_STATE then
                 return "SaveState"
             elseif ev.code == C.APP_CMD_GAINED_FOCUS
-                or ev.code == C.APP_CMD_INIT_WINDOW
-                or ev.code == C.APP_CMD_WINDOW_REDRAW_NEEDED then
+                or ev.code == C.APP_CMD_INIT_WINDOW then
                 this.device.screen:_updateWindow()
+            elseif ev.code == C.APP_CMD_WINDOW_REDRAW_NEEDED then
+                logger.dbg("APP_CMD_WINDOW_REDRAW_NEEDED")
+            elseif ev.code == C.APP_CMD_WINDOW_RESIZED then
+                logger.dbg("APP_CMD_WINDOW_RESIZED")
+            elseif ev.code == C.APP_CMD_CONTENT_RECT_CHANGED then
+                logger.dbg("APP_CMD_CONTENT_RECT_CHANGED")
             elseif ev.code == C.APP_CMD_CONFIG_CHANGED then
                 -- orientation and size changes
                 if android.screen.width ~= android.getScreenWidth()
```

The rest of changes shouldn't affect KO behaviour, but it is cool to log them anyways.

zwim:
@pazos: This PR seems to work. I have noticed no problem.

pazos:
> This PR seems to work. I have noticed no problem.

Thank you for testing!. Some questions:

Did you applied https://github.com/koreader/android-luajit-launcher/pull/257#issuecomment-712259755? If not I assume you didn't find any difference on the behaviour (ie: when returning to the activity, when closing the frontlight dialog ...)

Does the frontlight dialog looks any different?. It shouldn't if the firmware is older than Lollipop. But if it looks the same and works the same is a big win (and what's expected based on what I saw on old emulators)

zwim:
On 4.4.2 (Tolino) I did not notice any difference.
Android dialog looks the same.

pazos:
@Frenzie: still waiting your review :)

NiLuJe:
I think he's still in the "just moved" frenzy (ha!), and may not yet have a stable landline ;).

pazos:
> I think he's still in the "just moved" frenzy (ha!), and may not yet have a stable landline ;).

I can wait. Is just that I have another pile or changes to submit and don't want to push on top of this, yet.

Frenzie:
I have a stable line, but no time. But I'll check if it's short.

app/src/org/koreader/launcher/MainApp.kt
```diff
@@ -1,111 +1,72 @@
...
     override fun onCreate() {
         super.onCreate()
-        getAppInfo()
-        Log.i(name, "Application started\n" + formatAppInfo())
+        val pm = packageManager
+        val packageInfo = pm.getPackageInfo(packageName, 0)
+        val applicationInfo = pm.getApplicationInfo(packageName, 0)
+        val hasLargeHeap = (applicationInfo.flags and ApplicationInfo.FLAG_LARGE_HEAP != 0)
+        val isDebuggable = (applicationInfo.flags and ApplicationInfo.FLAG_DEBUGGABLE != 0)
+        val isSystemApp = (applicationInfo.flags and ApplicationInfo.FLAG_SYSTEM == 1)
```
Frenzie, line 30: What are these for exactly? :-)

Frenzie (approved):
Looks fairly straightforward; didn't have time to check in the Android docs for some of the new (at least to us) calls/constants of which I don't know exactly what they're doing though.

app/src/org/koreader/launcher/MainApp.kt
pazos: Build flags. Used for logging mainly. They're all booleans.

hasLargeHeap can be removed, because it was added to log the size of the dalvik VM heap when we disabled largeHeap app property. Nobody reported a crash due to a not enough dalvik heap size.

pazos:
> Looks fairly straightforward; didn't have time to check in the Android docs for some of the new (at least to us) calls/constants of which I don't know exactly what they're doing though.

Probably there's nothing very new. Just a few minor changes:

- We start activities (dictionary apps, browser, ...) using NO_ANIMATION flag, to prevent the animation during activity transition.
- Add a couple of api calls to check if the battery is optimized and to change that.
- Permission code was refactored to be easily upgraded for new api requirements.

The size of the APK grows about 800kb because of the new required library "androidx.appcompat:appcompat". That's probably the last app size grow that's caused by android dependencies.

PR #258, opened 2020-11-03
move eink constants from base

pazos:
Rationale:

Each time a new epd driver is added we need to tell KOReader which epd modes, waveforms and delays to use. That involved changes in https://github.com/koreader/koreader-base/blob/master/ffi/framebuffer_android.lua.

This PR adds a new function `android.getEinkConstants()` to retrieve these values in framebuffer_android.lua, so hopefully changes there are not needed (or at least will be tiny).

@Galunid  made some valid points in https://github.com/koreader/koreader-base/pull/1221#issuecomment-717258983, about moving these constants to each EPD driver, but IMO they belong to the "lua bridge" as they describe the behaviour expected by the frontend and not the **only valid** way of drive each eink screen.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/258)

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1993,6 +1993,38 @@ local function run(android_app_state)
         end)
     end
 
+    android.getEinkConstants = function()
+        local isEink, platform = android.isEink()
+        if not isEink then return end
+        if platform == "rockchip" then
+            -- rockchip devices
+            local full, partial, fast, auto = 1, 2, 3, 4
+            return full, partial, auto, auto, fast
+        else
+            -- freescale/qualcomm
+            local mode_full, mode_partial = 32, 0
+            local wf_du, wf_gc16 = 1, 2
+            local delay_fast = 0
+            local fast, partial = wf_du + mode_partial, wf_gc16 + mode_partial
+            local full, full_ui, partial_ui, delay, delay_ui
+            if platform == "freescale" then
+                local wf_regal = 7
+                full = wf_gc16 + mode_full
+                full_ui = wf_regal + mode_full
+                partial_ui = wf_regal + mode_partial
+                delay, delay_ui = 0, 0
+            elseif platform == "qualcomm" then
+                local wf_regal = 6
+                local mode_wait = 64
+                full = wf_gc16 + mode_full + mode_wait
+                full_ui = wf_regal + mode_full
+                partial_ui = wf_gc16 + mode_partial
+                delay, delay_ui = 250, 100
+            end
+            return full, partial, full_ui, partial_ui, fast, delay, delay_ui, delay_fast
+        end
+    end
+
     android.needsWakelocks = function()
         return JNI:context(android.app.activity.vm, function(jni)
             return jni:callIntMethod(
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -1996,17 +1996,17 @@ local function run(android_app_state)
     android.getEinkConstants = function()
         local isEink, platform = android.isEink()
         if not isEink then return end
+        local full, partial, full_ui, partial_ui, fast, delay, delay_ui, delay_fast
         if platform == "rockchip" then
-            -- rockchip devices
-            local full, partial, fast, auto = 1, 2, 3, 4
-            return full, partial, auto, auto, fast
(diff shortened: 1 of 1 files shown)


assets/android.lua
```diff
@@ -1993,6 +1993,38 @@ local function run(android_app_state)
         end)
     end
 
+    android.getEinkConstants = function()
+        local isEink, platform = android.isEink()
+        if not isEink then return end
+        if platform == "rockchip" then
+            -- rockchip devices
+            local full, partial, fast, auto = 1, 2, 3, 4
+            return full, partial, auto, auto, fast
```
Galunid, line 2002: Wouldn't it be a good idea to add "dummy"/0 variables, so that both returns match?
`return full, partial, full_ui, partial_ui, fast, delay, delay_ui, delay_fast`

Other than that, it looks fine.

Galunid:
It works fine after adding #259.I already pushed the rest of the needed things in base.

assets/android.lua
pazos: on Rk devices just `full` is used. The rest are ignored (I added them here to document what they do - and because it was documented before :p) . The rest of values will be nil.
on Non eink devices the function is not called right now, but even if called all values will be nil.

I think is fine as is, but can add more verbose comments if you wish :smile:

Galunid: Well, I was thinking of adding those delays, but if you don't want to, we can keep it as it is, and add comment in base explaining why they may be `nil`

Galunid (approved):

assets/android.lua
pazos: Mmm, I'm afraid I don't follow. 

We don't use any value on non-eink devices
We don't use any delay on Rockchip devices

In any case we use different values on freescale devices for **full-blinking** updates based on their support for full refreshes or not. Actually the value is hardcoded to `500` for Nooks. Shall I handle the eink vs einkFull split here too?

Galunid: TL;DR I think it's better to do 
`return full, partial, auto, auto, fast, 0, 0, 0` instead of `return full, partial, auto, auto, fast`

If someone was developing stuff for rockchip devices that are einkFull (there are none yet, but who knows what happens in the future), and used those delays, things would implode. I think it's just better to return more than less, since we could do just `return full`. When you read
`local full, partial, full_ui, partial_ui, fast, delay_page, delay_ui, delay_fast = android.getEinkConstants()`
you'd expect that those values are not nil, since we don't have any nil guards in the code, and function has "Constants" in name. So it may end up misleading.

pazos: Ufff :p 

I understand the nil guards argument but don't like your proposal. Give me some time to figure out how to solve this without too much boilerplate code :+1:

pazos: @Galunid: updated.

Added a guard against nil modes.
Added a comment that points to rockchip epd update routine.

I'm afraid I didn't found a better way of doing this. There's no purpose of adding a delay to rockchip devices since they won't use it anyways. Its epd update routine accepts a single argument **mode**.

While the rest of devices accept the arguments: mode, delay, x, y, width, height.
These devices can also be called with only the two first arguments **mode** and **delay** but that's us calculating the width and height and setting x and y to 0 ourselves in `assets/android.lua`.

Galunid:
Alright, since it seems rockchip won't ever use delay, then I'm fine with this. Wouldn't it be a good idea to add some kind of log if `mode == nil` in `android.einkUpdate`? If mode is nil, then something must have gone really wrong. :)

pazos:
mode won't be nil unless called on purpose. The backend takes care of it. Indeed the backend takes care of all abstraction.

what koreader does is (pseudocode):

```lua
if eink then
    updateFull(mode)
elseif eink and einkFull
    updateFull(mode) + updateRegion(mode,delay,x, y, w, h)
end
```
`getEinkConstants` will return nil **only** if *not eink* and we do nothing with these constants (all of them nil) if *not eink*
The nil guards avoid Java VM crashes when calling `android.einkUpate` with a nil mode (it wasn't guarded, the rest of arguments were)

PR #261, opened 2020-11-08
use proper flavors, enable minify on release builds, adaptative icons

pazos:
## use proper flavors:

- Merged base.xml and full.xml together.
- rocks flavor changes:
    - add name, mimetype to apk download (needed to show the notification on some roms)
    - add a permission to install the APK directly from the notification (needed in API25+)

- fdroid flavor doesn't use the new permission.
- fix some lint warnings on OnyxEPDController

## adaptative icons:

- Include adaptative icons
- use BuildConfig instead of resource strings
- change android Tv banner to match the same design as the icon
- debug builds use a different name: "KOReader-debug"

## minify on release builds:

- Shrinks the APK by 1MB of useless android classes.
- annotation `@Keep` used to avoid removing classes used by JNI or by reflection.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/261)


NiLuJe:
>       * add a permission to install the APK directly from the notification (needed in API25+)

:tada: \o/

pazos:
Shall we restrict APK downloads to Wifi? Or it is ok to download them on mobile/metered networks?

NiLuJe:
That's not really a' issue on my end, but then again we probably have some
of the cheapest mobile data plans out there in France, so, eh ;p.

Frenzie:
It's a user initiated action, not something automated.

pazos:
Good. There're plenty of switches for roaming, metered, transport... Good that we can avoid writting glue code for all of them.

About the PR. Works as expected in the rocks flavor (except I cannot install the new package as signatures don't match). Need to check that fdroid flavor still builds fine. It should.

pazos:
Ready for review

Tested all flavor combos. Also barely tested the minify/R8 stuff. App doesn't crash after adding `@Keep`. Probably some bugs can be catched in the wild by final users and never by me and the few devices/emulators that I have available,  as always happens on Android.

Makefile
```diff
@@ -5,6 +5,12 @@ ifdef ANDROID_ARCH
 	endif
 endif
 
+ifdef ANDROID_FLAVOR
+	ifeq ($(ANDROID_FLAVOR), fdroid)
+		FLAVOR?=Fdroid
```
Frenzie, line 10: Capitalization on purpose? I'd expect fdroid or FDroid (as in F-Droid).

app/build.gradle
```diff
@@ -16,7 +16,7 @@ android {
         release {
             manifestPlaceholders.APP_NAME = projectName
             buildConfigField 'String', 'APP_NAME', '"' + projectName + '"'
-            minifyEnabled false
+            minifyEnabled true
```
Frenzie, line 19: What does it do precisely?

Frenzie (approved):
lgtm

Makefile
pazos: Yup, gradle tasks are capitalized. So the flavors are x86, arm, fdroid, rocks but the task to assemble them are something like X86Fdroid or ArmRocks.

app/build.gradle
pazos: All the java classes used are converted to dalvik bytecode and stored in `classes.dex`. That includes our own classes, which are fully used, and classes from implementation libraries, which we barely use.

Minify does check the signature of all compiled classes and strips out the classes with non used signatures. There's another thing called `shrinkResources` which does exactly the same for resources, but in our case it saves only a few KB.

Until now it didn't make much sense since we had to keep all the stuff used from lua in a proguard file, but with the androidx annotation makes sense , specially now with android compat methods, which we just use to show a few dialogs and setup a theme but contains a lot of useless things for us.

pazos:
Apparently it is enough to `@Keep` the MainActivity, as it is the only part of the code called by JNI. I'm not sure about the abstract EPD classes, since they're used in `EPDTestActivity` and their methods shouldn't be stripped.

Anyways, will do a few tests tonight on my hw and merge ASAP if I find no crashes. Lets see how it behaves in the wild.

Issue #262, opened 2020-11-12
cannot build luajit


ildar:
excerpt:
```
+ make -C luajit-2.0 install 'HOST_CC=gcc -m32' 'CFLAGS=-O2 -pipe' 'HOST_CFLAGS=-O2 -pipe -mtune=generic' LDFLAGS= HOST_LDFLAGS= TARGET_CFLAGS= TARGET_LDFLAGS= 'CROSS=/tmp/a/ndk/18.1.5063045//toolchains/arm-linux-androideabi-4.9/prebuilt/*/bin/arm-linux-androideabi-' 'TARGET_FLAGS=--sysroot /tmp/a/ndk/18.1.5063045//platforms/android-19/arch-arm -march=armv7-a -mfloat-abi=softfp -Wl,--fix-cortex-a8' TARGET_SYS=Linux DESTDIR=/usr/src/android-luajit-launcher/jni/luajit/build/armeabi-v7a PREFIX=
make: Entering directory '/usr/src/android-luajit-launcher/jni/luajit/luajit-2.0'
==== Building LuaJIT 2.1.0-beta3 ====
make -C src
make[1]: Entering directory '/usr/src/android-luajit-launcher/jni/luajit/luajit-2.0/src'
clang: warning: -Wl,--fix-cortex-a8: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -Wl,--fix-cortex-a8: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -Wl,--fix-cortex-a8: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -Wl,--fix-cortex-a8: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -Wl,--fix-cortex-a8: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -Wl,--fix-cortex-a8: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -Wl,--fix-cortex-a8: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -Wl,--fix-cortex-a8: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -Wl,--fix-cortex-a8: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -Wl,--fix-cortex-a8: 'linker' input unused [-Wunused-command-line-argument]
[...]

PR #263, opened 2020-11-12
Make explicit ./mk-luajit.sh is intended to be used with NDK r15c

pazos:
cf. https://github.com/koreader/android-luajit-launcher/issues/262

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/263)


ildar:
Wait. The problem may be trivial to fix. Gimme a few days to try

pazos:
> Wait. The problem may be trivial to fix. Gimme a few days to try

Is not a problem. We need to keep old toolchain too keep compatibility with ICS, so we aren't going to change `mk-luajit.sh`

If you're going to use the project with a different codebase than KOReader I guess you can fork and change `mk-luajit.sh` there.

PR #264, opened 2020-11-13
lzma compression, libs as raw assets

pazos:
Last PR for the moment, I promise.

~~Add a new flavor dimension: COMPRESSION, with two values: zip & lzma~~

~~Zip is basically what we're doing now. It works and it is compatible with all devices.
Lzma is new, using the awesome un7zip library, and reduces the APK size 10MB. The library has a minVersion of API15 but should also work in API14.~~

~~With this PR in the program is still using the zip flavor. We can switch to lzma again in the future.~~

**Edit:** switch back to lzma, embedding our own version of un7zip library (and using upstream 7z version)

I will submit the change to the frontend too, so you can see the entire picture.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/264)


Frenzie (commented):
Sounds fine as a research project and wrt inclusion why not, but if old devices don't support it I don't think adding complexity for build infra & users sounds very appealing?

Frenzie:
Sorry, slightly misinterpreted this line in a hurry.

> The library has a minVersion of API15 but should also work in API14.

pazos:
> Sounds fine as a research project and wrt inclusion why not, but if old devices don't support it I don't think adding complexity for build infra & users sounds very appealing?

We shouldn't build both of them. If lzma works fine in all APIs as is we can switch to it, at least in the rocks flavor (I think F-Droid is against picking prebuilt libraries so they need to still use zip)

About the minSdkVersion, it is a bit contradictory. The repo https://github.com/hzy3774/AndroidUn7zip signals `API14+` but the library is built against API15. I actually skip the check with `<uses-sdk tools:overrideLibrary="com.hzy.lib7z" />` but until somebody proves that it works on API14 the result is unknown.

Actually, if F-Droid rejects prebuilt libraries **and** un7zip doesn't work on API14 I might consider to build the library from source (at least the little part we use). If the library fails to extract on other APIs on low-end devices the better move is stay as is :)

pazos:
> Actually, if F-Droid rejects prebuilt libraries and un7zip doesn't work on API14 I might consider to build the library from source (at least the little part we use). If the library fails to extract on other APIs on low-end devices the better move is stay as is :)

@Frenzie: It was easier than I expected.

I removed the lzma/zip flavors and just revert back to pre https://github.com/koreader/android-luajit-launcher/pull/174 era, but keeping the error checking and the asset logic in kotlin and using un7zip to handle 7z decompression. I restored the 7z submodule, which is newer than the one bundled with un7zip.

This way everything is build from source, with the same minimum api. It should work **just fine** :tm: in all supported devices. It is a matter of measuring decompression speeds between this and zips.

pazos:
@Frenzie: ready for review!

Moved all the mess to Assets class (including the progress dialog/dots). While it is possible to create a single method to extract recursivelly all assets - including compressed files - it clashes with our way to check if assets are the same version, so the proposed solution is the simplest to keep my hair black :smile:

.gitmodules
```diff
@@ -1,3 +1,6 @@
-[submodule "luajit-2.0"]
-	path = jni/luajit/luajit-2.0
-	url = https://github.com/LuaJIT/LuaJIT
+[submodule "jni/luajit/luajit"]
+	path = jni/luajit/luajit
+	url = https://github.com/LuaJIT/LuaJIT.git
+[submodule "jni/lzma/7z"]
+	path = jni/lzma/7z
+	url = https://github.com/kornelski/7z.git
```
Frenzie, line 6: Déjà vu. :-P

app/src/org/koreader/launcher/Assets.kt
```diff
@@ -0,0 +1,233 @@
...
+import android.view.ViewGroup
+import android.widget.ProgressBar
+import androidx.core.content.ContextCompat
+import java.io.*
+
+class Assets {
+    companion object {
+        private const val TAG = "AssetsHelper"
+        private const val CONTAINER = "7z"
+        private const val BUFFER_SIZE = 1024
```
Frenzie, line 17: Sounds a bit small?

Frenzie (approved):
Sure, lgtm :+1:

app/src/org/koreader/launcher/Assets.kt
pazos: Yup. TBH it wasn't used in KO, just by the example code. Now it is used to copy libraries. I have no idea which is a good value.

NiLuJe: When in doubt, a page's worth is usually a sensible default. (That'd be 4KB?).

pazos: The function is https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-input-stream/copy-to.html

```kotlin
fun InputStream.copyTo(
    out: OutputStream,
    bufferSize: Int = DEFAULT_BUFFER_SIZE
): Long
```

So maybe we can omit bufferSize and let kotlin handle that with its own `DEFAULT_BUFFER_SIZE`?

Kotlin docs aren't very specific on this regard: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/-d-e-f-a-u-l-t_-b-u-f-f-e-r_-s-i-z-e.html

NiLuJe: Sounds good to me, yeah ;).

Frenzie: Yup! :-)

PR #266, opened 2020-11-28
get absolute path of external sdcard

pazos:
Based on a hack, but it is the only way to get the path and it is extensively used elsewhere.

It should be reliable enough on 4.4+ and should produce a catched exception on 4.0 - 4.3.
At the same time on 4.0 - 4.3 probably there're no restrictions to move between volumes using our filemanager.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/266)

### DIFF
--- a/app/src/org/koreader/launcher/utils/FileUtils.kt
+++ b/app/src/org/koreader/launcher/utils/FileUtils.kt
@@ -4,9 +4,11 @@ import android.content.ContentResolver
 import android.content.Context
 import android.net.Uri
 import android.os.Build
+import android.os.Environment
 import android.os.ParcelFileDescriptor
 import android.provider.MediaStore
 import android.system.Os
+import androidx.core.content.ContextCompat
 import org.koreader.launcher.Logger
 import java.io.*
 import java.util.*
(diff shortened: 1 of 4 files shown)


app/src/org/koreader/launcher/utils/FileUtils.kt
```diff
@@ -59,6 +61,27 @@ object FileUtils {
...
+     * @param context
+     * @return absolute path or "null"
+     */
+
+    fun getExtSdcardPath(context: Context): String {
+        val ctx = context.applicationContext
+        val volumes: Array<out File> = ContextCompat.getExternalFilesDirs(ctx, null)
+        return if (Environment.getExternalStorageState() == Environment.MEDIA_MOUNTED) {
+            try {
+                volumes[1].absolutePath.replace("/Android/data/${ctx.packageName}/files", "")
```
Frenzie, line 76: Assuming this is sufficiently dependable. ;-)

Frenzie (approved):
lgtm

app/src/org/koreader/launcher/utils/FileUtils.kt
pazos: Not sure, but documentation isn't clear enough: https://developer.android.com/reference/android/content/Context#getExternalFilesDirs(java.lang.String)

We use ContextCompat instead of Context, so they array will return only one File up to android 4.3 and thus accesing the second File of the array will throw an exception.

In theory there're a lot things that can be checked and wrapped between if...else, but the one liner looks elegant and can be fixed later, if is buggy on a specific device (hint: it will)

PR #268, opened 2020-12-02
int to bool

pazos:
Since we have `JNI:callBooleanMethod`.

~~Untested yet.~~

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/268)

### DIFF
--- a/app/src/org/koreader/launcher/Clipboard.kt
+++ b/app/src/org/koreader/launcher/Clipboard.kt
@@ -44,12 +44,12 @@ class Clipboard(activity: Activity) {
         return result.value ?: ""
     }
 
-    fun hasClipboardText(): Int {
+    fun hasClipboardText(): Boolean {
         val clipdata = clipboard.primaryClip
         return if (clipdata != null) {
             val number = clipdata.itemCount
-            if (number > 0) 1 else 0
-        } else 0
+            (number > 0)
+        } else false
     }
 
     fun setClipboardText(activity: Activity, text: String) {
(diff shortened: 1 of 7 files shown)


Frenzie (approved):
Hm, how come we didn't do this before? JNI update or something?

pazos:
> Hm, how come we didn't do this before? JNI update or something?

both `callBooleanMethod` and the first usage of `return callIntMethod == 1` are in https://github.com/koreader/android-luajit-launcher/commit/d0e2274673d683fbbd7e6a2e11d46f2b60b440d5. Since then I think we basically copy-pasted code based on other lua functions.

pazos:
@Frenzie: now tested. It works fine :tada: 

Added a few fixes for bugs in previous commits. I will rebase when you think is ready!

Frenzie (approved):
All good to me, that splashscreen thing is a bit random in this PR though. :-P

pazos:
> All good to me, that splashscreen thing is a bit random in this PR though. :-P

Indeed. Also now I find that some true/false values are swapped. No idea why. I will split this PR in two and wait a bit for the int -> bool changes. The rest of things work fine and can be merged soon.

pazos:
Something fishy happens with this commit. 1 used to be true and ~= 1 false. Now all things are inversed :cry:

pazos:
> Something fishy happens with this commit. 1 used to be true and ~= 1 false. Now all things are inversed

It turns out that any boolean was true, because jboolean is uint8_t and gets casted to a lua number, and no number is nil :) 

Works fine now.

PR #271, opened 2020-12-08
return to target API28

pazos:
Keeping compileSdk in 30, to repurpose parts of the code, the new docker image and keep updated tools. 

Related to https://github.com/koreader/koreader/issues/6953
We can't bump targetSdk until `sdcv` is replaced by a library.

Refs:

https://developer.android.com/about/versions/10/behavior-changes-10#execute-permission

Somehow this is not enforced on some emulator versions and doesn't apply to custom LineageOS builds which set selinux to permissive. :facepalm: :man_facepalming: :woman_facepalming:

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/271)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 
 android {
-    compileSdkVersion targetSdk
+    compileSdkVersion compileSdk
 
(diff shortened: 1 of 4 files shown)


Frenzie (approved):
sigh

NiLuJe:
> sigh

Ditto :/

PR #278, opened 2020-12-22
Attempt to modernize the mcode alloc hackery...

NiLuJe:
AFAICT, the mmap trick is *way too late* to be any good, so, get rid of
it, and mention that a proper approach would require moving it to the C
frontend, which in turn would require a shared build of LuaJIT, instead
of embedding it statically.

Also, use modern syntax for jit.opt, and update to the current defaults,
to actually make sure it won't try to grow it later.

This is milldly related to https://github.com/koreader/koreader-base/pull/1263 & https://github.com/koreader/koreader/pull/7023

Previous history of said piece of hackery: #33 & #40

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/278)

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -4,28 +4,15 @@ Java Native Interface (JNI) wrapper.
 @module android
 ]]
 
+-- Attempt to grab the full 512K of maxmcode in one go on startup,
+-- to avoid mcode_alloc failures later on at runtime...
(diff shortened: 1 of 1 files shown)


NiLuJe:
Here be giant fiery dragons, obviously.

pazos (approved):
Lets see how many crashes we catch!

PR #279, opened 2020-12-23
Re-tune the mcode hack

NiLuJe:
Because 512K really doesn't work :/.

Also include a stupid hack so that base can detect Android -_-".

----

As the commit log says, I may ultimately try the mmap trick the proper way, but since it requires a shared lib and I'm not at all familiar with all that would entail, not today.

For now, this should just make sure performance doesn't regress because of #278 ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/279)

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -4,12 +4,14 @@ Java Native Interface (JNI) wrapper.
 @module android
 ]]
 
--- Attempt to grab the full 512K of maxmcode in one go on startup,
+-- Attempt to grab the full maxmcode region in one go on startup,
 -- to avoid mcode_alloc failures later on at runtime...
 -- c.f., https://www.freelists.org/post/luajit/Android-performance-drop-moving-from-LuaJIT201-LuaJIT202
 -- The other workaround mentioned earlier in that thread would require moving to a shared LuaJIT build,
 -- and move the mmap hackery in jni/android-main.c before a dlopen of the luaJIT lib...
-jit.opt.start("sizemcode=512")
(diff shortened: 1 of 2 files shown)


Frenzie (approved):
ho hum

pazos:
>  I may ultimately try the mmap trick the proper way, but since it requires a shared lib and I'm not at all familiar with all that would entail, not today.

the libluajit.so built as part of this repo should be called libmain.so instead. Because it is not a shared library of luaJIT but the whole entry point for a native app + native glue + a bunch of links to access stuff using ffi + a static luaJIT library.

Not very familiar either but should be possible parts of stuff we do using ffi to plain C, get the path to the shared LuaJIT library in android-main.c and dlopen it, without affecting all the other behaviour. Still no idea of what the mmap trick involves and how that relates to shared vs static libraries.

pazos (approved):

NiLuJe:
@pazos: But would that allow us entire control of *when* it's loaded?

Because the whole point of this hack would be to game the linker, and ensure the library is mapped far, far, far away from the entry point, so that LuaJIT has a better chance of finding free memory in the very restricted range around where its alloc function points.

NiLuJe:
Hence the idea of:

* Large mmap
* dlopen luajit
* unmap

NiLuJe:
Oh, wait, I misunderstood what you said. You mean that `android-main.c` ends up built as `libmain.so`, right?

(And, yes, it currently links against a static copy of LuaJIT, which is what I was talking about ^^).

PR #283, opened 2020-12-31
Tackle mcode_alloc issues once and for all

NiLuJe:
Okay, ignore the insane commit history, because this went through a *lot* of experimentation, especially since I found out my old Nexus 5 on 7.1 didn't give a damn about the initial workaround that yielded decent results on Android 10...

Anyway.

* Move to a shared luajit library. This gets rid of the weird hack where the launcher was named libluajit for... reasons (related, ironically, to static linking, and old broken API levels).
* Simplified the dlopen replacement given that fact, and actually implemented the system libs blacklist properly (not that it should be necessary).
* And, as far as the LuaJIT mcode_alloc issues go:
* Make sure the launcher requests a single 512K block right at startup.
* Patch LuaJIT so that:
  * Flushing all traces doesn't unmap the mcode area, but instead zero it, so we don't "lose" our hard-earned mmap, given how tighly packed the virtual memory layout is on Android.
  * Reserve a 1MB area *inside* LuaJIT's address space via a global non-const, initialized array. This pretty much guarantees it'll be in the jumprange ;).
  * Make sure the very first alloc will punch a hole in *that* array, via MAP_FIXED

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/283)

### DIFF
--- a/jni/Android.mk
+++ b/jni/Android.mk
@@ -40,9 +40,11 @@ LOCAL_PATH := $(BASE_PATH)
 LOCAL_MODULE := luajit-launcher
 LOCAL_SRC_FILES := android-main.c
 LOCAL_STATIC_LIBRARIES := android_native_app_glue
-# NOTE: I don't actually want to link to it, I just want the headers in the searchpath...
+# NOTE: By default, we link against LuaJIT directly.
 #LOCAL_SHARED_LIBRARIES := luajit
+# But, we can also dlopen it, as a remnant of mcode_alloc workaround experimentations...
 LOCAL_C_INCLUDES := $(LOCAL_PATH)/luajit/build/$(TARGET_ARCH_ABI)/include
+LOCAL_CFLAGS += -DKO_DLOPEN_LUAJIT
 
 # Adding libraries that we plan to use via FFI here isn't strictly necessary anymore
 LOCAL_EXPORT_LDLIBS := -lm -ldl -llog -landroid
--- a/jni/Android.mk
+++ b/jni/Android.mk
@@ -26,9 +26,10 @@ include $(LOCAL_PATH)/luajit/Android.mk
 include $(CLEAR_VARS)
 
 # Dummy target that will ensure we ship the LuaJIT shared library
-LOCAL_PATH := $(BASE_PATH)
-include $(LOCAL_PATH)/foo/Android.mk
-include $(CLEAR_VARS)
+# NOTE: Only necessary if we want to dlopen LuaJIT, which isn't the default ;).
+#LOCAL_PATH := $(BASE_PATH)
+#include $(LOCAL_PATH)/foo/Android.mk
+#include $(CLEAR_VARS)
 
 # lib7z shared library
 LOCAL_PATH := $(BASE_PATH)
@@ -40,11 +41,11 @@ LOCAL_PATH := $(BASE_PATH)
 LOCAL_MODULE := luajit-launcher
 LOCAL_SRC_FILES := android-main.c
 LOCAL_STATIC_LIBRARIES := android_native_app_glue
-# NOTE: By default, we link against LuaJIT directly.
-#LOCAL_SHARED_LIBRARIES := luajit
-# But, we can also dlopen it, as a remnant of mcode_alloc workaround experimentations...
-LOCAL_C_INCLUDES := $(LOCAL_PATH)/luajit/build/$(TARGET_ARCH_ABI)/include
-LOCAL_CFLAGS += -DKO_DLOPEN_LUAJIT
+# NOTE: By default, we link against the shared LuaJIT library directly.
+LOCAL_SHARED_LIBRARIES := luajit
+# But, we can also dlopen it, as a remnant of mcode_alloc experimentations...
+#LOCAL_C_INCLUDES := $(LOCAL_PATH)/luajit/build/$(TARGET_ARCH_ABI)/include
+#LOCAL_CFLAGS += -DKO_DLOPEN_LUAJIT
 
 # Adding libraries that we plan to use via FFI here isn't strictly necessary anymore
 LOCAL_EXPORT_LDLIBS := -lm -ldl -llog -landroid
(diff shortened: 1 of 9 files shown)


NiLuJe:
This works out fine on both of my test devices, but I'm obviously waiting with bated breath for people to break it on Android 4 ;p.

NiLuJe:
(Oh, also: a debug build will add some logging to LuaJIT to make tracking the mcode life cycle easier).

jni/Android.mk
```diff
@@ -20,25 +20,34 @@ LOCAL_PATH := $(BASE_PATH)
...
+# NOTE: By default, we link against the shared LuaJIT library directly.
+LOCAL_SHARED_LIBRARIES := luajit
+# But, we can also dlopen it, as a remnant of mcode_alloc experimentations...
+#LOCAL_C_INCLUDES := $(LOCAL_PATH)/luajit/build/$(TARGET_ARCH_ABI)/include
+#LOCAL_CFLAGS += -DKO_DLOPEN_LUAJIT
 
-# remember to add libraries here that you plan to use via FFI:
-LOCAL_EXPORT_LDLIBS := -lm -llog -landroid
+# Adding libraries that we plan to use via FFI here isn't strictly necessary anymore
+LOCAL_EXPORT_LDLIBS := -lm -ldl -llog -landroid
```
Frenzie, line 51: Should it still stay then? What do you mean by that exactly?

jni/android-main.c
```diff
@@ -109,8 +113,73 @@ void android_main(struct android_app* state) {
...
+    void* p = mmap(NULL, map_size, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_NORESERVE, -1, 0);
+    if (p == MAP_FAILED) {
+        LOGE("%s: error allocating mmap for mcode alloc workaround", TAG);
+        goto quit;
+    }
+
+    // Resolve everything *now*, and put the symbols in the global scope, much like if we had linked it statically.
+    // This is necessary in order to be able to require Lua/C modules, c.f., LuaJIT docs on embedding.
+    // (Beware, Android's dynamic linker has a long history of weird and broken behavior,
+    // c.f., https://android.googlesource.com/platform/bionic/+/refs/heads/master/android-changes-for-ndk-developers.md)
```
Frenzie, line 133: lol, that's not even the half of it, you should take a peek at some of the 2010-era bionic commits

Frenzie (approved):
I'm not sure I fully grok this one in all the details but the explanation sounds simple enough.

jni/Android.mk
NiLuJe: That's mostly redundant anyway, I'm fairly sure the NDK will always link a BUILD_SHARED_LIB w/ `-ldl -llog -landroid`.

The -lm is arguably needed by LuaJIT, so it might help on braindead API levels. or it might also be part of the default NDK set of flags ;p.

NiLuJe: But to answer the question: it was necessary *before*, when LuaJIT was embedded statically, because the DT_NEEDED entries (mostly) ensures these were loaded in the global namespace (e.g., accessicle in ffi.C in Lua).

As the comment below show (and another comment in android.lua confirmed), this didn't quite work out on APIs where DT_NEEDED entries *aren't* put in the global namespace ^^.

NiLuJe:
As for why this appeared to randomly break when doing the most trivial things, it mostly revolves around the "lj_trace_flushall will unmap mcarea" aspect of things.

The JIT writes code into that area until it can't anymore (e.g., there's a couple of top & bottom pointers, with a red zone, and if code would go beyond the redzone => it requests more mcode area.
Requesting a single mcarea block already helped with this: it meant that those "more mcode area" requests didn't mean a trip to mmap.
But what happens when, besides reaching the end of an mcode area, you've alsoi reached the maxmcode limit?
Well, in this instance, LuaJIT ends up flushing all traces. And *that* implied unmapping all the mcode areas.
Which means new code needed new mmaps => kablooey.

As we'd kind of noticed, the Lua blitter is very mcode hungy. So, doing the most trivial things (say, switching tab in a menu) could mean hitting the redzone of the final mcode area => flushall.

That's what the "memset zero instead of unmap" patch helps with.

NiLuJe:
As for why requesting a new mmap is tricky: LuaJIT requires that the entire mcode area fit inside the architecture's jump range. That's +/- 32MB on ARM, which isn't much to begin with, and is made extremely tricky in the way Android lays out a process's virtual address space: it's *extremely* tightly packed, so finding even a 32K hole that fits in those constraints gets trickier and trickier as more stuff in mapped in the process.

Hence the "make a hole *inside* LuaJIT itself", and stay there ;).

NiLuJe:
Most (if not all) of those LuaJIT hacks are probably extremely *not* thread safe. But we're not threading Lua itself, so, we don't care.

Frenzie:
> But we're not threading Lua itself

Ho hum, pleasant for the future. :-P

NiLuJe:
Well, to be fair, threading wouldn't really be viable in the first place, because that'd be even more pressure in the same address space ;p.

Frenzie:
Wooo, Android! :partying_face:

NiLuJe:
You could arguably tweak the "make a hole" trick to be somewhat thread-aware, so that each thread makes a hole in a slightly *different* place, but that'd be fairly nasty ^^.

pazos (approved):
@NiLuJe: lgtm

A minor nit is the dummy c file needed to bundle libluajit.so. That could be avoided if we copy libluajit.so to a given path (libs/$(ANDROID_ABI)) and tell gradle to pick external libraries from there.

I think everything related to luajit can be managed from `mk-luajit.sh` and we can get rid of it in Android.mk.

NiLuJe:
@pazos: This iteration got rid of that ;). (I mean, the source is still there because I left the behavior optional for documentation purposes, but it's commented out and ifdef'ed out ;)).

NiLuJe:
But, yes, gradle would probably be the place to handle that properly. I just don't know how the hell this thing works ^^.

pazos:
> But, yes, gradle would probably be the place to handle that properly. I just don't know how the hell this thing works ^^.

Please merge this PR and I can take care of it next. I would merge it myself but I'm not sure if you want to keep the commit history or squash all commits together

NiLuJe:
Oh, definitely squash ;p.

Just have to double-check that the last commit didn't break everything ;p.

PR #287, opened 2021-01-05
Update LuaJIT mcode hack

NiLuJe:
* Fix pointer arithmetic on !x86/!x64

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/287)

### DIFF
--- a/jni/luajit/koreader-luajit-mcode-debug.patch
+++ b/jni/luajit/koreader-luajit-mcode-debug.patch
@@ -1,5 +1,5 @@
 diff --git a/src/lj_api.c b/src/lj_api.c
-index ecaf8a2..8af78dc 100644
+index 8b00436d..69540639 100644
 --- a/src/lj_api.c
 +++ b/src/lj_api.c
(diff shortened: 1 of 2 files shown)


NiLuJe:
This one shouldn't have been an issue on Android, but it could conceivably have been.

On the other hand, it was completely broken on plain Linux with the default settings ;).

pazos (approved):
I can pretend I understand what you coded but I won't pretend I'm going to review a diff of a patch :smile:

Frenzie (approved):
@pazos Well, I'm not sure I fully understand the issue but it doesn't look like it should break anything. :-D

NiLuJe:
All the pointers in question are `MCode *`, and... https://github.com/LuaJIT/LuaJIT/blob/1e66d0f9e6698fdee672c40a9a5b4159c9254091/src/lj_jit.h#L170-L175

So, on arm, it's an `uint32_t*`, which means instead of adding `size` bytes, we were adding `size * 4` bytes ;).

Issue #238, opened 2020-07-07
Multiwindow / free mode support.


pazos:
Mostly useful for ChromeOS or big tablets. Doesn't make much sense in a phone/e-ink device.

Android docs in https://developer.android.com/guide/topics/ui/multi-window.

Some notes:

1. Resize when the app has a native window to draw. That happens on ChromeOS when we're resizing but not on Android. On android resizing needs to happen in onStart() or when we have focus back, not onConfigurationChanged()
2. Disable fullscreen layout while in multiwindow mode. Also need to take constraints into account to calculate width and height (to avoid draw on top of the menu bar).
3. width and height in multiwindow mode are calculated based on reported data by window.decorView. Current implementations of android.getScreenWidth() and android.getScreenHeight will return bogus values.

I will label as won't fix because I won't but PR are indeed *very* welcome.

HoanPiTa:
I just wanna to read a book in 2 languages in a time.

pazos:
> I just wanna to read a book in 2 languages in a time.

I'm afraid you'll need another program. There're plenty of them in android and most should support multiple instances :)

HoanPiTa:
> > I just wanna to read a book in 2 languages in a time.
> 
> I'm afraid you'll need another program. There're plenty of them in android and most should support multiple instances :)
Yes. I am using Neo reader, moon reader for 2 book in a time. But i like koreader :((

beetlejuice16:
Hello, this is a really important feature for me and I would like to try and contribute with a PR. I sadly have no Android app development experience but I do have experience with Lua, would someone be willing to give me some basic pointers on which parts of the code need to be looked at so I can attempt to work my way through it?

Awesome project btw. Thanks for all your work!

PR #288, opened 2021-01-12
Fix & enable HW rotation on notched devices on P+

NiLuJe:
Looks simple enough (he said, hopefully ;p) :).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/288)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/MainActivity.kt
+++ b/app/src/main/java/org/koreader/launcher/MainActivity.kt
@@ -328,7 +328,7 @@ class MainActivity : NativeActivity(), JNILuaInterface,
 
     override fun getScreenHeight(): Int {
         return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-            // It handles the all insets for us, and is rotation-aware
+            // It handles all the insets for us, and is rotation-aware
             ScreenUtils.getScreenAvailableHeight(this)
         } else {
             ScreenUtils.getScreenHeight(this)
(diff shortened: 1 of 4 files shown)


NiLuJe:
Ah. Knew it was too easy: this excludes the status bar, too.

pazos (commented):
As you know `getScreenSizeWithConstraints(activity)` only works on pre KitKat devices (apis 17 & 18) and it is completely deprecated in immersive mode (apis19+).

Since legacy APIs doesn't use HW rotation I think it is safe to ignore `getScreenAvailableWidth` et al.

NiLuJe:
Note to self: nav bar, not status bar.

NiLuJe:
> As you know `getScreenSizeWithConstraints(activity)` only works on pre KitKat devices (apis 17 & 18) and it is completely deprecated in immersive mode (apis19+).
> 
> Since legacy APIs doesn't use HW rotation I think it is safe to ignore `getScreenAvailableWidth` et al.

Err, the only thing it does is using `getMetrics` instead of `getRealMetrics`. AFAICT, that's very much [not deprecated until SDK 30](https://developer.android.com/reference/android/view/Display#getMetrics(android.util.DisplayMetrics)), and it indeed worked perfectly fine on my API 29 phone. It just happens to *also* exclude the nav bar, which is a problem ^^.

Speaking of, the [SDK 30 variant](https://developer.android.com/reference/android/view/WindowMetrics#getBounds()) has a code excerpt to do that manually, and it appears to be able to tell display cutout apart from the nav bar ;).

pazos:
@NiLuJe: You're right. I was talking about our usage :P. As we always enable immersive mode in KitKat+ devices, we don't need `getMetrics` there. It could be relevant for https://github.com/koreader/android-luajit-launcher/issues/238, but there're better ways to retrieve screen size on [each surface change](https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/MainActivity.kt#L59-L62)

But having that available doesn't hurt either. :+1:

pazos (approved):

PR #293, opened 2021-02-02
add all keycodes

pazos:
As a prerrequisite for keyboard input.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/293)


Frenzie (approved):
It is? But surely the actual typed text must come as UTF-8 chars or some such? :-)

pazos:
See https://wladimir-tm4pda.github.io/porting/keymaps_keyboard_input.html

Apps receive keycodes, not scancodes. That's is handled by the window manager according with current IME settings.

What we receive is a bunch of keyDown and keyUp events. We ignore keyUp events for most keycodes, unless we want to set custom shortcuts using key modifiers. In that case we listen to Ctrl, Alt and such keys, and act differently when a keypress happens until their keyUp events.

Frenzie:
> Apps receive keycodes, not scancodes. 

Do apps or is that just what we're requesting? Cf. when I said to use startTextInput in SDL instead of doing exactly this https://github.com/koreader/koreader-base/commit/8614f997e487c440a9bb57252c20401f2aec54f3

Frenzie:
Why would we need to use NativeActivity? Most of the launcher thing is about passing things on from Android-y Java/Kotlin into a format we can work with in Lua, whether it's the clipboard (now quite old news) or a power connection event in https://github.com/koreader/android-luajit-launcher/pull/294.

Please do what you think is best, but I'm not sure if you're responding to the concept of what I'm saying. Because it comes down to being completely unable to type éëåéþfðgáßgßáfðhúgöøæ¶œï (that's just me mashing my keyboard while pressing Alt Gr on my particular layout in software).

I fully expect Android to be terrible, but that'd be beyond terrible.

pazos:
Because the two things a **NativeActivity** does is write to a buffer and handle input events. These two things cannot be done using the "normal" SDK.

That's why most activities reimplement their on NativeActivity on top of a normal Activity/AppCompatActivity: to be able to override `onKeyUp` and `onKeyDown` . See, for instance, the implementation in [ppsspp](https://github.com/hrydgard/native/blob/master/android/src/com/henrikrydgard/libnative/NativeActivity.java)

A link to documentation explaining how to handle key events from the JVM on apps like that would be awesome. But, AFAICT there isn't any, because is not possible :/

Frenzie:
A quick look at the docs brought me to https://developer.android.com/reference/android/view/inputmethod/BaseInputConnection and
https://developer.android.com/reference/android/view/inputmethod/InputConnectionWrapper

SDL does pretty much exactly what I was thinking (plus a bit more) `SDLInputConnection.nativeCommitText(text.toString(), newCursorPosition);`: https://github.com/libsdl-org/SDL/blob/c59d4dcd38c382a1e9b69b053756f1139a861574/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L2182-L2328

I'm sure I must be missing something obvious, but in any case SDL should make my train of thought abundantly clear. :-)

pazos:
Is a bit complicated on native activities. `KeyEvent.Callback` can be implemented on `MainActivity` and override its methods. These overrides will happen *after* [native event dispatching](https://github.com/koreader/koreader-base/blob/master/ffi/input_android.lua#L184-L198), so we need to ignore key events there while we're in *input mode*. We already [do that with certain keys](https://github.com/koreader/koreader-base/blob/master/ffi/input_android.lua#L114-L148). (don't push key events to frontend and return that they weren't handled, so they're forwarded to other parts of the activity/system services)

We still need to figure out how to pass the unicode char from `event.getUnicodeChar()` back to lua. I can sketch something that prints the result to logcat, but no idea how to implement the communication. Feel free to suggest options :)

Frenzie:
It doesn't seem that it should be much different than the clipboard?

PR #294, opened 2021-03-06
Power Connection Receiver 

zwim:
This adds a receiver for the power connection. -> Will be used for the upcoming power connections events.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/294)


app/src/main/java/org/koreader/launcher/PowerConnection.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import android.os.BatteryManager
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import org.koreader.launcher.interfaces.JNILuaInterface
+
+
+class PowerConnection : BroadcastReceiver() {
+    var isPowerConnected = 666
```
Frenzie, line 11: Wouldn't it make more sense to initialize it to 0?

zwim: The evil number just shows, that the var is not initialized - as we don't know the state here.
A call to `getBatteryState` will initialize the the value. 
Maybe it will be better to call `getBatteryState` here (instead of 666).

NiLuJe: `-1`?

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -268,6 +284,27 @@ class MainActivity : NativeActivity(), JNILuaInterface,
         return getBatteryState(true)
     }
 
+    override fun getBatteryState(isPercent: Boolean): Int {
```
pazos, line 287: Please don't!!

getBatteryState is a private function for a good reason. It is already exposed to lua in both variants: percent and isCharging. Leave it that way.

app/src/main/java/org/koreader/launcher/PowerConnection.kt
```diff
@@ -0,0 +1,42 @@
...
+class PowerConnection : BroadcastReceiver() {
+    var isPowerConnected = -1
+    var oldPowerConnection = -1
+
+	override fun onReceive(context: Context?, intent: Intent?) {
+        val action = intent?.action
+		when (action) {
+            Intent.ACTION_POWER_CONNECTED -> {
+				this.isPowerConnected = 1
+				// replace this by a better method to send a keyevent to self (so ALooper_pollAll wakes up)
```
pazos, line 19: THAT! Using `su` in a general purpose class is a big **nope**.

app/src/main/java/org/koreader/launcher/interfaces/JNILuaInterface.kt
```diff
@@ -6,13 +6,15 @@ package org.koreader.launcher.interfaces
...
     fun canWriteSystemSettings(): Boolean
+    fun powerConnectionEvent(): Boolean
     fun dictLookup(text: String?, action: String?, nullablePackage: String?)
     fun download(url: String, name: String): Int
     fun einkUpdate(mode: Int)
     fun einkUpdate(mode: Int, delay: Long, x: Int, y: Int, width: Int, height: Int)
     fun enableFrontlightSwitch(): Boolean
     fun extractAssets(): Boolean
     fun getBatteryLevel(): Int
+    fun getBatteryState(isPercent: Boolean): Int
```
pazos, line 17: ^^

Already exposed as `getBatteryLevel` and `isCharging`

assets/android.lua
```diff
@@ -1938,6 +1938,17 @@ local function run(android_app_state)
         end)
     end
 
+    android.powerConnectionEvent = function()
```
pazos, line 1941: I don't understand why you need that. Isn't that polling? How is this different that calling `isCharging`?

pazos (changes requested):
If you manage to gen a fake key event using public SDK/NDK methods then I think it should be ok. In that case:

1. Get rid of the PowerConnection class and make it an inner class of MainActivity (it's always called from there and will be the same in all devices, there's no need for an extra file)
2. Get rid of the extra android functions since you won't need them. The key event will be delivered to the frontend and can be handled there
3. On the same vein, since you're registering the broadcast receiver in onCreate you should keep the var `isPowerConnected` or `lastPowerConnectedState` as a MainActivity class var. The same for the function `connectionEvent` since now it compares values that are available on MainActivity.

app/src/main/java/org/koreader/launcher/MainActivity.kt
zwim: With this PR `getBatteryState()` is only called from `getBatteryLevel()`.
`isCharging()` uses the stored `isPowerConnected` property, which is way faster than `getBatteryLevel()`.

 So no problem in keeping that private.

app/src/main/java/org/koreader/launcher/PowerConnection.kt
zwim: I totally agree that this is the point to solve in this draft PR. That's why the comment is there :))

zwim:
@pazos : My idea would be:
1.) Create a fifo/pipe.
2.) When MainActivity/PowerConnection/onReceive receives a PowerConnection-event then the new state (on/off) is written to the fifo.
3.) In android-main.c append a `ALooper_addfd` on that fifo,  so that `ALooper_pollAll` finishes.

Then the new connection state could be used in the LUA-space after ALooper_pollAll.

What do you think about this approach?

NiLuJe:
I don't know how this affects Android, but be aware that polling a RW pipe has some *very* funky corner-cases: https://stackoverflow.com/a/17384067

(I had to deal with the weirdness in [FBInk](https://github.com/NiLuJe/FBInk/blob/e685897a1bb4e37903e4896e518368d1f220846d/fbink_cmd.c#L2064-L2069)).

zwim:
Now we have a FIFO (`alooper.fifo`) for communication from Kotlin to Lua. The data transferred is rudimentary for now. Namely this is just a flag, that something notable happened on Android. What this was has to be checked from Lua. Up to now, only one event (Charging/NotCharging) is signalled with this, so there is no "dispatcher" to detect anything else :))

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -22,6 +22,11 @@ import org.koreader.launcher.interfaces.JNILuaInterface
 import org.koreader.launcher.utils.*
 import java.util.*
 
+import android.content.BroadcastReceiver
+import java.io.*
+
+val ALOOPER_FIFO = "/data/data/org.koreader.launcher/files/alooper.fifo"
```
NiLuJe, line 28: This probably won't work on newer APIs with weird-ass storage containers, will it?

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -215,16 +236,32 @@ static void* android_app_entry(void* param) {
...
+    ALooper_addFd(native_glue_looper, android_app->msgread, LOOPER_ID_MAIN, ALOOPER_EVENT_INPUT, NULL,
             &android_app->cmdPollSource);
-    android_app->looper = looper;
 
+    if (mkfifo(FIFO_PATH, 0666) == -1) {
+        if (errno == EEXIST)
+            LOGD("%s: FIFO \"%s\" already exists", TAG, FIFO_PATH);
+        else
+            LOGE("%s: FIFO \"%s\" cannot be created!", TAG, FIFO_PATH);
+    } else
```
NiLuJe, line 248: Braces or no braces (I don't know the house style for the rest of the file), but only half or the braces is super-weird ;).

NiLuJe: (I'm personally rather fond of enforcing braces, always, everywhere, even for single-lines. But that's me.)

app/src/main/java/org/koreader/launcher/MainActivity.kt
zwim: `getFilesDir()` is used from MainActivity.kt and `activity->internalDataPath` on the native_app_glue.c

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -238,17 +238,24 @@ static void* android_app_entry(void* param) {
     ALooper_addFd(native_glue_looper, android_app->msgread, LOOPER_ID_MAIN, ALOOPER_EVENT_INPUT, NULL,
             &android_app->cmdPollSource);
 
-    if (mkfifo(FIFO_PATH, 0666) == -1) {
+
+    const char* fifo_path = android_app->activity->internalDataPath;
+    char fifo_file[strlen(fifo_path) + strlen(FIFO_NAME) + 2]; // +1 for "/" and +1 for NULL
+    strcpy(fifo_file, fifo_path);
+    strcat(fifo_file, "/");
+    strcat(fifo_file, FIFO_NAME);
```
NiLuJe, line 246:
```suggestion
char fifo_file[strlen(fifo_path) + strlen(FIFO_NAME) + 2] = { 0 };
snprintf(fifo_file, sizeof(fifo_file) - 1U, "%s/%s", fifo_path, FIFO_NAME);
```

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -200,6 +206,18 @@ static void process_cmd(struct android_app* app, struct android_poll_source* sou
     android_app_post_exec_cmd(app, cmd);
 }
 
+ALooper *native_glue_looper;
+
+int fifoCallback(int fd, int events, void *data)
+{
+    char c;
+    int result = read(fd, &c, sizeof(c));
```
NiLuJe, line 214:
```suggestion
    ssize_t result = read(fd, &c, sizeof(c));
```

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -215,18 +233,48 @@ static void* android_app_entry(void* param) {
...
+        if (errno == EEXIST) {
+            LOGD("%s: FIFO \"%s\" already exists", TAG, fifo_file);
+        } else {
+            LOGE("%s: FIFO \"%s\" cannot be created!", TAG, fifo_file);
+        }
+    } else {
+        LOGD("%s: FIFO \"%s\" created", TAG, fifo_file);
+    }
+
+    int fifo_fd = open(fifo_file, O_RDWR | O_CLOEXEC);
```
NiLuJe, line 258: Shouldn't this be `O_NONBLOCK`, too?

jni/android_native_app_glue/android_native_app_glue.c
zwim: `{ 0 }` Initializing cannot be done with variable sized arrays at least in C.

NiLuJe: Whoops, indeed those are strlen calls, not sizeof calls ;p.

Uglier option, then:
```C
char fifo_file[strlen(fifo_path) + strlen(FIFO_NAME) + 2U];
snprintf(fifo_file, sizeof(fifo_file) - 1U, "%s/%s", fifo_path, FIFO_NAME);
fifo_file[sizeof(fifo_file) - 1U] = '\0';
```

NiLuJe: (Or just use `PATH_MAX`).

zwim: Shouldn't that be `sizeof(fifo_file)` without the `-1U`?

jni/android_native_app_glue/android_native_app_glue.c
zwim: I have tested this on a phone and a Tolino. 
With `O_RDWR`  `O_NONBLOCK` is not necessary.
Only with `O_RDONLY` we would need nonblock.

jni/android_native_app_glue/android_native_app_glue.c
NiLuJe: The last slot of an n-bytes array is at n-1 ;).

(`sizeof` does just what it says on the tin. You were perhaps thinking of `strlen`, which doesn't count NUL bytes. But since the goal here is to ensure NUL termination, we can't rely on strlen ;)).

NiLuJe: Yep, you were talking about the `snprintf` ;).

In this specific instance it's harmless, but it'll made static analyzers scream at you nonetheless ;).

jni/android_native_app_glue/android_native_app_glue.c
NiLuJe: I'm not quite sure what the R/W status would change, but since it's only used in a polling loop, it just makes more sense to me as NONBLOCK (again, mostly harmless in practice, until the day something possibly goes wrong ^^).

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -215,18 +234,45 @@ static void* android_app_entry(void* param) {
     android_app->inputPollSource.app = android_app;
     android_app->inputPollSource.process = process_input;
 
-    ALooper* looper = ALooper_prepare(ALOOPER_PREPARE_ALLOW_NON_CALLBACKS);
-    ALooper_addFd(looper, android_app->msgread, LOOPER_ID_MAIN, ALOOPER_EVENT_INPUT, NULL,
+    native_glue_looper = ALooper_prepare(ALOOPER_PREPARE_ALLOW_NON_CALLBACKS);
```
pazos, line 237: Please don't change names without a good reason. It makes harder to understand the changes. `looper` is a good name for an ALooper instance. There's no need to change it to `native_glue_looper`

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -22,6 +22,9 @@ import org.koreader.launcher.interfaces.JNILuaInterface
 import org.koreader.launcher.utils.*
 import java.util.*
 
+import android.content.BroadcastReceiver
```
pazos, line 25: please sort the imports, like the rest of the file

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -200,6 +207,18 @@ static void process_cmd(struct android_app* app, struct android_poll_source* sou
     android_app_post_exec_cmd(app, cmd);
 }
 
+ALooper *native_glue_looper;
+
+int fifoCallback(int fd, int events, void *data)
```
pazos, line 212: Please add a multiline comment on this one, because the rest of the file is mostly vanilla from Android NDK sources. It can be useful if we need to backport newer changes to our tree (last update on android native glue was made with NDKr22)

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -29,6 +33,9 @@
 
 #define  TAG "[NativeGlue]"
 
+// ToDO: How can I retrieve the path to the cache directory from here (native_app_glue)?
```
pazos, line 36: old comment?

jni/android_native_app_glue/android_native_app_glue.c
zwim: I think it is like the following:
string1="xyz", string2="abcd"; strlen1=3 strlen2=4.
If we concat "xyz/abcd" we would need an array with 3+1+4+1 =9bytes (strlen1+strlen2+2).
When calling `strnprintf` we give the size 9, as the man page say, that size bytes are written (including the terminating NUL).
`fifo_file[sizeof(fifo_file) - 1U] = '\0';` is not necessary then, but does no harm.
Am I wrong?

jni/android_native_app_glue/android_native_app_glue.c
zwim: We can leave the name `looper`, but then we need an additional global variable `native_glue_looper=looper` in the `ALooper_addFd` to wake the thing.

app/src/main/java/org/koreader/launcher/MainActivity.kt
zwim: OK

jni/android_native_app_glue/android_native_app_glue.c
zwim: OK

jni/android_native_app_glue/android_native_app_glue.c
zwim: Yepp.

jni/android_native_app_glue/android_native_app_glue.c
pazos: yup, new stuff is ok. For existing stuff please keep the names. The files are mostly vanilla from https://android.googlesource.com/platform/development/+/4948c163663ecc343c97e4c2a2139234f1d3273f/ndk/sources/android/native_app_glue

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -67,6 +70,45 @@ class MainActivity : NativeActivity(), JNILuaInterface,
         }
     }
 
+    inner class PowerConnection : BroadcastReceiver() {
+        var isPowerConnected = -1
+        private lateinit var fifo_name: String
+        private lateinit var alooper_fifo: FileWriter
+
+        fun feed_alooper_fifo(state: Int) {
```
pazos, line 78: very interesting approach :+1: 

IIRC this can be used to execute arbitrary lua chunks based on android events, isn't it?

In that case I would propose to move the code outside the broadcast receiver class.  For instance we keep some values that are the same for the entire life of the app in https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/MainApp.kt and just use them with `MainApp.whatever`. You can put there the path to `"alooper.fifo" and just have a private function in MainActivity that post messages to it. Not sure if you can do the initialization in MainApp too. If that's possible it would be amazing

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -67,6 +70,45 @@ class MainActivity : NativeActivity(), JNILuaInterface,
...
+                 Logger.e("BroadcastReceiver: ERROR writing to  ALooper fifo: \"$fifo_name\"")
+                 Logger.e("$e")
+             }
+        }
+
+        override fun onReceive(context: Context?, intent: Intent?) {
+            val action = intent?.action
+            when (action) {
+                Intent.ACTION_POWER_CONNECTED -> {
+                    isPowerConnected = 1
```
pazos, line 101: since we can feed other "states" through the fifo it would be nice to have a couple of constants instead of raw numbers. See https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/Device.kt#L60-L77, for example. You can name them as you wish, and being very verbose is suggested. (Something like `ALOOPER_FIFO_MSG_CONNECTED` is fine)

assets/android.lua
```diff
@@ -110,6 +110,12 @@ typedef int (*ALooper_callbackFunc)(int fd, int events, void* data);
 int ALooper_pollAll(int timeoutMillis, int* outFd, int* outEvents, void** outData);
 int ALooper_pollOnce(int timeoutMillis, int* outFd, int* outEvents, void** outData);
 
+// for alooper.fifo event
```
pazos, line 113: Is this still relevant?

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -413,7 +465,7 @@ class MainActivity : NativeActivity(), JNILuaInterface,
     }
 
     override fun isCharging(): Boolean {
-        return (getBatteryState(false) == 1)
+        return powerConnection.isPowerConnected != 0
```
pazos, line 468: This one was fine as is. Keep in mind there're some pretty borked android devices out there. A function that returns if the device is charging should actually check if it is changing right now and not returing the last state registered.

The medium overhead per JNI call is at most a ms. I think we can afford it :)

jni/android_native_app_glue/android_native_app_glue.c
NiLuJe: Yup, exactly ;).

assets/android.lua
zwim: no.

app/src/main/java/org/koreader/launcher/MainActivity.kt
zwim: Ok.

app/src/main/java/org/koreader/launcher/MainActivity.kt
zwim: I am afraid, it's not that easy. The only thing I can feed through the fifo is `ALOOPER_POLL_WAKE` as result of the `ALooper_addFD` callback.
If I delete the `ALooper_wake` from the callback and try to put some values to the `*data` pointer, I don't get any data in lua. Well not quite right. I get the data continuously flooding after a 40 secs or so and KOReader freezes.
With the `ALooper_wake` I get an immediate result.

So we could use the fifo for signalling that there is something and then maybe ask some data from java-land, what the "event" was but not the `android_poll_source` struct.

zwim:
In the next time (until Easter) I won't have time to work on that. So if someone would like to overtake the work, fell free to do so.

app/src/main/java/org/koreader/launcher/MainActivity.kt
pazos: maybe store the code that caused the wake in a var and use a new function to check that var in base (instead of asumming is a power event)?

We already do something like that to check the path of the file returned from SAF: [lastImportedPath](https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/MainActivity.kt#L34-L35). It is nullable and, when not null, provides a string with a path to a file. The function `getLastImportedPath` returns its value and resets it to null.

Maybe you can do the same with an `Int?`, that you feed with a non null value in the broadcast receiver. In the poll loop when Alooper wakes you check its code and act accordly

What do you think? :)

It would be a pitty to have such a nice feature constrained to a very minor/specific event :)

zwim: > What do you think? :)

That would have been my idea: First we need to define the data-structure which is written to the fifo (e.g. 2x16Bit: 65356keys with 65356values each; this has to be decided by YOU before we start implementing it). The Alooper_addFd callback reads these four bytes and stores them. When an ALOOPER_POLL_WAKE occurs, we read the stored data with a function in base (android-input.lua) and create a misc event (which we can handle with handeMiscEv in android/device.lua throwing some events) or do whatever we want directly in base (call some base function ...).

Stumbling blocks are: reading from the fifo:
1: I didn't get O_NONBLOCK to work reliably (maybe my fault). Have to dig into that (no time now :(  ).
2 : With more than one byte of data we have to synchronize the communication. Otherwise we might get freezes if a byte gets lost. -> Could be done by not allowing a zero byte in the data (zero-byte==wrong data) -> needs quite some test (and time :( ).
3: Two or more senders writing at the same time to the fifo (I know, fifo atomic blablabla), has to be tested before the feature is getting cool. (Maybe some sort of checksum will be needed?)
   
Benefits: Every part of KOReader (including the java and native parts) can send a huge number of different messages to lua-land. In lua-land for each message a helper can be called to retrieve the associated data.

> 
> It would be a pitty to have such a nice feature constrained to a very minor/specific event :)

Yes, the idea is expandable and should be pursued further, but I just don't have the time to do (and test it) now.
In the Easter holidays my time budget is better.

pazos:
> In the next time (until Easter) I won't have time to work on that. So if someone would like to overtake the work, fell free to do so.

We can wait a few weeks without doubt :) Very good job up to now

zwim:
> We can wait a few weeks without doubt :) Very good job up to now

I had good and tolerant teachers :))

app/src/main/java/org/koreader/launcher/MainActivity.kt
NiLuJe: About 3.: Yeah, on Linux, a FIFO should ensure atomicitiy, boundaries and order are respected up to PIPE_BUF (4K).

NiLuJe: And about 1.: What kind of issues did you encounter, in practice?

(Re: https://github.com/koreader/android-luajit-launcher/pull/294#issuecomment-793013837 ;)).

zwim: good, so 2+3 wont need so mcch tests.

zwim: To 1: I got the same value over and over again, flooding. -> This means, that the addFd callback is called continuously, isnt it.

pazos: We don't need fancy stuff passed using the fifo. Just use the fifo to tell the main thread (where LuaJIT runs and controls the loop) that there's a pending message and get that message using the new call. Something like this in base should work:

```lua
elseif poll_state == C.ALOOPER_POLL_WAKE then
    local event = android.getEvent()
    if event and event == C.EVENT_POWER_CONNECTION then
        -- stuff here to check that usercase
    end
end
```

the getEvent function (use a name you like) should return nil if the wake event is unexpected or a number described in a enum in `assets/android.lua`

You can also repurpose the event name to tell to the frontend. So, instead of `commandHandler(C.MSC_CHARGE, 1)` you'll use `commandHandler(C.EVENT_POWER_CONNECTION , 1)`

Again, names are mine and can be changed. But the general idea is:

1. Events that need to execute lua chunks use a general purpose function that stores an integer in a var somewhere in kotlin and send the same bit of info using the fifo to trigger Alooper_wake
2. The poll loop then checks that specific value (once per Alooper_wake) and send the message to the frontend accordly.

NiLuJe: > To 1: I got the same value over and over again, flooding. -> This means, that the addFd callback is called continuously, isnt it.

That smells like one the many FIFO corner-cases, yeah (e.g., POLLIN never cleared or POLLHUP permanently set or something). I'd need to know how the Android stuff wraps `poll` to tell what's really happening, though ;).

Then again, if the actual `read` call returned content every time, maybe more complex than that... :/

zwim:
The following commits seem to work. But due to my restricted time resources they are not tested thoroughly.  I will do that in a few days.

app/src/main/java/org/koreader/launcher/MainApp.kt
```diff
@@ -17,8 +19,53 @@ class MainApp : android.app.Application() {
...
+        private lateinit var alooper_fifo: FileWriter
+
+        // feeds an 32 bit value to the fifo
+        // low byte first
+        fun feed_alooper_fifo(state: Int) {
+            if (!this::app.isInitialized) {
+                -1
+            }
+
+            if (!this::context.isInitialized) {
```
pazos, line 35: Not sure why are you using kotlin constructs for these kind of lateinit vars.

They aren't needed at all. If you need a context and can't get one here just move the function to MainActivity. In the same vein the FileWriter should be local in the function. Two or more broadcast receivers can write to the file "at the same time"

But, for the specific usage of the context you shouldn't have problems. (both the app and activity classes  inherit from a context)

See **https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/MainApp.kt#L26**

you don't need to use `context.getFilesDir()` Whenever possible prefer the more idiomatic `filesDir`

Assuming the same name for an example, but I think `postMessage` makes more sense than `feed_alooper_fifo`. The former shows intention and the later is a bit too techical and exposes implementation details

```kotlin
// Function in MainApp or MainActivity
fun feed_alooper_fifo(state: Int) {
    try {
        val writer = FileWriter(MainApp.fifo_path, true)
        val fifo_bytes = CharArray(4)
        fifo_bytes[0] = (state and 0xFF).toChar()
        fifo_bytes[1] = ((state ushr 8) and 0xFF).toChar()
        fifo_bytes[2] = ((state ushr 16) and 0xFF).toChar()
        fifo_bytes[3] = ((state ushr 24) and 0xFF).toChar()
        writer_write(fifo_bytes, 0, 4)
        writer.flush()
    } catch (e: Exception) {
        // whatever error msg
    }
}
```
}

pazos: The bit of info I suggested for `MainApp` was

```diff
diff --git a/app/src/main/java/org/koreader/launcher/MainApp.kt b/app/src/main/java/org/koreader/launcher/MainApp.kt
index 823434d..7a65255 100644
--- a/app/src/main/java/org/koreader/launcher/MainApp.kt
+++ b/app/src/main/java/org/koreader/launcher/MainApp.kt
@@ -7,12 +7,16 @@ import android.content.pm.PackageManager
 import android.os.Environment
 import android.os.StrictMode

+import java.io.File
+
 class MainApp : android.app.Application() {
     companion object {
         private val runtime = android.os.Build.VERSION.SDK_INT

         lateinit var info: String
             private set
+        lateinit var fifo_path: String
+            private set
         lateinit var storage_path: String
             private set
         lateinit var platform_type: String
@@ -40,6 +44,10 @@ class MainApp : android.app.Application() {

         @Suppress("DEPRECATION")
         storage_path = Environment.getExternalStorageDirectory().absolutePath
+
+        /* */
+
+        fifo_path = File(filesDir, "alooper.fifo").path
         platform_type = if (pm.hasSystemFeature("org.chromium.arc.device_management")) {
             "chrome"
         } else if ((runtime >= 21) && pm.hasSystemFeature(PackageManager.FEATURE_LEANBACK)) {
```

Sorry for the confussion :)

zwim: Thanks for the help. The code is much cleaner now.

zwim: But it has to be tested yet.

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -67,6 +69,20 @@ class MainActivity : NativeActivity(), JNILuaInterface,
         }
     }
 
+    inner class PowerConnection : BroadcastReceiver() {
```
pazos, line 72: We can use the same receiver to get all kind of system events. Most of them useless, of couse. The entire list as of API30 is https://developer.android.com/about/versions/11/reference/broadcast-intents-30

So I would suggest a more generic `Receiver` as the name of the class and `receiver` as the name of the object. Maybe `EventReceiver` and `event` ?

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -67,6 +69,20 @@ class MainActivity : NativeActivity(), JNILuaInterface,
...
+        override fun onReceive(context: Context?, intent: Intent?) {
+            val action = intent?.action
+            when (action) {
+                Intent.ACTION_POWER_CONNECTED -> {
+                    MainApp.postMessages(MainApp.ALOOPER_FIFO_POWER_CONNECTED)
+                }
+                Intent.ACTION_POWER_DISCONNECTED -> {
+                    MainApp.postMessages(MainApp.ALOOPER_FIFO_POWER_DISCONNECTED)
+                }
```
pazos, line 81: Kotlin needs an `else` branch when you use `when`.
Docs are https://kotlinlang.org/docs/control-flow.html#when-expression and seems a good habit to include an else branch even if is not strictly neccesary.

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -91,6 +107,15 @@ class MainActivity : NativeActivity(), JNILuaInterface,
         clipboard = Clipboard(this)
         device = Device(this)
         timeout = Timeout()
+
+        powerConnection = PowerConnection()
```
pazos, line 111: ^^

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -217,6 +242,7 @@ class MainActivity : NativeActivity(), JNILuaInterface,
     override fun canIgnoreBatteryOptimizations(): Boolean {
         return Permissions.isIgnoringBatteryOptimizations(this)
     }
+
```
pazos, line 245: unrelated change

app/src/main/java/org/koreader/launcher/MainApp.kt
```diff
@@ -6,17 +6,45 @@ import android.content.pm.ApplicationInfo
...
 
         lateinit var info: String
             private set
+        lateinit var fifo_path: String
+            private set
         lateinit var storage_path: String
             private set
         lateinit var platform_type: String
             private set
+
```
pazos, line 24: I'm not sure where to put this chunk of code but very much not here :)

The rest of constants are placed in a companion object within the class. See, for example, https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/Device.kt#L60 or https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/MainActivity.kt#L70

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -215,18 +244,49 @@ static void* android_app_entry(void* param) {
     android_app->inputPollSource.app = android_app;
     android_app->inputPollSource.process = process_input;
 
-    ALooper* looper = ALooper_prepare(ALOOPER_PREPARE_ALLOW_NON_CALLBACKS);
+    ALooper *looper = ALooper_prepare(ALOOPER_PREPARE_ALLOW_NON_CALLBACKS);
+    native_glue_looper = looper;
```
pazos, line 248: ?

Please use `looper` everywhere.

zwim: I need the `native_glue_looper` as a global variable for the `ALooper_wake` in the callback.

zwim:
From my side this PR would be ready. I have tested it on two Tolinos and my mobile.

app/src/main/java/org/koreader/launcher/MainApp.kt
```diff
@@ -6,17 +6,39 @@ import android.content.pm.ApplicationInfo
...
+        // low byte first
+        fun postMessages(state: Int) {
+            try {
+                val writer = FileWriter(MainApp.fifo_path, true)
+                val fifo_bytes = CharArray(4)
+                fifo_bytes[0] = (state and 0xFF).toChar()
+                fifo_bytes[1] = ((state ushr 8) and 0xFF).toChar()
+                fifo_bytes[2] = ((state ushr 16) and 0xFF).toChar()
+                fifo_bytes[3] = ((state ushr 24) and 0xFF).toChar()
+                writer.write(fifo_bytes, 0, 4)
```
NiLuJe, line 35: Stupid question: why the gymnastics here? this is not crossing a network boundary, what's wrong with writing an int on this side and reading an int on the other?

app/src/main/java/org/koreader/launcher/Device.kt
```diff
@@ -75,6 +75,12 @@ class Device(activity: Activity) {
         private const val LINUX_LANDSCAPE = 1
         private const val LINUX_REVERSE_PORTRAIT = 2
         private const val LINUX_REVERSE_LANDSCAPE = 3
+
+        // These are the messages going over the fifo to the native glue.
```
pazos, line 79: Device is a good place (as MainActivity). If you go this route then make the constants private and use them in the same file. Both the broadcast receiver and the postMessages function should be placed here. (You can instantiate the broadcast receiver and the filter on the class constructor `init`.

app/src/main/java/org/koreader/launcher/Device.kt
```diff
@@ -75,6 +75,12 @@ class Device(activity: Activity) {
...
+        // The defines here should correspond to the ones in android.lua
```
pazos, line 80: The comment is generic for the rest of the constants here. All of them need to be placed in `android.lua`

app/src/main/java/org/koreader/launcher/Device.kt
```diff
@@ -75,6 +75,12 @@ class Device(activity: Activity) {
...
+        // 32bit (4-byte) get transmitted. The low byte is the command
+        val ALOOPER_FIFO_POWER_CONNECTED = 100
```
pazos, line 82: A sugestion that hides implementation details is 

```kotlin
private const val EVENT_POWER_CONNECTED = 100
```

app/src/main/java/org/koreader/launcher/MainApp.kt
zwim: I use this, because the fileWriter doesn't have a method to write a 32bit int.

jni/android_native_app_glue/android_native_app_glue.c
zwim: Am I to stupid? 
The callback is called directly from `ALooper_pollAll()` and the parameters are (int filedescriptor, int events, void *data). How can I pass an additional parameter to the callback?
Or do you mean to fill the `message_to_lua` buffer with a reference to `looper` and then use a static variable within the callback to hold the reference for consecutive calls? That wouldn't be much clearer, wouldn't it?

app/src/main/java/org/koreader/launcher/Device.kt
```diff
@@ -75,13 +82,60 @@ class Device(activity: Activity) {
...
 
         Logger.v(TAG, String.format(Locale.US,
             "native orientation: %s", if (this.screenIsLandscape) "landscape" else "portrait"))
+
+        event = EventReceiver()
+
+        val filter = IntentFilter()
+        filter.addAction(Intent.ACTION_POWER_CONNECTED)
+        filter.addAction(Intent.ACTION_POWER_DISCONNECTED)
+        activity.getApplicationContext().registerReceiver(event, filter)
```
pazos, line 120: nit: whenever possible use kotlin getters. In this case `activity.applicationContext` instead of `activity.getApplicationContext()`

app/src/main/java/org/koreader/launcher/Device.kt
```diff
@@ -75,13 +82,60 @@ class Device(activity: Activity) {
...
+    }
+
+    // post an 32 bit message which can be evaluated after an ALooper_pollAll()
+    // low byte first
+    fun postMessages(state: Int) {
```
pazos, line 125: nit: private

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -200,6 +204,29 @@ static void process_cmd(struct android_app* app, struct android_poll_source* sou
...
 
+/* We use a fifo to communicate from Java-land to luaspace.
+ * fifoCallback is called, when writen a message to that fifo.
+ * native_glue_looper is needed to wake that looper
+ */
+ALooper *native_glue_looper;
+#define MESSAGE_SIZE 4
+u_int8_t message_to_lua[MESSAGE_SIZE];  //maximum 4 byte messages
+
+int fifoCallback(int fd, int events, void *data)
```
pazos, line 215: `int fifoCallback(int fd, int events, void *data, ALooper &looper)` should work the same and avoids to modify the file more than necessary.

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -219,14 +246,43 @@ static void* android_app_entry(void* param) {
     ALooper_addFd(looper, android_app->msgread, LOOPER_ID_MAIN, ALOOPER_EVENT_INPUT, NULL,
             &android_app->cmdPollSource);
     android_app->looper = looper;
-
```
pazos, line 222: nit: unrelated change

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -219,14 +246,43 @@ static void* android_app_entry(void* param) {
...
+        native_glue_looper = looper;
+        ALooper_addFd(looper, fifo_fd, 0, ALOOPER_EVENT_INPUT, fifoCallback, message_to_lua);
+    }
+
     android_main(android_app);
 
+    // clean up fifo
+    if (fifo_fd != -1) {
+        ALooper_removeFd(native_glue_looper, fifo_fd);
+        close(fifo_fd);
```
pazos, line 283: question: Do you want to remove the file here or shall we keep the file created from one execution to the next one?

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -219,14 +246,43 @@ static void* android_app_entry(void* param) {
...
     pthread_mutex_unlock(&android_app->mutex);
 
+    // create and open fifo for communication with MainActivity
+    const char* fifo_path = android_app->activity->internalDataPath;
+    char fifo_file[strlen(fifo_path) + strlen(FIFO_NAME) + 2U]; // +1 for "/" and +1 for NULL
+    snprintf(fifo_file, sizeof(fifo_file), "%s/%s", fifo_path, FIFO_NAME);
+
+    if (mkfifo(fifo_file, 0666) == -1) {
+        if (errno == EEXIST) {
+            LOGD("%s: FIFO \"%s\" already exists", TAG, fifo_file);
```
pazos, line 261: Actually we can use `LOGW`  or `LOGV` macros if they make more sense.  For example I wouldn't mind to have these logged on release builds as verbose messages. The macro is already used elsewhere, see https://github.com/koreader/android-luajit-launcher/blob/master/jni/android-main.c#L42

jni/android_native_app_glue/android_native_app_glue.c
zwim: I would prefer to leave the file. Less disk accesses are better.

jni/android_native_app_glue/android_native_app_glue.c
zwim: OK

jni/android_native_app_glue/android_native_app_glue.c
zwim: OK

jni/android_native_app_glue/android_native_app_glue.c
pazos: I didn't know.

You already have a pointer to a ALooper. Why don't use it directly? Can't you use its memory address with `ALooper_wake(&looper)`?

zwim: Hmmm. The callback gets called from ALooper_pollAll. The bad thing is, that the callback only gets those three parameters (and nothing more, we are on C here, aren't we?). So the callback has no clue about the looper structure/address or whatever. 
So my clumsy/smart (?) solution is the global variable `native_glue_looper`.

Sorry can't test more now, as I am fighting with git. I have done an update of `base` (wanted to try the awesome mathml things in crengine) and now I get the error:
`cre.cpp:504:20: error: 'PROP_FALLBACK_FONT_FACE' was not declared in this scope`

NiLuJe: (try a clean rebuild (e.g., make dist-clean).

zwim: I have allrdy done a kodev clean android and after that a kodev build android. Did not help.

Thank you.

NiLuJe: Okay, read the context, and I'd tend to agree with @pazos.

One way i can think of to handle that: Use a struc,t allocate it on the stack, pass a pointer to *that* to the callback, and free it when the message has been consumed (in the callback?).

```
typedef struct {
    ALooper * looper;
    unsigned char msg[MESSAGE_SIZE];
} FifoCtx;

...

FifoCtx msg = calloc(1, sizeof(FifoCtx);
msg.looper = looper;

...

FifoCtx* msg = data;

...

```

Sidebar: What's the android_app->userData thing for?

NiLuJe: So, new stupid question: can't we just read the fd in base?

This gets rid of *all* this nonsense, since we basically just have to write to the fifo and call it a day.

Base just needs to read the fd when the ident is `fifo_id` (and no need for the extra wake either, the fact that data has been written means pollAll will detect it).

zwim: yessssss. 
And therfore, to make the things easier in base I would like to turn the thing round in the struct. First the data and then the pointer. 
It should be freed on the end of the progam as we need the struct, the data and the pointer on every event!

NiLuJe: > I have allrdy done a kodev clean android and after that a kodev build android. Did not help.
> 
> Thank you.

`make dist-clean` is not `kodev clean`, it's better ;p.

NiLuJe: > yessssss.
> And therfore, to make the things easier in base I would like to turn the thing round in the struct. First the data and then the pointer.
> It should be freed on the end of the progam as we need the struct, the data and the pointer on every event!

c.f., my following answer, which should make *everything* even much simpler, if polling behaves sanely.

NiLuJe: In practice: Make the FIFO's addFd ident use a specific constant that we can retrieve from base, instead of the actual fd.

In base:

Actually store the outFd param:

```
 local outFd = ffi.new("int*[1]")
 local poll_state = android.lib.ALooper_pollAll(timeout, outFd, events, ffi.cast("void**", source))

 if poll_state >= 0 and poll_state == fifo_id, open & read on outFd[0]
```

zwim:
I don't think this works. Alooper is sitting there and does not end. The callback wakes it up. And only after that wake we can read the data. 
If we do that in the android part or the base part or the frontend part is no real gamechanger.
Logically it is an Android event so it should be decoded there?

Am 22. März 2021 23:06:01 MEZ schrieb NiLuJe ***@***.***>:
>@NiLuJe commented on this pull request.
>
>
>
>> @@ -215,18 +244,49 @@ static void* android_app_entry(void* param) {
>     android_app->inputPollSource.app = android_app;
>     android_app->inputPollSource.process = process_input;
> 
>-    ALooper* looper =
>ALooper_prepare(ALOOPER_PREPARE_ALLOW_NON_CALLBACKS);
>+    ALooper *looper =
>ALooper_prepare(ALOOPER_PREPARE_ALLOW_NON_CALLBACKS);
>+    native_glue_looper = looper;
>
>So, new stupid question: can't we just read the fd in base?
>
>This gets rid of *all* this nonsense, since we basically just have to
>write to the fifo and call it a day.
>
>Base just needs to read the fd when the ident is `fifo_id` (and no need
>for the extra wake either, the fact that data has been written means
>pollAll will detect it).
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/koreader/android-luajit-launcher/pull/294#discussion_r599106543

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

NiLuJe:
But it should, this a wrapper around `poll`, the docs say:

```
ALOOPER_EVENT_INPUT | The file descriptor is available for read operations.
```

That's poll's `POLLIN`.

If you write to the FIFO, it'll be flagged POLLIN, that'll "wake" the poller.

jni/android_native_app_glue/android_native_app_glue.c
zwim: Ah that's a new quality :))) Very nice.
I will work on it the next days.

NiLuJe:
This might make what I was talking about earlier:

https://github.com/koreader/koreader-base/pull/1332/commits/7084a36a41e9042e5c55b30e626a4556bbd166e0

(Which means, yeah, the whole data pointer thing is completely bonkers and redundant, and we don't actually need it).

NiLuJe:
What looking at this revealed is that the CMD handling already functions in a similar manner, except it's using a pipe, not a named pipe.

Which makes sense: you don't actually need to access this outside of the process, so a simple anonymous pipe might make all the I/O and FS-related mess much clearer (c.f., msgread & msgwrite).

NiLuJe:
Ah. It's NDK code, so, who knows why the API design is weird and redundant ;o).

https://android.googlesource.com/platform/ndk/+/master/sources/android/native_app_glue
https://github.com/android/ndk-samples/blob/main/native-activity/app/src/main/cpp/main.cpp

assets/android.lua
```diff
@@ -111,6 +111,9 @@ int ALooper_pollAll(int timeoutMillis, int* outFd, int* outEvents, void** outDat
 int ALooper_pollOnce(int timeoutMillis, int* outFd, int* outEvents, void** outData);
 
 // for alooper.fifo messages
+enum  {
+    ALOOPER_EVENT_FIFO_MESSAGE = 1<<5,
+};
```
NiLuJe, line 116: There's already en enum w/ a list of `LOOPER_ID_` in `jni/android_native_app_glue/android_native_app_glue.h`, i'd add it there, > `LOOPER_ID_USER` ;).

app/src/main/java/org/koreader/launcher/Device.kt
```diff
@@ -75,13 +82,60 @@ class Device(activity: Activity) {
...
+
+        // These are the messages going over the fifo to the native glue.
+        // 32bit (4-byte) get transmitted. The low byte is the command
+        private const val EVENT_POWER_CONNECTED = 100
+        private const val EVENT_POWER_DISCONNECTED = 101
+    }
+
+    inner class EventReceiver : BroadcastReceiver() {
+        override fun onReceive(context: Context?, intent: Intent?) {
+            val action = intent?.action
```
pazos, line 94: There's a idomatic way in kotlin of handling nullable vals. You can do:

```kotlin

intent?.let {
    when(it.action) {
    // here the cases
    }
}
```

you can also rename `it` to something you like more:

```kotlin

intent?.let { event ->
    when(event.action) { 
    // here the cases
    }
}
```

This way you can avoid NPE when evaluating with `when` :)

Not sure if is strictly neccesary here, but it is a good way of dealing with null types

app/src/main/java/org/koreader/launcher/Device.kt
```diff
@@ -75,13 +82,60 @@ class Device(activity: Activity) {
...
+            val writer = FileWriter(MainApp.fifo_path, true)
+            val fifo_bytes = CharArray(4)
+            fifo_bytes[0] = (state and 0xFF).toChar()
+            fifo_bytes[1] = ((state ushr 8) and 0xFF).toChar()
+            fifo_bytes[2] = ((state ushr 16) and 0xFF).toChar()
+            fifo_bytes[3] = ((state ushr 24) and 0xFF).toChar()
+            writer.write(fifo_bytes, 0, 4)
+            writer.close()
+        } catch (e: Exception) {
+            Logger.e("Feed ALooper fifo: ERROR writing to  ALooper fifo: \"$MainApp.fifo_path\"")
```
pazos, line 136: A few nits:

1. You can avoid escaping the string with

```kotlin

"this is a string with a value inside. The value is ${MainApp.fifo_path}"
```

2. Please print the exception you catched or a stack trace (or both)
3. Not sure about the message body, sounds pretty arcane :/. In any case I would prefeer to use a proper tag and a message. Asumming the tag is `"Feed Alooper fifo"` and the message is `"error writing to file ${MainApp.fifo.path}: \n$e"`

You write: 
```kotlin
Logger.e("Feed Alooper fifo", "error writing to file ${MainApp.fifo.path}: \n$e")
```

All messages delivered by managed code (code that runs within the JVM) use a tag to differenciate them. Only the frontend code (and a few bits of very generic info) are used without a tag (they use the generic `KOReader` tag)

app/src/main/java/org/koreader/launcher/MainActivity.kt
pazos: please remove this one: very nit picky but you can add changes in a separate PR if you think is worth your time :)

assets/android.lua
```diff
@@ -110,6 +110,12 @@ typedef int (*ALooper_callbackFunc)(int fd, int events, void* data);
 int ALooper_pollAll(int timeoutMillis, int* outFd, int* outEvents, void** outData);
 int ALooper_pollOnce(int timeoutMillis, int* outFd, int* outEvents, void** outData);
 
+// for alooper.fifo messages
```
pazos, line 113: another nit:

Our own enums begin in https://github.com/koreader/android-luajit-launcher/blob/c53e6b2b29d8bb53e449e4994cf644ac1f99dcc8/assets/android.lua#L228

There're a few of them (for network, keycodes, screen timeout, haptic feedback ...) Please place them where you see it fits

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -225,8 +230,38 @@ static void* android_app_entry(void* param) {
...
+        }
+    } else {
+        LOGV("%s: FIFO \"%s\" created", TAG, fifo_file);
+    }
+
+    int fifo_fd = open(fifo_file, O_RDWR | O_CLOEXEC);
+    if (fifo_fd  == -1) {
+        LOGE("%s: FIFO errnor=0%x", TAG, errno);
+    } else {
+//        android_app->userData = message_to_lua;
```
pazos, line 252: please remove these comments

NiLuJe:
Friendly reminder that switching to a pipe will probably save you the trouble of dealing with the FS and paths ;).

> What looking at this revealed is that the CMD handling already functions in a similar manner, except it's using a pipe, not a named pipe.
> 
> Which makes sense: you don't actually need to access this outside of the process, so a simple anonymous pipe might make all the I/O and FS-related mess much clearer (c.f., msgread & msgwrite).

pazos (commented):
Untested Ack. Looks very good and clean. A few very minor nits and it is ready for merging!!

I'm going to (ab)use this method for a few thingies. Namely: multiwindow state changes and apk download callbacks (ie: prompt to install the update when it is downloaded, like in other platforms)

But this method open a lot of doors: you can now write all kind of widgets and [shortcuts](https://developer.android.com/guide/topics/ui/shortcuts)

Very good work!!

zwim:
> Friendly reminder that switching to a pipe will probably save you the trouble of dealing with the FS and paths ;).
> 

> Friendly reminder that switching to a pipe will probably save you the trouble of dealing with the FS and paths ;).
> 
> > What looking at this revealed is that the CMD handling already functions in a similar manner, except it's using a pipe, not a named pipe.
> > Which makes sense: you don't actually need to access this outside of the process, so a simple anonymous pipe might make all the I/O and FS-related mess much clearer (c.f., msgread & msgwrite).

Thank you for the reminder. Unfortunately I think this is above my ability to implement. 

I don't see how/where I should create the pipe and how I can write to the pipe from `device.kt`. Where do I get the fd from?

NiLuJe:
On the C side: You create it basically the same place you currently use `mkfifo`, you can store both ends of the pipe in the giant struct like CMD does for `msgwrite`/`msgread`.

On the Lua side: it's returned by ALooper in the fd pointer.

zwim:
C and LUA side seems clear, but I need the filedescriptor of the pipe in kotlin (Device.kt), so that I can write the messages there. Kotlin is my problem.

NiLuJe:
Ah, that may be a good point. I have zero clue about Java/Kotlin, so I completely ignored this aspect of the equation ;p.

If you don't have access to the `android_app` object from there (we do in Lua, so I assumed we did in Java, but, again: no clue ^^), then, yeah, the file handle avoids that issue.

app/src/main/java/org/koreader/launcher/Device.kt
```diff
@@ -75,13 +82,60 @@ class Device(activity: Activity) {
...
+            val writer = FileWriter(MainApp.fifo_path, true)
+            val fifo_bytes = CharArray(4)
+            fifo_bytes[0] = (state and 0xFF).toChar()
+            fifo_bytes[1] = ((state ushr 8) and 0xFF).toChar()
+            fifo_bytes[2] = ((state ushr 16) and 0xFF).toChar()
+            fifo_bytes[3] = ((state ushr 24) and 0xFF).toChar()
+            writer.write(fifo_bytes, 0, 4)
+            writer.close()
+        } catch (e: Exception) {
+            Logger.e("Feed ALooper fifo:", "ERROR writing to file: ${MainApp.fifo_path}: \n$e")
```
pazos, line 137:
```suggestion
            Logger.e("FIFO", "Cannot write to file ${MainApp.fifo_path}: \n$e")
```

pazos: As it is already logged as an ERROR in logcat :p

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -225,8 +230,36 @@ static void* android_app_entry(void* param) {
...
     pthread_mutex_unlock(&android_app->mutex);
 
+    // create and open fifo for communication with MainActivity
+    const char* fifo_path = android_app->activity->internalDataPath;
+    char fifo_file[strlen(fifo_path) + strlen(FIFO_NAME) + 2U]; // +1 for "/" and +1 for NULL
+    snprintf(fifo_file, sizeof(fifo_file), "%s/%s", fifo_path, FIFO_NAME);
+
+    if (mkfifo(fifo_file, 0666) == -1) {
+        if (errno == EEXIST) {
+            LOGV("%s: FIFO \"%s\" already exists", TAG, fifo_file);
```
pazos, line 240:
```suggestion
            LOGV("%s: file %s already exists", TAG, fifo_file);
```

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -225,8 +230,36 @@ static void* android_app_entry(void* param) {
...
+        } else {
+            LOGE("%s: FIFO \"%s\" cannot be created!", TAG, fifo_file);
```
pazos, line 242:
```suggestion
            LOGE("%s: file %s cannot be created!", TAG, fifo_file);
```

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -225,8 +230,36 @@ static void* android_app_entry(void* param) {
...
+        }
+    } else {
+        LOGV("%s: FIFO \"%s\" created", TAG, fifo_file);
```
pazos, line 245:
```suggestion
        LOGV("%s: file %s created", TAG, fifo_file);
```

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -225,8 +230,36 @@ static void* android_app_entry(void* param) {
     pthread_cond_broadcast(&android_app->cond);
     pthread_mutex_unlock(&android_app->mutex);
 
+    // create and open fifo for communication with MainActivity
```
pazos, line 233:
```suggestion
    */ from https://github.com/koreader/android-luajit-launcher/pull/294

    We use a named pipe to push events to our main loop running in LuaJIT.
    Here we create the file, open it and register it as a poll source if everything went well.
    
    Its state is available as part of the android ffi module
*/
```

pazos: or something as verbose as that :)

assets/android.lua
```diff
@@ -362,6 +362,11 @@ enum {
     ASCREEN_ORIENTATION_FULL_SENSOR = 10,
 };
 
+enum {
+    EVENT_POWER_CONNECTED = 100,
```
pazos, line 366: A little nit I forgot.

Please use AEVENT instead of EVENT. Android uses that in their native activity API and we use that too in our own glue code

`AHAPTIC*`, `ASCREEN*` ...

Maybe there're better ways of calling the part we implement on our own but should be consistent and in other PR :)

pazos (approved):
Final round of nits. Looks almost ready to me, @NiLuJe @zwim

pazos:
One question that I missed on the review: what happens when a system broadcast event happens and the app is running in the background? Is the event received by the kotlin side?. 

I'm pretty sure our own lua loop is running from `onCreate` to `onDestroy` and we're able to consume relayed events when the app is not focused but I'm not sure how broadcast receivers work when they're part of an activity

NiLuJe:
Nothing to add on the C/Lua front, thanks for bearing with me and my wild ideas ;).

app/src/main/java/org/koreader/launcher/Device.kt
```diff
@@ -75,13 +82,60 @@ class Device(activity: Activity) {
...
+
+        event = EventReceiver()
+
+        val filter = IntentFilter()
+        filter.addAction(Intent.ACTION_POWER_CONNECTED)
+        filter.addAction(Intent.ACTION_POWER_DISCONNECTED)
+        activity.applicationContext.registerReceiver(event, filter)
+    }
+
+    // post an 32 bit message which can be evaluated after an ALooper_pollAll()
```
NiLuJe, line 124: Stupid grammar nit:
```suggestion
    // post a 32-bit message which can be evaluated after an ALooper_pollAll()
```

app/src/main/java/org/koreader/launcher/Device.kt
```diff
@@ -75,13 +82,60 @@ class Device(activity: Activity) {
         private const val LINUX_LANDSCAPE = 1
         private const val LINUX_REVERSE_PORTRAIT = 2
         private const val LINUX_REVERSE_LANDSCAPE = 3
+
+        // These are the messages going over the fifo to the native glue.
+        // 32bit (4-byte) get transmitted. The low byte is the command
```
NiLuJe, line 87:
```suggestion
// 32 bits (4 bytes) get transmitted. The low byte is the command.
```

zwim:
> One question that I missed on the review: what happens when a system broadcast event happens and the app is running in the background? Is the event received by the kotlin side?.
> 
> I'm pretty sure our own lua loop is running from `onCreate` to `onDestroy` and we're able to consume relayed events when the app is not focused but I'm not sure how broadcast receivers work when they're part of an activity

On my phone: 
I start KOReader connected to power -> I CAN see the charging symbol in the status line.
KOReader to background, playing with some apps.
Unplugging the phone.
Switch to KOReader -> I CAN'T see the charging symbol in the status line.

So I think the mechanism works in background too.

jni/android_native_app_glue/android_native_app_glue.c
pazos: @zwim: this one too, please. And it's ready to be merged :)

jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -225,8 +230,42 @@ static void* android_app_entry(void* param) {
...
+    /* from https://github.com/koreader/android-luajit-launcher/pull/294
+
+    We use a named pipe to push events to our main loop running in LuaJIT.
+    Here we create the file, open it and register it as a poll source if everything went well.
+
+    Its state is available as part of the android ffi module.
+    */
+    const char* fifo_path = android_app->activity->internalDataPath;
+    char fifo_file[strlen(fifo_path) + strlen(FIFO_NAME) + 2U]; // +1 for "/" and +1 for NULL
+    snprintf(fifo_file, sizeof(fifo_file), "%s/%s", fifo_path, FIFO_NAME);
```
NiLuJe, line 242: Didn't we conclude that `sizeof` on a VLA was a bad idea?

NiLuJe: I'm genuinely curious what this does. It's *conceivable* that the compiler groks what happens and actually does

```C
size_t len = strlen(fifo_path) + strlen(FIFO_NAME) + 2U;
char fifo_file[len];
snprintf(fifo_file, len, "%s/%s", fifo_path, FIFO_NAME);
```

But I'd probably be more confident in making that explicit ;).

PR #298, opened 2021-04-12
Broadcast Receiver class

pazos:
Sorry @zwim, as most intents here were covered by your original PR (ie: separate file for the broadcast receiver class). I just found it confusing on previous iterations but now it makes all the sense in the world (to me :p).

A few changes:

- Be very verbose with the event filter and when a new event is registered/dispatched.

On app launch something like this will be logged:

```
V  [EventReceiver] Filtering 2 events:
V  android.intent.action.ACTION_POWER_CONNECTED
V  android.intent.action.ACTION_POWER_DISCONNECTED
```

- Use a hashmap between android broadcast events/intents and our event codes

That makes easier to watch new events, just asign the map between the intent and the event code in `init` and the new event will be automatically added to the filter and its asigned code pushed to the named pipe on event arrival.

~~- WIP: use the mechanism to manage APK downloads, so we can repurpose the ("app is ready to be updated, do you want to do it now?" used on other platforms).~~. Will be in a separate PR

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/298)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/Device.kt
+++ b/app/src/main/java/org/koreader/launcher/Device.kt
@@ -2,10 +2,6 @@ package org.koreader.launcher
 
 import android.app.Activity
 import android.app.AlertDialog
-import android.content.BroadcastReceiver
-import android.content.Context
-import android.content.Intent
-import android.content.IntentFilter
 import android.os.Build
 import android.view.Gravity
 import android.view.Surface
@@ -16,7 +12,6 @@ import android.widget.TextView
 import org.koreader.launcher.device.DeviceInfo
 import org.koreader.launcher.device.EPDFactory
 import org.koreader.launcher.device.LightsFactory
-import java.io.FileWriter
 import java.util.*
 
 class Device(activity: Activity) {
@@ -62,8 +57,6 @@ class Device(activity: Activity) {
     // Hardware orientation for this device (matches android logo)
     private var screenIsLandscape: Boolean = false
 
-    private var event: EventReceiver
-
     companion object {
         private const val TAG = "DeviceHelper"
         private const val LIGHT_DIALOG_CLOSED = -1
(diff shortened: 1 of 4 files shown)


zwim:
Cannot test this now, but it looks like an evolution (not like a mutation :) ) to me.

As I am not so familiar with kotlin, I might not find all pitfalls, I belive you.

Maybe we can add an update of the status bars after a a creen unlock in this way, too (here I notice freezes of the time). We will seeee.

pazos:
> Maybe we can add an update of the status bars after a a creen unlock in this way, too (here I notice freezes of the time). We will seeee.

there's no broadcast event for that. But if you manage to find a way to do it on the frontend (ie: a new event that updates the time) we can do it on MainActivity's `onResume`. Maybe we can keep the function `push` public to be available from MainActivity? That makes possible to push our own codes that are not broadcast events.

NiLuJe:
Isn't there a window destroy/focus cycle on suspend/resume?

pazos:
> Isn't there a window destroy/focus cycle happening on suspend/resume?

IIRC A surface is created `onCreate` and destroyed `onDestroy`. The surface is changed `onResume`/`onPause`. A hw rota means the destruction and recreation of the surface.

> I mean, we get a destroy when switching tasks, so I'd assume a suspend would be at least as "violent" ^^.

Not really, unless the system is tight on ram. We get onPause -> onStop. Some systems never trigger onStop until a few minutes in the background. onDestroy is usually used by the framework to restore things when the config changes, but we prevent all that in https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/AndroidManifest.xml#L42

The main thread (outside the JVM) runs happily until onDestroy happens, no matter if the app is in the background or in the foreground.

A suspend (from ~~app's~~ activities' PoV) is no different than pressing the home button to return to the launcher.

zwim:
An idea to throw in: Wakening the device from sleep means to press a key (maybe powerBotton)y. We use that key to fire onRefreshWhatever

NiLuJe:
Oops, my bad, I mean `APP_CMD_TERM_WINDOW` and the FOCUS stuff, not `DESTROY`. I don't have a debug build ready to get you (an dme, hence the confusion ;p) proper logs right now, but, IIRC, yeah, a HW rota tripped the same kind of stuff than a task switch, which is what I was thinking of.

pazos:
https://github.com/koreader/android-luajit-launcher/pull/298#issuecomment-818235798 doesn't make any sense. I tend to overcomplicate things :p. There's an already available [cmd for resume](https://github.com/koreader/koreader/blob/master/frontend/device/android/device.lua#L169). Just need to hook events there. Sorry about the confusion

Frenzie (approved):
lgtm

PR #303, opened 2021-05-02
Misc changes

pazos:
1. Bump a few app dependencies. Including both the gradle and kotlin plugins, the gradle distribution and leakcanary.
2. Setup a couple of settings at build time instead of guessing them at run time based on flavor. `IN_APP_UPDATES` & `SUPPORTS_RUNTIME_CHANGES`
3. Improved download & install APK functions. The frontend is notified when a new APK is download, so it can prompt for installation.
4. Moved build constants to MainApp.

Everything should work fine as is, but I'm not sure if there's a way to test that on the build server without doing the usual dance: merge here, merge frontend changes, see if something crashes.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/303)


app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -231,14 +239,41 @@ class MainActivity : NativeActivity(), JNILuaInterface,
...
-        return ApkUpdater.download(this, url, name)
+        return if (MainApp.has_ota_updates) {
+            val path = Environment.getExternalStoragePublicDirectory(
+                Environment.DIRECTORY_DOWNLOADS).toString() + "/" + name
+
+            val request = DownloadManager.Request(Uri.parse(url))
+            request.setTitle(MainApp.name)
+            request.setDescription(name)
+            request.setMimeType("application/vnd.android.package-archive")
+            request.setNotificationVisibility(DownloadManager.Request.VISIBILITY_HIDDEN)
```
pazos, line 257: that's wrong, It should be visible for now. I played a bit with the following:

hide notification
register a receiver on download complete
prompt for APK update when it is ready.

In that case VISIBILITY_HIDDEN could be welcome as would be redundant. It requires a manifest permission for the rocks flavor.

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -231,14 +239,41 @@ class MainActivity : NativeActivity(), JNILuaInterface,
...
+                val manager = applicationContext.getSystemService(Context.DOWNLOAD_SERVICE)
+                    as DownloadManager
+                manager.enqueue(request)
+                DOWNLOAD_OK
+            } catch (e: Exception) {
+                e.printStackTrace()
+                DOWNLOAD_FAILED
+            }
+        } else  {
+            DOWNLOAD_NOT_SUPPORTED
```
pazos, line 275: We don't use that in the frontend, as we check if the flavor hasOtaUpdates, but for the sake of completion I think it is fine to include it here.

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -231,14 +239,41 @@ class MainActivity : NativeActivity(), JNILuaInterface,
...
-        } ?: Logger.e(TAG_MAIN, "invalid lookup: no text")
+            } ?: Logger.e(tag, "invalid lookup: no action")
+        } ?: Logger.e(tag, "invalid lookup: no text")
     }
 
     override fun download(url: String, name: String): Int {
-        return ApkUpdater.download(this, url, name)
+        return if (MainApp.has_ota_updates) {
+            val path = Environment.getExternalStoragePublicDirectory(
+                Environment.DIRECTORY_DOWNLOADS).toString() + "/" + name
```
Frenzie, line 251: Not super relevant to Android, but isn't that usually something like File.separator?

Frenzie (approved):
lgtm

app/src/main/java/org/koreader/launcher/MainActivity.kt
pazos: done :)

PR #305, opened 2021-05-08
simplify assets extraction

pazos:
A few changes.

1. Instead of extracting the 1st 7z file iterate over the entire assets store and extract them all.
2. APK bundled version isn't part of the 7z name. Now it is stored as a separate text file within the assets module.
3. Instead of complicated logic based on the assets present on the APK we use a for loop and extract everything available based on its type.

For standalone apps based on android-luajit-launcher module the behaviour would be the same. If the APK contains a version.txt then extract only if bundled differs from installed. Else extract on every launch.

See how the inner of the APK looks:
test

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/305)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/Assets.kt
+++ b/app/src/main/java/org/koreader/launcher/Assets.kt
@@ -11,183 +11,126 @@ import androidx.core.content.ContextCompat
 import java.io.*
 
 class Assets {
-    companion object {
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
lgtm

PR #307, opened 2021-05-12
Bump LuaJIT

jimman2003:
Used by https://github.com/koreader/koreader-base/pull/1370

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/307)


NiLuJe (approved):

pazos:
Sadly this LuaJIT bump fails to build against arm-v7a ABI. No problem with x86. @NiLuJe could you take a look?. If not I would suggest to revert this commit as changes are fairly minor and it is blocking submodule bumps in front

NiLuJe:
Do you have the exact compiler output?

pazos:
Nevermind. It works now. I have no idea why :smile:

NiLuJe:
Yeah, was going to say it builds in a clean build ;p.

Frenzie:
Also I believe I already tried it a few days ago, too. But sometimes I have
to remove/reset some stuff in that repo; haven't bothered to investigate in
depth.

PR #313, opened 2021-05-27
[wip]: refactor + proguard

pazos:
Fairly minor changes in the grand scheme of things :)

Mainly to behave like a "normal" android app: with proguard/R8 rules and JIT/AOT compilation of kotlin classes.

I changed a few other things and removed some useless logs. 

* add proguard rules (don't obfuscate, keep line numbers for debugging stack traces, keep the entire `org.koreader.launcher` package)
* avoid packaging useless metadata into the APK
* refactor classes, finish migration to MultidexApp
* simplify `launcher.lua`
* use different log levels on `dlopen.lua`
* enable AOT compilation of JVM code

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/313)

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -1,0 +1,9 @@
+# proguard is used to shrink and optimize dependencies.
+-keep class org.koreader.launcher.** { *; }
+
+# keep kotlin.Metadata annotations
+-keepattributes RuntimeVisibleAnnotations
+-keep class kotlin.Metadata { *; }
+
+# preserve the line number information for debugging stack traces.
+-keepattributes SourceFile,LineNumberTable
--- a/assets/launcher.lua
+++ b/assets/launcher.lua
@@ -5,39 +5,14 @@ local C = ffi.C
 
 ffi.cdef[[
 int chdir(const char *path);
-char *getcwd(char *buf, size_t size);
 ]]
-
-local function chdir(path)
-    if C.chdir(path) == 0 then
-        return true
-    else
-        return nil, "Unable to change working directory to '" ..path.."'"
-    end
-end
-
-local max_path = 4096
-local function currentdir()
-    return ffi.string(
-        C.getcwd(ffi.new('char[?]', max_path), max_path))
-end
-
---require("test")
 
 -- the default current directory is root so we should first of all
 -- change current directory to application's data directory
-if chdir(A.dir) then
-    local msg = "Change directory to assets dir"
-    local cwd = currentdir()
-    if A.dir ~= cwd then
-        -- multi user environment
-        A.LOGI(string.format("%s: %s -> %s", msg, A.dir, cwd))
-    else
-        -- single user environment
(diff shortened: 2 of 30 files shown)


app/proguard-rules.pro
```diff
@@ -0,0 +1,3 @@
+-dontobfuscate
+-keepattributes SourceFile,LineNumberTable
+-printusage /home/pazos/usage.txt
```
pazos, line 3: This line needs to be removed.

assets/launcher.lua
```diff
@@ -5,39 +5,14 @@ local C = ffi.C
...
-    local cwd = currentdir()
-    if A.dir ~= cwd then
-        -- multi user environment
-        A.LOGI(string.format("%s: %s -> %s", msg, A.dir, cwd))
-    else
-        -- single user environment
-        A.LOGI(string.format("%s: %s", msg, cwd))
-    end
-else
-    A.LOGE("Cannot change directory to "..A.dir)
```
Frenzie, line 40: Was this moved into Java somewhere?

Frenzie (approved):
Not sure I quite grok the why for the most part but it looks like fairly meaningless changes. :-P

pazos:
> Not sure I quite grok the why for the most part but it looks like fairly meaningless changes. :-P

Do you mean the refactor, right?

The package `org.koreader.launcher.device` is now standalone. You can copy the entire folder into another project and use the few nice things that are unique to that package (ie: Lights & EPD drivers, device & platform ids ...)

```kotlin
val device = org.koreader.device.Device(aContext)
```

The package `org.koreader.launcher` contains stuff that's mostly useful for this app and/or has references to MainApp companion object.

There're a few things that I didn't touch yet but I'm planning to do:

-  move EPDTest to `org.koreader.launcher.device` package.
- add LightsTest to `org.koreader.launcher.device` package.
- add a hub `DeviceTestActivity` to `org.koreader.launcher.device` that links to all device specific test
- refactor the whole package as a library

Frenzie:
> Do you mean the refactor, right?

Yes, it basically looks like some meaningless renaming to me. :-)

assets/launcher.lua
pazos: Not really, I added a few logs in the past which aren't too useful. This specific bits logs:

For multi-user environments: `Change directory to assets dir: /data/data/org.koreader.launcher/files -> /data/user/user_id/org.koreader.launcher.files`

For single-user environments: `Change directory to assets dir: /data/data/org.koreader.launcher/files`

The path where the app is started is always `/data/data/org.koreader.launcher`. But google says: do not hardcode it, it might change. Use `context.filesDir`. We keep it as `MainApp.assets_path`.

Also, `chdir` should never fail. Raising and error and log failures makes sense there.

Frenzie: Cool, thanks.

PR #314, opened 2021-05-31
migrate helper objects to kotlin extensions.

pazos:
---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/314)


app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -18,6 +18,7 @@ import androidx.core.app.ActivityCompat
 import androidx.core.content.ContextCompat
 import org.koreader.launcher.device.Device
 import org.koreader.launcher.utils.*
+import org.koreader.launcher.extensions.*
```
Frenzie, line 21: Is there a specific order? If not I'd go with alphabetic.

Frenzie (approved):

Frenzie (approved):
lgtm! :-)

PR #316, opened 2021-06-09
Some early init changes

pazos:
- Block until permissions are granted before launching the LuaJIT vm. 
- Don't raise errors if runtime permissions were rejected (can happen). Just show a toast and finish the activity.
- Annotate all methods called from lua as `@WorkerThread`
- Continue migration to kotlin extensions.
- Reworked mandatory permissions

I refactored a bit the C/JNI code. Probably some stuff in `jni_helper.c` can be abstracted away, but I'm fine as is and it seems to work fine :)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/316)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/MainActivity.kt
+++ b/app/src/main/java/org/koreader/launcher/MainActivity.kt
@@ -15,7 +15,6 @@ import android.net.Uri
 import android.os.BatteryManager
 import android.os.Build
 import android.os.Bundle
-import android.provider.Settings
 import android.util.Log
 import android.view.DisplayCutout
 import android.view.SurfaceHolder
@@ -31,7 +30,7 @@ import java.io.File
 import java.util.Locale
 
 class MainActivity : NativeActivity(), LuaInterface,
-    ActivityCompat.OnRequestPermissionsResultCallback{
+    ActivityCompat.OnRequestPermissionsResultCallback {
 
     private val tag = this::class.java.simpleName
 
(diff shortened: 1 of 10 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -7,6 +7,9 @@
     <string name="epd_test_question">Did you see a flashing black to white eink update?</string>
     <string name="epd_test_instructions">Go to github.com/koreader/koreader/issues/4551 and share the following information with us</string>
     <string name="epd_test_description">Supported platforms</string>
+    <string name="error_no_permissions">Insufficient permissions</string>
+    <string name="error_extracting_assets">Error extracting assets</string>
+    <string name="error_quit_button">Quit</string>
```
pazos, line 12: unused. Remove me

pazos:
Somehow I killed the 3 dots animation while uncompressing assets. Need to figure out why.

pazos:
> Somehow I killed the 3 dots animation while uncompressing assets. Need to figure out why.

Extracting assets in the main thread blocks the animation. We still need to call `android.extractAssets()` from Lua. It is blocking on that thread (`NativeThread`) but leaves the `UIThread` responsive to show the animation.

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -195,14 +183,15 @@ class MainActivity : NativeActivity(), LuaInterface,
...
         Array<String>, grantResults: IntArray) {
-        Log.d(tag, "onRequestPermissionResult()")
         super.onRequestPermissionsResult(requestCode, permissions, grantResults)
         if (hasStoragePermissionCompat()) {
             Log.i(tag, String.format(Locale.US,
                     "Permission granted for request code: %d", requestCode))
         } else {
-            Log.e(tag, String.format(Locale.US,
-                    "Permission rejected for request code: %d", requestCode))
+            Log.e(tag, "Permission rejected. Bye!")
```
Frenzie, line 191: lol

Frenzie (approved):
Looks sane to me.

app/src/main/java/org/koreader/launcher/MainActivity.kt
pazos: :smile: 

We can do it better while keeping the spirit. Wait a bit  :p

pazos: Fixed with a general approach instead of our hackish one: for each mandatory permission exit with a message if it wasn't granted.

Of course, the only mandatory permission right now is WRITE_EXTERNAL_STORAGE, but that can be changed at will.

The message would look like

```
mandatory permission rejected: android.permission.WRITE_EXTERNAL_STORAGE. Bye!
```

Frenzie: That is indeed better. But I was mainly just amused at the message. :-)

pazos: It's fun because it's true. Also it is much better than the workflow Google suggests: 
workflow-runtime

workflow-runtime

pazos:
wops, pushed the button without changing the squash message :/

PR #325, opened 2021-07-01
fix symlink on Pre-Lollipop

pazos:
Should fix https://github.com/koreader/koreader/issues/7920

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/325)


Frenzie (approved):
Alright. :+1:

PR #327, opened 2021-07-25
add Boyue Lemon Read M1

pazos:
https://github.com/koreader/koreader/issues/4551

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/327)


NiLuJe:
Most unfortunate device name ever? :D

PR #333, opened 2021-08-16
nabuk regal hd

pazos:
https://github.com/koreader/koreader/issues/4551

wip: waiting for confirmation.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/333)

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -2236,7 +2236,7 @@ local function run(android_app_state)
             delay_fast = 0
             fast, partial = wf_du + mode_partial, wf_gc16 + mode_partial
             if platform == "freescale" then
-                local wf_regal = 7
+                local wf_regal = android.prop.product == "nabukreg_hd" and 6 or 7
                 full = wf_gc16 + mode_full
                 full_ui = wf_regal + mode_full
                 partial_ui = wf_regal + mode_partial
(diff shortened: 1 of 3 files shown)


assets/android.lua
```diff
@@ -2236,7 +2236,7 @@ local function run(android_app_state)
             delay_fast = 0
             fast, partial = wf_du + mode_partial, wf_gc16 + mode_partial
             if platform == "freescale" then
-                local wf_regal = 7
+                local wf_regal = android.prop.product == "nabukreg_hd" and 6 or 7
```
Frenzie, line 2239: Strange.

Frenzie (approved):
lgtm

assets/android.lua
NiLuJe: Did we ever get a look at the decompiled code of the system service that handles the refreshes on these devices?

pazos: Yup, @char11 did all the work. It is documented in https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/device/epd/freescale/NTXEPDController.kt.

It is the controller used for all freescale/ntx.

IIRC we don't interface directly with the library that handles the update. the view.View class contains methods that perform native calls via jni.

Some freescale boards have system drivers that don't work with native activities and thus require that we handle all updates. On the rest we handle only full eink updates.

PR #350, opened 2022-01-07
Minor fixes

pazos:
Don't squash!

1. The test activity now probes the device itself and shows if it is already supported in KOReader (and which drivers it uses)
device-2022-01-07-161316

2. Fixed title not displaying properly on LightDialog. Also added an horizontal line and some margin if the dialog title isn't empty
device-2022-01-07-161523

3. If the title is empty the dialog will look much simpler. That's my suggestion for the frontend. It will look like:

device-2022-01-07-161647

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/350)


Frenzie (approved):
Surprisingly simple addition! :-)

PR #372, opened 2022-06-12
Add Onyx SDK lights controller, Onyx Note Air2, & Onyx Nova Air C

bb010g:
Follow-up to #342. When classes or methods don't exist, KOReader doesn't crash, addressing koreader/koreader#8584.

Tested on an Onyx Nova Air C. cc/ @bezmi @Galunid

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/372)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/lights/OnyxSdkLightsController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/lights/OnyxSdkLightsController.kt
@@ -1,0 +1,146 @@
+package org.koreader.launcher.device.lights
+
+import android.app.Activity
+import android.util.Log
+import org.koreader.launcher.device.LightsInterface
+import android.content.Context
+import java.lang.Class.forName
+import java.lang.reflect.Method
+
+class OnyxSdkLightsController : LightsInterface {
+    companion object {
+        private const val TAG = "Lights"
+        private const val BRIGHTNESS_MAX = 255
+        private const val WARMTH_MAX = 255
+        private const val MIN = 0
+    }
+
+    override fun getPlatform(): String {
+        return "onyx-sdk-lights"
+    }
+
+    override fun hasFallback(): Boolean {
+        return false
+    }
+
+    override fun hasWarmth(): Boolean {
+        return true
+    }
+
+    override fun needsPermission(): Boolean {
+        return false
+    }
+
+    override fun getBrightness(activity: Activity): Int {
+        return FrontLight.getCold(activity)
+    }
+
+    override fun getWarmth(activity: Activity): Int {
+        return FrontLight.getWarm(activity)
+    }
+
+    override fun setBrightness(activity: Activity, brightness: Int) {
+        if (brightness < MIN || brightness > BRIGHTNESS_MAX) {
+            Log.w(TAG, "brightness value of of range: $brightness")
+            return
+        }
+        Log.v(TAG, "Setting brightness to $brightness")
+        FrontLight.setCold(brightness, activity)
+    }
+
+    override fun setWarmth(activity: Activity, warmth: Int) {
+        if (warmth < MIN || warmth > WARMTH_MAX) {
+            Log.w(TAG, "warmth value of of range: $warmth")
+            return
+        }
+        Log.v(TAG, "Setting warmth to $warmth")
+        FrontLight.setWarm(warmth, activity)
+    }
+
+    override fun getMinWarmth(): Int {
+        return MIN
+    }
+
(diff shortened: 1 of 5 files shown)


pazos:
@bb010g: thanks for your contribution!

Please keep refactoring (sorting, device ids simplification) separated from new features.

You can make a different PR with the refactor. It will be merged soon. Then you can rebase this PR on top of master with only the bits of logic you've added.

bb010g:
@pazos The reformatting commit has been split out, and this branch's commits are on top of the latest `master`. Does the reformatting commit still need to be split out? These commits are atomic, so you shouldn't need to squash them while merging.

pazos:
> Does the reformatting commit still need to be split out to a separate PR? 

Nope. Works for me as is. :+1:

app/src/main/java/org/koreader/launcher/device/lights/OnyxSdkLightsController.kt
```diff
@@ -0,0 +1,164 @@
...
+    override fun hasWarmth(): Boolean {
+        return true
+    }
+
+    override fun needsPermission(): Boolean {
+        return false
+    }
+
+    override fun getBrightness(activity: Activity): Int {
+        return try {
```
pazos, line 35: `FrontLight.getCold` should never throw an exception, so we can skip the try catch block here and just

```kotlin
override fun getBrightness(activity: Activity): Int {
    return FrontLight.getCold(activity)
}
```

app/src/main/java/org/koreader/launcher/device/lights/OnyxSdkLightsController.kt
```diff
@@ -0,0 +1,164 @@
...
+            FrontLight.getCold(activity)
+        } catch (e: Exception) {
+            Log.w(TAG, Log.getStackTraceString(e))
+            0
+        }
+    }
+
+    override fun getWarmth(activity: Activity): Int {
+        return try {
```
pazos, line 44: same as above

app/src/main/java/org/koreader/launcher/device/lights/OnyxSdkLightsController.kt
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+
+    override fun setBrightness(activity: Activity, brightness: Int) {
+        if (brightness < MIN || brightness > BRIGHTNESS_MAX) {
+            Log.w(TAG, "brightness value of of range: $brightness")
+            return
+        }
+        Log.v(TAG, "Setting brightness to $brightness")
+        try {
```
pazos, line 58: same as above

app/src/main/java/org/koreader/launcher/device/lights/OnyxSdkLightsController.kt
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+
+    override fun setWarmth(activity: Activity, warmth: Int) {
+        if (warmth < MIN || warmth > WARMTH_MAX) {
+            Log.w(TAG, "warmth value of of range: $warmth")
+            return
+        }
+        Log.v(TAG, "Setting warmth to $warmth")
+        try {
```
pazos, line 71: same

app/src/main/java/org/koreader/launcher/device/lights/OnyxSdkLightsController.kt
```diff
@@ -0,0 +1,164 @@
...
+        null
+    }
+    private const val BRIGHTNESS_CONFIG_WARM_IDX: Int = 2
+    private const val BRIGHTNESS_CONFIG_COLD_IDX: Int = 3
+
+    fun getWarm(context: Context?): Int {
+        return try {
+            getCoolWarmBrightness!!.invoke(flController!!, context, BRIGHTNESS_CONFIG_WARM_IDX) as Int
+        } catch (e: Exception) {
+            Log.w(TAG, "$e")
```
pazos, line 135: For public members of the `FrontLight` object you can choose between ` Log.w(TAG, "$e")` or `e.printStackTrace()` as you wish.

pazos (commented):
Looks quite ok!

pazos:
hey @bb010g, could you take a look at the review and update the PR?

There're some users that might benefit from the new driver waiting in https://github.com/koreader/koreader/issues/8482.

It doesn't need to be soon :tm: , though.

If you're busy elsewhere I can take care of it ;)

pazos:
Done, thanks for your work!

pazos (approved):

PR #378, opened 2022-08-01
update standalone example.

pazos:
Fixes https://github.com/koreader/android-luajit-launcher/issues/367

Didn't check if it crashes with the previous ffi def, but `poll` is the way to go.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/378)


Frenzie (approved):
Sounds good

PR #381, opened 2022-09-10
feat(MainActivity): provide "hardwareType" and "Model" from Build as "android.prop"

hasezoey:
This PR provides `Build.HARDWARE` (`ro.product.hardwareType`) and `Build.MODEL`(`ro.product.model`) in lua for a PR for the `frontend` to detect Tolino

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/381)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/LuaInterface.kt
+++ b/app/src/main/java/org/koreader/launcher/LuaInterface.kt
@@ -24,8 +24,10 @@ interface LuaInterface {
     fun getExternalSdPath(): String
     fun getFilePathFromIntent(): String?
     fun getFlavor(): String
+    fun getHardwareType(): String
     fun getLastImportedPath(): String?
     fun getLightDialogState(): Int
+    fun getModel(): String
     fun getName(): String
     fun getNetworkInfo(): String
     fun getPlatformName(): String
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/koreader/launcher/LuaInterface.kt
```diff
@@ -43,6 +43,8 @@ interface LuaInterface {
     fun getScreenWidth(): Int
     fun getStatusBarHeight(): Int
     fun getVersion(): String
+    fun getHardwareType(): String
+    fun getModel(): String
```
Frenzie, line 47: I think they're supposed to be more alphabetical?

Frenzie (approved):
lgtm give or take alphabetization nits but @pazos has the final word regardless

app/src/main/java/org/koreader/launcher/LuaInterface.kt
hasezoey: an change, but i had added it below `getVersion` to have it in a similar place

hasezoey: fixed with e77d0b848b86ebec7ea009acb7eed4fa0b635eed

pazos:
Hey @hasezoey, `android.product` is enough to identify a given tolino model.

If you want to export `Build` things to lua keep in mind we're talking about java Strings, so please double check they're not nullable. But that's already done in https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt#L168-L173
with https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt#L528-L534

So please repurpose those instead.

hasezoey:
updated the 2 functions to use `DeviceInfo` instead

pazos:
superseeded by #383

hasezoey:
closing in favor of #383

PR #382, opened 2022-09-11
Add support for setting Warmth for Tolino with `ntx_io`

### LINKED ISSUE #351: LightController for Tolinos without/with different warmth
Hi,
I run KOReader on a "Tolino Shine 3". I think it's an Android 4.4.2, the device has a frontlight and a warmth setting.

The log file doesnt save any test info below "Test logs:", only the device info. So I will type all info below. I hope that is sufficient.

Manufacturer: rakuten kobo inc.
Brand: rakutenkobo
Model: tolino
Device: ntx_6sl
Product: ntx_6sl
Hardware: e60k00
Platform: imx6

E-Ink works fine on "freescale/ntx" setting.

As for lights: Dimming works with "Tolino (warmth)" setting, but the warmth slider does not change the warmth.

I know there is a TolinoWarmthController somewhere in the code, but it seems it's not working for the shine 3.

Should I attach the report?

EDIT: KOReader version 2021.12.1

_Originally posted by @pbethke in https://github.com/koreader/koreader/issues/8482#issuecomment-1011305516_

hasezoey:
This PR adds support in `TolinoWarmthController.kt` to use `/dev/ntx_io` to set the Warmth

fixes #351
~~requires https://github.com/koreader/koreader/pull/9511~~

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/382)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/Ioctl.kt
+++ b/app/src/main/java/org/koreader/launcher/device/Ioctl.kt
@@ -29,7 +29,7 @@ open class Ioctl {
             Log.v(tag, "$device: ioctl ok, code $status")
             true
         } else {
-            val err = when(abs(status)) {
+            val err = when (abs(status)) {
                 ENOENT -> "no such file or directory"
                 EBADF -> "bad file number"
                 EACCES -> "permission denied"
--- a/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
+++ b/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
@@ -13,6 +13,10 @@ object LightsFactory {
                     logController("Tolino")
                     TolinoWarmthController()
                 }
+                DeviceInfo.LightsDevice.TOLINO_VISION5 -> {
+                    logController("Tolino")
+                    TolinoVision5WarmthController()
+                }
                 DeviceInfo.LightsDevice.ONYX_DARWIN7,
                 DeviceInfo.LightsDevice.ONYX_FAUST3,
                 DeviceInfo.LightsDevice.ONYX_KON_TIKI2,
--- a/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
+++ b/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
@@ -10,11 +10,11 @@ object LightsFactory {
         get() {
             return when (DeviceInfo.LIGHTS) {
                 DeviceInfo.LightsDevice.TOLINO_EPOS -> {
-                    logController("Tolino")
+                    logController("TolinoRoot")
                     TolinoRootController()
                 }
                 DeviceInfo.LightsDevice.TOLINO_VISION5 -> {
-                    logController("Tolino")
+                    logController("TolinoNTX")
                     TolinoNtxController()
                 }
                 DeviceInfo.LightsDevice.ONYX_DARWIN7,
--- a/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
+++ b/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
@@ -11,11 +11,11 @@ object LightsFactory {
             return when (DeviceInfo.LIGHTS) {
                 DeviceInfo.LightsDevice.TOLINO_EPOS -> {
                     logController("Tolino")
-                    TolinoWarmthController()
+                    TolinoRootController()
                 }
                 DeviceInfo.LightsDevice.TOLINO_VISION5 -> {
                     logController("Tolino")
-                    TolinoVision5WarmthController()
+                    TolinoNtxController()
                 }
                 DeviceInfo.LightsDevice.ONYX_DARWIN7,
                 DeviceInfo.LightsDevice.ONYX_FAUST3,
--- a/app/src/main/java/org/koreader/launcher/device/lights/TolinoNtxController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/lights/TolinoNtxController.kt
@@ -11,10 +11,10 @@ import java.io.File
 /* Special controller for Tolino Vision5
  * see https://github.com/koreader/android-luajit-launcher/pull/382
  *
- * Original Controller by @zwim, see "./TolinoWarmthController.kt"
+ * Original Controller by @zwim, see "./TolinoRootController.kt"
  */
 
-class TolinoVision5WarmthController : Ioctl(), LightsInterface {
+class TolinoNtxController : Ioctl(), LightsInterface {
 
     companion object {
         private const val TAG = "Lights"
--- a/app/src/main/java/org/koreader/launcher/device/lights/TolinoRootController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/lights/TolinoRootController.kt
@@ -14,7 +14,7 @@ import java.io.File
  * Thanks to @zwim
  */
 
-class TolinoWarmthController : LightsInterface {
+class TolinoRootController : LightsInterface {
     companion object {
         private const val TAG = "Lights"
         private const val BRIGHTNESS_MAX = 255
--- a/app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
@@ -19,7 +19,7 @@ class TolinoVision5WarmthController : Ioctl(), LightsInterface {
     companion object {
         private const val TAG = "Lights"
         private const val BRIGHTNESS_MAX = 255
-        private const val WARMTH_MAX = 10
+        private const val WARMTH_MAX = 10 // can also be read by ""/sys/class/backlight/lm3630a_led/max_color" on Vision5
         private const val MIN = 0
         private const val NTX_IO_FILE = "/dev/ntx_io"
         private const val NTX_WARMTH_ID = 248
--- a/app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
@@ -19,7 +19,7 @@ class TolinoVision5WarmthController : Ioctl(), LightsInterface {
     companion object {
         private const val TAG = "Lights"
         private const val BRIGHTNESS_MAX = 255
-        private const val WARMTH_MAX = 10 // can also be read by ""/sys/class/backlight/lm3630a_led/max_color" on Vision5
+        private const val WARMTH_MAX = 10 // can also be read by "/sys/class/backlight/lm3630a_led/max_color" on Vision5
         private const val MIN = 0
         private const val NTX_IO_FILE = "/dev/ntx_io"
         private const val NTX_WARMTH_ID = 248
--- a/app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
@@ -8,10 +8,10 @@ import org.koreader.launcher.device.LightsInterface
 import org.koreader.launcher.extensions.read
 import java.io.File
 
-/* Special controller for Tolino Epos/Epos2.
- * see https://github.com/koreader/koreader/pull/6332
+/* Special controller for Tolino Vision5
+ * see https://github.com/koreader/android-luajit-launcher/pull/382
  *
- * Thanks to @zwim
+ * Original Controller by @zwim, see "./TolinoWarmthController.kt"
  */
 
 class TolinoVision5WarmthController : Ioctl(), LightsInterface {
--- a/app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
@@ -14,19 +14,18 @@ import java.io.File
  * Thanks to @zwim
  */
 
-class TolinoWarmthController : LightsInterface {
+class TolinoVision5WarmthController : LightsInterface {
     companion object {
         private const val TAG = "Lights"
         private const val BRIGHTNESS_MAX = 255
         private const val WARMTH_MAX = 10
         private const val MIN = 0
+        private const val NTX_IO_FILE = "/dev/ntx_io"
+        private const val NTX_WARMTH_ID = 248
+        private const val NTX_IO_LUA_FILE = "frontend/device/kobo/ntx_io.lua"
         private const val ACTUAL_BRIGHTNESS_FILE = "/sys/class/backlight/mxc_msp430_fl.0/actual_brightness" // always readable, same for Epos2 and Vision4
-        private const val COLOR_FILE_EPOS2 = "/sys/class/backlight/tlc5947_bl/color"
-        private const val COLOR_FILE_VISION4HD = "/sys/class/backlight/lm3630a_led/color"
-        private val COLOR_FILE = if (File(COLOR_FILE_VISION4HD).exists())
-            COLOR_FILE_VISION4HD
-        else
-            COLOR_FILE_EPOS2
+        private const val COLOR_FILE = "/sys/class/backlight/lm3630a_led/color"
+        private var currentWarmth: Int? = null
     }
 
     override fun getPlatform(): String {
@@ -102,7 +101,10 @@ class TolinoWarmthController : LightsInterface {
     }
 
     override fun getWarmth(activity: Activity): Int {
-        return WARMTH_MAX - File(COLOR_FILE).read()
+        if (currentWarmth == null) {
+            currentWarmth = WARMTH_MAX - File(COLOR_FILE).read()
+        }
+        return currentWarmth ?: WARMTH_MAX
     }
 
     override fun setBrightness(activity: Activity, brightness: Int) {
@@ -124,13 +126,47 @@ class TolinoWarmthController : LightsInterface {
             Log.w(TAG, "warmth value of of range: $warmth")
             return
         }
+
+        Log.v(TAG, "Setting warmth to $warmth of $WARMTH_MAX")
+
+        try {
+            val ntxFile = File(NTX_IO_FILE)
+            // check that the ntx_io file exists
+            if (ntxFile.canWrite()) {
+                // invert, because 0 is the "warmest"(red) and 10 is the "coldest"(blue)
+                // this makes the bar for "warmth" most left blue while the right most is red
+                val invertedWarmth = (warmth - WARMTH_MAX) * -1
+                val luajitBinFile = File("./luajit")
+                // check and fix the "luajit" binary file
+                if (!luajitBinFile.canExecute()) {
+                    val chmodProcess = Runtime.getRuntime().exec("chmod 775 $luajitBinFile")
+                    val chmodProcessExitStatus = chmodProcess.waitFor()
+                    Log.v(TAG, "\"chmod 775 $luajitBinFile\" exit status $chmodProcessExitStatus")
+                }
+
+                // execute ntx_io.lua script from kobo
+                val luaBinProcess = Runtime.getRuntime().exec("$luajitBinFile $NTX_IO_LUA_FILE $NTX_WARMTH_ID $invertedWarmth")
+                val luaBinProcessExitStatus = luaBinProcess.waitFor()
+                Log.v(TAG, "\"$luajitBinFile $ntxFile $NTX_WARMTH_ID $invertedWarmth\" exit status $luaBinProcessExitStatus")
+                currentWarmth = warmth
+                return
+            } else {
+                Log.w(TAG, "\"/dev/ntx_io\" is not writeable")
+            }
+        } catch (e: Exception) {
+            Log.w(TAG, "$e")
+        }
+
+        // fallback to the original "su" way
+
         val colorFile = File(COLOR_FILE)
-        Log.v(TAG, "Setting warmth to $warmth")
+
         try {
             if (!colorFile.canWrite()) {
                 Runtime.getRuntime().exec("su -c chmod 666 $COLOR_FILE")
             }
             colorFile.write(WARMTH_MAX - warmth)
+            currentWarmth = warmth
         } catch (e: Exception) {
             Log.w(TAG, "$e")
         }
--- a/app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
@@ -23,8 +23,7 @@ class TolinoVision5WarmthController : Ioctl(), LightsInterface {
         private const val MIN = 0
         private const val NTX_IO_FILE = "/dev/ntx_io"
         private const val NTX_WARMTH_ID = 248
-        private const val ACTUAL_BRIGHTNESS_FILE = "/sys/class/backlight/mxc_msp430_fl.0/actual_brightness" // always readable, same for Epos2 and Vision4
-        private const val COLOR_FILE = "/sys/class/backlight/lm3630a_led/color"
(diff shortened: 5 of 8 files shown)


app/src/main/java/org/koreader/launcher/device/lights/TolinoWarmthController.kt
```diff
@@ -124,8 +127,40 @@ class TolinoWarmthController : LightsInterface {
...
+
+        Log.v(TAG, "Setting warmth to $warmth of $WARMTH_MAX")
+
+        try {
+            val ntxFile = File(NTX_IO_FILE)
+            // check that the ntx_io file exists
+            if (ntxFile.canWrite()) {
+                // invert, because 0 is the "warmest"(red) and 10 is the "coldest"(blue)
+                // this makes the bar for "warmth" most left blue while the right most is red
+                val invertedWarmth = (warmth - WARMTH_MAX) * -1
```
hasezoey, line 139: currently i have inverted the scale, so that the left side of the slider is the coldest and the right side is the warmest, i dont know if this is the correct way koreader wants this to have set-up, but i assume that MIN on the scale would mean coldest and MAX would mean the warmest
let me know if i should change it

NiLuJe: IIRC, the HW scale is indeed inverted (and we do want coldest at 0/left).

pazos:
Please test your code before opening a PR ;)

Also, even if this worked without root it must go on a different driver and leave this one as is.

But it won't work without root. And if you have root there're better ways of doing the same.

hasezoey:
> Please test your code before opening a PR ;)

i dont know what you mean, i did test it

> Also, even if this worked without root it must go on a different driver and leave this one as is.

why? from what i can tell, this is a controller for *Tolino* devies, and i set it up for a *tolino* device, with fallback to the old way

> But it won't work without root. And if you have root there're better ways of doing the same.

like i already said in https://github.com/koreader/koreader/pull/9511#issuecomment-1243049473, it works **without root**

pazos:
> i dont know what you mean, i did test it

Sorry about that. My fault. I was biased based on previous experiences (not with you).

> why? from what i can tell, this is a controller for Tolino devies, and i set it up for a tolino device, with fallback to the old way

because we don't have all past tolino devices to test this new way of doing stuff, which is **fragile**. And we have past experiences with changes in drivers that work on newer ones but borked legacy devices.

Just create a new driver, let people test your driver and report if it also work. We can do a device migration from one to another without nuking the old one.

> it works without root

That's fantastic. You can code it like a simple c function and call it from your driver. Like we're already doing in https://github.com/koreader/android-luajit-launcher/blob/master/jni/lzma/un7z.cpp#L9,

https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/Assets.kt#L21, 

https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/Assets.kt#L188

hasezoey:
> That's fantastic. You can code it like a simple c function and call it from your driver. Like we're already doing in 

good to hear that it is possible this way, but i dont trust my self with c

hasezoey:
also i forgot to note in the opening description of this PR: getting the warmth value does not work and defaults to MAX, if it is possible to read from ntx_io then i have not found that way yet (and `/sys/class/backlight/lm3630a_led/color` is only readable by root)

NiLuJe:
That also sounds like a familiar NTX quirk, so I doubt there's a sane solution ;).

pazos:
> also i forgot to note in the opening description of this PR: getting the warmth value does not work and defaults to MAX, if it is possible to read from ntx_io then i have not found that way yet (and `/sys/class/backlight/lm3630a_led/color` is only readable by root)

That's a drag but I think it won't be too bad. After all it will affect just the first read. After that you can keep current level into a var and update it each time the ioctl call doesn't fail.

Ofc it would be great to retrieve current warmth level from system somehow.

> good to hear that it is possible this way, but i dont trust my self with c

No problem with that. I can help. Could you please rework this PR as a different driver? I can submit a PR against your branch so you can test it.

hasezoey:
updated 8995d0da046405b9f228998c4ddc1d720f923e59 to have a new controller named `TolinoVision5WarmthController.kt` which is basically a copy of `TolinoWarmthController.kt` with modifications from this PR applied and unused values removed
also adds `TOLINO_VISION5` in `DeviceInfo` for detection

for now i have named all new things (controller and enum value in `DeviceInfo`) to be `Vision5` / `VISION5` because that is what i use and test it on, but i am up for renaming this

i have also tested the change and it works great (after keeping the state) without root, but it still requires https://github.com/koreader/koreader/pull/9511

> No problem with that. I can help. Could you please rework this PR as a different driver? I can submit a PR against your branch so you can test it.

if you can / want to do it, thanks, but i will probably replay later because i need to stop for today

pazos:
@hasezoey: done in https://github.com/hasezoey/android-luajit-launcher/pull/1

Please test when you have a moment.

I'm also interested in @NiLuJe review.

I've tried to be very verbose about errors, mostly to be useful in the test activity. `ioctl` should return `-1` and populate `errno` if there's an error or return 0 or above if ok. So in native code we just return different negative values for different errors and log them in the driver.

If some other device uses `/dev/ntx_io` or another special device that can be manipulated with `ioctl()` we can refactor the code as the function signature is generic enough to work in such cases.

hasezoey:
re https://github.com/hasezoey/android-luajit-launcher/pull/1#issuecomment-1246741868

> Then it is a mater of merging that PR and rebasing this branch on top of the new master branch :)

it seems like i dont need to merge master, because it is still up-to-date

NiLuJe:
The C bits look sound ;).

app/src/main/java/org/koreader/launcher/device/Ioctl.kt
```diff
@@ -0,0 +1,48 @@
...
+        private const val ENOTTY = 25
+    }
+
+    fun io(device: String, command: Int, args: Int): Boolean {
+        val status = ioctl(device, command, args)
+        return if (status >= 0) {
+            Log.v(tag, "$device: ioctl ok, code $status")
+            true
+        } else {
+            val err = when(abs(status)) {
```
Frenzie, line 32:
```suggestion
            val err = when (abs(status)) {
```

Frenzie:
Yup, I can't fully judge it but it doesn't smell. ;-)

app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
```diff
@@ -0,0 +1,152 @@
...
+    override fun hasWarmth(): Boolean {
+        return true
+    }
+
+    override fun needsPermission(): Boolean {
+        return false
+    }
+
+    // try to toggle on frontlight switch on Tolinos, returns the former switch state
+    override fun enableFrontlightSwitch(activity: Activity): Int {
```
pazos, line 49: AFAIK this function was implemented by @zwim to toggle the frontlight switch and requires root. Please do check if you still require this or just return an integer (see any other driver)

app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
```diff
@@ -0,0 +1,152 @@
...
+                "screen_brightness")
+        } catch (e: Exception) {
+            Log.w(TAG, e.toString())
+            0
+        }
+    }
+
+    override fun getWarmth(activity: Activity): Int {
+        if (currentWarmth == null) {
+            currentWarmth = WARMTH_MAX - File(COLOR_FILE).read()
```
pazos, line 106: IIRC you said this doesn't work without root. In that case please remove until we find a way to retrieve current warmth.

app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
```diff
@@ -0,0 +1,152 @@
...
+class TolinoVision5WarmthController : Ioctl(), LightsInterface {
+
+    companion object {
+        private const val TAG = "Lights"
+        private const val BRIGHTNESS_MAX = 255
+        private const val WARMTH_MAX = 10
+        private const val MIN = 0
+        private const val NTX_IO_FILE = "/dev/ntx_io"
+        private const val NTX_WARMTH_ID = 248
+        private const val ACTUAL_BRIGHTNESS_FILE = "/sys/class/backlight/mxc_msp430_fl.0/actual_brightness" // always readable, same for Epos2 and Vision4
```
pazos, line 26: See my other comments below. Both `ACTUAL_BRIGHTNESS_FILE` and `COLOR_FILE` will be unused if you address them.

app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
hasezoey: i dont think i need to remove that, because it works with the current implementation, it sets the correct brightness, though i dont know how exactly, but i think the following is responsible to actually set the brightness:
https://github.com/koreader/android-luajit-launcher/pull/382/files#diff-71413edbf08bf283f1da05d576b9713c940de041a959fbab0e0b9fe43f142cfeR118-R119

hasezoey: after some testing, it seems to still work, even when the whole function is commented-out, is it set somewhere else?

app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
hasezoey: it does not work a Vision5, but it would work on other models, if the old controller is anything to indicate by - i dont know if this controller will just be for Vision5 or also for other models, what do you think?

app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
hasezoey: fixed in 9603a0e6b7f7a23c1797670d0be883dca4601bf2

hasezoey:
Updated to resolve some review questions, among other things:
- change "TAG" in both the new driver and old driver to differentiate between them
- add comments explaining some variables
- add comments explaining that some value can also be read somewhere on the system

app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
hasezoey: half solved with 9603a0e6b7f7a23c1797670d0be883dca4601bf2

app/src/main/java/org/koreader/launcher/device/lights/TolinoVision5WarmthController.kt
pazos: This specific controller is easier to read than the old one. Sadly the old one does a couple of things that this one cannot do currently.

Since legacy tolinos (anything older than vision5/6) require root to install 3rd party apps I would assume the old driver is going to be used on those and your new driver will be useful on newer models.

About keeping that piece of code: now with documentation it doesn't hurt.

hasezoey: > Since legacy tolinos (anything older than vision5/6) require root to install 3rd party apps I would assume the old driver is going to be used on those and your new driver will be useful on newer models.

i am not quite sure if this applies, but i think the update version for tolino `15` seems to be android 4.4 across devices, i dont know if things are the same there, but i would assume that all devices that have tolino software `15` would have the same permissions - though like i said, i am not sure because i cannot test

also as a note, from what i can tell, installing apks on tolino (at least in tolino software 14 & 15) does not require root, you only need adb enabled OR use the secret code to access the debug menu to enable adb and/or install a apk directly

pazos:
As you can see in other driver's name I'm not very good naming stuff :p. So feel free to suggest alternatives.

My recommendation is: rename the old TolinoWarmth to TolinoRoot and the new one to TolinoNtx. IMHO the `warmth` part can be skipped since there's no need for a tolino without warmth (ie: the behaviour of brightness is like on regular android devices)

hasezoey:
Updated to rename the controllers `TolinoRoot` and `TolinoNtx` as suggested (because i also think its a better name)

app/src/main/java/org/koreader/launcher/device/lights/TolinoRootController.kt
```diff
@@ -14,9 +14,9 @@ import java.io.File
  * Thanks to @zwim
  */
 
-class TolinoWarmthController : LightsInterface {
+class TolinoRootController : LightsInterface {
     companion object {
-        private const val TAG = "Lights"
+        private const val TAG = "Lights TolinoRoot"
```
pazos, line 19: same here

app/src/main/java/org/koreader/launcher/device/lights/TolinoNtxController.kt
```diff
@@ -0,0 +1,110 @@
...
+/* Special controller for Tolino Vision5
+ * see https://github.com/koreader/android-luajit-launcher/pull/382
+ *
+ * Original Controller by @zwim, see "./TolinoRootController.kt"
+ */
+
+class TolinoNtxController : Ioctl(), LightsInterface {
+
+    companion object {
+        private const val TAG = "Lights TolinoNtx"
```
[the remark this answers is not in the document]
hasezoey: done in 87cb89124f57e0a7fd4c125068c1f2615b2d2dd0, though personally i think it is better to differentiate

app/src/main/java/org/koreader/launcher/device/lights/TolinoRootController.kt
hasezoey: done in 87cb89124f57e0a7fd4c125068c1f2615b2d2dd0, though personally i think it is better to differentiate

pazos:
great, almost ready to go!

to-do: handle the new driver in LightsFactory. I still need to figure a proper way to do that.

could you rebase on top of master to see an updated codacy analysis?

hasezoey:
> could you rebase on top of master to see an updated codacy analysis?

merged latest master

pazos:
hey, just force pushed.

When I requested a rebase I meant:

```
git checkout master
git pull origin master
git checkout tolinoNTX_IOWarmth
git rebase master
git push myremote tolinoNTX_IOWarmth -f
```

I'm fairly sure that git being git there're a million ways of doing the same

hasezoey:
> hey, just force pushed.
> When I requested a rebase I meant:

i know what rebasing is and i used it myself, but i dont like to do it when it has merge commits or a simple merge of master

hasezoey:
as asked, force-pushed a rebase onto latest master, while preserving merges (and proper signatures)
also dropped the renaming of `TAG` variables commits

pazos:
Yes, your force push broke things. It was ok when I force pushed.

hasezoey:
> Yes, your force push broke things

broke what?

pazos:
your last commit was left in the void.

Could you please fix that? Thanks

hasezoey:
> your last commit was left in the void.

what last commit?

like i said, i dropped the commits that renamed the `TAG` variable because it was reverted again (and the revert also got dropped), and from what i can tell, that were the only (aside from the "merge master" commits that are missing)

pazos:
Yup, I'm at work and understood your comment https://github.com/koreader/android-luajit-launcher/pull/382#issuecomment-1249116057 wrong. All good as is.

pazos:
And a final question because I got it wrong in the first place: with this PR in both the TestActivity's and KOReader's lightdialog work as expected?

app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
```diff
@@ -11,7 +11,11 @@ object LightsFactory {
             return when (DeviceInfo.LIGHTS) {
                 DeviceInfo.LightsDevice.TOLINO_EPOS -> {
                     logController("Tolino")
-                    TolinoWarmthController()
+                    TolinoRootController()
+                }
+                DeviceInfo.LightsDevice.TOLINO_VISION5 -> {
+                    logController("Tolino")
```
pazos, line 17: "TolinoNTX"

app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
```diff
@@ -11,7 +11,11 @@ object LightsFactory {
             return when (DeviceInfo.LIGHTS) {
                 DeviceInfo.LightsDevice.TOLINO_EPOS -> {
                     logController("Tolino")
```
pazos, line 13: "TolinoRoot"

hasezoey:
> And a final question because I got it wrong in the first place: with this PR in both the TestActivity's and KOReader's lightdialog work as expected?

no, with this PR the new driver works as expected, and is correctly set in the TestActivity, but not "globally", i (locally) added a fix in a local branch, but i will wait for the official solution you mentioned in https://github.com/koreader/android-luajit-launcher/pull/382#issuecomment-1248400785

my local solution currently for that is:
[DeviceInfo.kt.patch](https://github.com/koreader/android-luajit-launcher/files/9582898/DeviceInfo.kt.patch.txt)

app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
hasezoey: fixed with 93160a5ed4db0cdc178d0555f2601964b48d616b

app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
hasezoey: fixed with 93160a5ed4db0cdc178d0555f2601964b48d616b

pazos:
> > And a final question because I got it wrong in the first place: with this PR in both the TestActivity's and KOReader's lightdialog work as expected?
> 
> no, with this PR the new driver works as expected, and is correctly set in the TestActivity, but not "globally", i (locally) added a fix in a local branch, but i will wait for the official solution you mentioned in [#382 (comment)](https://github.com/koreader/android-luajit-launcher/pull/382#issuecomment-1248400785)
> 
> my local solution currently for that is: [DeviceInfo.kt.patch](https://github.com/koreader/android-luajit-launcher/files/9582898/DeviceInfo.kt.patch.txt)

Hmm. My solution isn't required. I got it wrong. There're no TOLINO on the Lights iterator, just TOLINO_EPOS and TOLINO_VISION, so the only way the factory is assigning you a bad driver is because your device is flagged as an EPOS before it is flagged as a VISION. 

You can test that by removing TOLINO_EPOS from the lightsMap.

If that is the case then we need to fine tweak the TOLINO_EPOS condition. I think it got reported somewhere, let me check.

hasezoey:
> You can test that by removing TOLINO_EPOS from the lightsMap.
> If that is the case then we need to fine tweak the TOLINO_EPOS condition. I think it got reported somewhere, let me check.

i think i wrote about this before, but i cannot find it anymore, so i will write it here again:
i had already tried this, and noticed that `HashMap` is always outputting a sorted iterator, so it does not matter if `TOLINO_EPOS` or `TOLINO_VISION5` gets added first / last, and i had also already tried to comment-out `TOLINO_EPOS`, which i noted to be working, so i just added (locally) a overwrite condition

pazos:
> > You can test that by removing TOLINO_EPOS from the lightsMap.
> > If that is the case then we need to fine tweak the TOLINO_EPOS condition. I think it got reported somewhere, let me check.
> 
> i think i wrote about this before, but i cannot find it anymore, so i will write it here again: i had already tried this, and noticed that `HashMap` is always outputting a sorted iterator, so it does not matter if `TOLINO_EPOS` or `TOLINO_VISION5` gets added first / last, and i had also already tried to comment-out `TOLINO_EPOS`, which i noted to be working, so i just added (locally) a overwrite condition and i was also hesitant to change existing conditions (based on comments from other PR's), so i did not and also dont know enough to change the implementation of the lights selection or how the project wants it to be done - so i just added a overwrite when `TOLINO_VISION5` is `true`

aaaahhh, now all makes sense.

I would rather prefeer to tweak Tolino epos to be true with current condition if not HARDWARE.contentEquals("e70k00").
If somehow that makes Tolino EPOS/EPOS2 no longer useable with previous driver that would mean they're also "e70k00" and, if there're no other way to differenciate them, we can always move them to the new driver and call it a day.

Apologies for the missunderstanding, I'm doing all this in coffee breaks and have little time to get things right ;)

hasezoey:
> I would rather prefeer to tweak Tolino epos to be true with current condition if not HARDWARE.contentEquals("e70k00").
> If somehow that makes Tolino EPOS/EPOS2 no longer useable with previous driver that would mean they're also "e70k00" and, if there're no other way to differenciate them, we can always move them to the new driver and call it a day.

if the tolino software download site is anything to go by, then:
- Epos2 and Vision5 (also Shine3 and Page2) are the same download package*<sup>1</sup>
- Epos1 is its own package
- Epos3 is its own package

*<sup>1</sup>: but as can be seen in https://github.com/koreader/android-luajit-launcher/issues/351#issue-1100647150, the Shine3 is listed as `Hardware: e60k00`

[Tolino Software Update Download Site](https://mytolino.de/software-updates-tolino-ereader/)

pazos:
Thanks for the info.

I'm afraid we can't figure out how things are without temporarily breaking (or not?) the lights controller on the EPOS/EPOS2. Even if we break them it is not that traumatic. On a couple of months somebody will report the regression paired with test results and we will be in a better position to judge :)

hasezoey:
> I'm afraid we can't figure out how things are without temporarily breaking (or not?)

if you *really* dont wanna break it, then i guess the following options would be possible:
- change from `HashMap` to something that uses FIFO (unsorted) or do a custom implementation
- use a overwrite (like i do locally)

pazos:
I stated that I wouldn't mind to break it. And prefeer to break it better than refactoring for a worse implementation or custom workarounds.

hasezoey:
> I stated that I wouldn't mind to break it

to me it sounded more like you are considering to break it, so i just wrote some "options" i could think of

---

when do you wanna do this change, in this PR or separately?

pazos:
> > I stated that I wouldn't mind to break it
> 
> to me it sounded more like you are considering to break it, so i just wrote some "options" i could think of
> 
> when do you wanna do this change, in this PR or separately?

In this PR is fine.

It should be a matter of doing something like

```diff
diff --git a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
index 8b8b9d2..ec54674 100644
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -386,6 +386,7 @@ object DeviceInfo {
         TOLINO_EPOS = BRAND.contentEquals("rakutenkobo")
             && MODEL.contentEquals("tolino")
             && DEVICE.contentEquals("ntx_6sl")
+            && !HARDWARE.contentEquals("e70k00")
```

Please check it works fine that way. The rest of the PR is ready to go.

hasezoey:
> Please check it works fine that way. The rest of the PR is ready to go.

added with 2ea428e92b3ca2bcf6b816aebede4df3e0355c4f and also tested to work in non-TestActivity

personally, i also think this will not affect EPOS1 or EPOS2, because if the hardware numbering is anything to go by, it should be older though i dont know about EPOS3 (which is from what i know android 8.1)

NiLuJe:
> i think it is at least `e80k00` (which is from what i know android 8.1)

Nope, those are NTX board names ;).

e for EPDC, 80 for 8.0", and the rest probably has some deeper meaning, too ;).

hasezoey:
> e for EPD, 80 for 8.0", and the rest probably has some deeper meaning, too ;).

i guess makes sense, i didnt think it had much meaning but more like PCB version / hardware revision number (which probably would increase with generation)

pazos (approved):

Nepochal:
Hi,
like I mentioned in #351 this merge did not add ntx_io for non rooted tolino shine 3 devices. They use the hardware key e60k00 which was not considered in this change.

PR #383, opened 2022-09-15
export device properties

pazos:
Same intent as #381 but exporting them all.

It makes possible to have very specific branches to workaround broken aosp devices, like

```lua
if android.prop.is_tolino then
    if android.prop.hardware ~= "whatever" or android.prop.model ~= "device" then
        -- stuff for all tolino devices except whatever/device
    end
end
```

exports manufacturer, brand, device, model, product and hardware as strings and boolean values for `is_tolino` and `is_boyue` as both brands use some non AOSP keycodes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/383)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/LuaInterface.kt
+++ b/app/src/main/java/org/koreader/launcher/LuaInterface.kt
@@ -18,6 +18,7 @@ interface LuaInterface {
     fun extractAssets(): Boolean
     fun getBatteryLevel(): Int
     fun getClipboardText(): String
+    fun getDeviceProperties(): String
     fun getEinkConstants(): String
     fun getEinkPlatform(): String
     fun getExternalPath(): String
@@ -29,7 +30,6 @@ interface LuaInterface {
     fun getName(): String
     fun getNetworkInfo(): String
     fun getPlatformName(): String
-    fun getProduct(): String
     fun getScreenAvailableHeight(): Int
     fun getScreenAvailableWidth(): Int
     fun getScreenBrightness(): Int
--- a/app/src/main/java/org/koreader/launcher/MainActivity.kt
+++ b/app/src/main/java/org/koreader/launcher/MainActivity.kt
@@ -361,6 +361,10 @@ class MainActivity : NativeActivity(), LuaInterface,
         }
     }
 
+    override fun getDeviceProperties(): String {
+        return device.properties
+    }
+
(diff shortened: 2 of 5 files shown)


hasezoey (commented):
while you are at it, i would suggest changing the conditions for `TOLINO`, because the current conditions are specific to a model and (unlike the comment say) does not catch all

for example the Vision5 getprop for `ro.prop` is
```txt
[ro.product.board]: [EVK]
[ro.product.brand]: [RakutenKobo]
[ro.product.cpu.abi2]: [armeabi]
[ro.product.cpu.abi]: [armeabi-v7a]
[ro.product.device]: [ntx_6sl]
[ro.product.display]: [eink]
[ro.product.hardwareType]: [E70K00]
[ro.product.lightGestureEnabled]: [true]
[ro.product.locale.language]: [en]
[ro.product.locale.region]: [US]
[ro.product.manufacturer]: [Rakuten Kobo Inc.]
[ro.product.model]: [tolino]
[ro.product.name]: [ntx_6sl]
```

hasezoey:
> Not sure what your intention is here :/
> I won't plan to change that.

i though i might as well ask, because this would bring `TOLINO` and `TOLINO_EPOS` conditions (and coding style) in line with other things (like `BOYUE`)

as for updating the conditions on `TOLINO`, i though i also might as well ask to bring it in line with things like the `BOYUE` conditions and reduce the OR conditions because they will be similar

see
https://github.com/koreader/android-luajit-launcher/blob/58cb8377c73c19542a3f3b81149c69d9e2100587/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt#L378-L381
https://github.com/koreader/android-luajit-launcher/blob/58cb8377c73c19542a3f3b81149c69d9e2100587/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt#L383-L386

https://github.com/koreader/android-luajit-launcher/blob/58cb8377c73c19542a3f3b81149c69d9e2100587/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt#L174-L175

Frenzie (approved):
Annoying, but seems sensible enough.

NiLuJe:
Not breaking backward compatibility with brittle crap like this unfortunately trumps coding style ;).

PR #404, opened 2023-02-08
Add Boyue S62

RamKromberg:
Boyue Likebook LemonRead S62A

Just the basic place holder.

I've written a wip frontlight controller too. But, although it reads fine, to write the brightness it needs system write so I haven't tested or added it to the pull. Attached here for reference: [BoyueS62Controller.zip](https://github.com/koreader/android-luajit-launcher/files/10683215/BoyueS62Controller.zip)

References: 
1. PhoneStatusBar.java from SystemUI.apk (deodox and decompile with jadx) showing the frontlight logic: [PhoneStatusBar.zip](https://github.com/koreader/android-luajit-launcher/files/10683289/PhoneStatusBar.zip)
   It also mentions _/sys/class/backlight/rk28_bl_warm/brightness_ but permissions are required there as well.
2. RK3368Impl.java from Boyue-Launcher.apk (deodox and decompile with jadx) showing mostly the EPD entries (EPD_FULL_DITHER isn't the same but we don't use it anyhow): [RK3368Impl.zip](https://github.com/koreader/android-luajit-launcher/files/10683244/RK3368Impl.zip)
3. FunctionUtil.java from Boyue-Launcher.apk (deodox and decompile with jadx) showing an alternative method to modify the brightness (that's also requires root): [FunctionUtil.zip](https://github.com/koreader/android-luajit-launcher/files/10683380/FunctionUtil.zip)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/404)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -33,6 +33,7 @@ object DeviceInfo {
         BOYUE_P6,
         BOYUE_P61,
         BOYUE_P78,
+        BOYUE_S62,
         BOYUE_T61,
         BOYUE_T62,
         BOYUE_T65S,
@@ -144,6 +145,7 @@ object DeviceInfo {
     private val BOYUE_P6: Boolean
     private val BOYUE_P61: Boolean
     private val BOYUE_P78: Boolean
+    private val BOYUE_S62: Boolean
     private val BOYUE_T61: Boolean
     private val BOYUE_T62: Boolean
     private val BOYUE_T65S: Boolean
@@ -225,6 +227,9 @@ object DeviceInfo {
 
         // Boyue Likebook P78
         BOYUE_P78 = BOYUE && PRODUCT.contentEquals("p78")
+
+        // Boyue Likebook LemonRead S62A
+        BOYUE_S62 = BOYUE && PRODUCT.contentEquals("s62")
 
         // Boyue T61
         BOYUE_T61 = (BOYUE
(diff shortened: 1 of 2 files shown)


pazos:
Thanks for the contribution! Looks good.

> I've written a wip frontlight controller too. But, although it reads fine, to write the brightness it needs system write so I haven't tested or added it to the pull.

It also looks good.

Write settings permission check/prompt isn't wired into the frontlight widget yet. Mostly because no driver was using it. Now it sounds like a good idea to implement.

If you want to test your driver now you can grant the permission going to "settings->screen->timeout"  and requesting custom timeouts.

If everything works it should be a matter of implementing a similar logic (maybe without that warning message) on the frontlight widget.

Feel free to open a separate ticket in this repo to keep track of it, in case it takes time, so we can merge this and move conversation there

pazos (approved):

RamKromberg:
@pazos regrettably it doesn't work. Basically, I get _"Lights  : java.lang.IllegalArgumentException: You cannot keep your settings in the secure settings."_ when trying to toggle the lights regardless of the api. Closest I got is to update SCREEN_BRIGHTNESS after the user manually turns the light on but toggles the manual mode (so the light turn off). Then, it's possible to modify SCREEN_BRIGHTNESS without a permission error. But, to actually turn the light on to that value, the user needs to toggle manual mode again... So, useless.

Maybe there's a way to workaround the issue like tricking systemui to redraw or something that I'm missing... But the difference between tap-tap-slide-tap (Frontlight in koreader) and slide-slide-tap-tap-tap/slide-tap is just not worth the fragility unless it's an officially supported API. So, I'm giving up here.

regards

pazos:
Indeed, secure settings are reserved for system apps. Maybe there's an intent or some other workaround but I agree with you it doesn't worth out time!

Thanks again!

RamKromberg:
It's quite annoying too since the apis for [SCREEN_BRIGHTNESS](https://developer.android.com/reference/android/provider/Settings.System#SCREEN_BRIGHTNESS) and [SCREEN_OFF_TIMEOUT](https://developer.android.com/reference/android/provider/Settings.System#SCREEN_OFF_TIMEOUT) shouldn't be treated differently by default so boyue/rockchip/someone clearly changed it deliberately for some reason.

Anyhow, since [Boyue went out of business](https://goodereader.com/blog/electronic-readers/boyue-is-going-out-of-business-and-the-likebook-brand-might-be-dead) and this particular model didn't even come with an app store, there's no disadvantage to modifying the system partition. So, I've already replaced the default webview with a new one (had to modify framework-res.apk) and I'm looking up how to modify permissions so I'll be able to put SCREEN_BRIGHTNESS back under the system write permission and get the generic controller to work...


Anyhow, thanks for the review and the trick about the timeout. The latter might help with getting the generic brightness controller up and running once I figure out the permissions thing.

PR #408, opened 2023-02-28
Add Boyue S62 root light controller

RamKromberg:
requires root.

Notes:
1. This should also work on the [Boyue Likebook Mimas (T103D)](https://www.mobileread.com/forums/showpost.php?p=3890090&postcount=583) and possibly a few other models: https://www.mobileread.com/forums/showthread.php?t=311910&page=39

2. Technically, *boyue_warm_light* is accessible to the shell user (via adb) so it's possible to use shizuku for rootless frontlight. However, since this model is an Android 6 firmware, it will be highly inconvenient as the the user will have run "adb shell sh /sdcard/Android/data/moe.shizuku.privileged.api/start.sh" via the pc with every reboot.
Still, it's something to consider if later Haoqingtech (formerly boyue) models (Meebook P10 Pro, Meebook P78 pro and MeeBook M6 so far...) have similar issues as they're Android 11 based and can loopback adb via the wireless debugging and trust agent without rebooting as [covered in the user guide](https://shizuku.rikka.app/guide/setup/).

3. I commented out the use of settings and used wrote directly to the socket file instead since doing it "properly" introduced over ~500ms. Similar models will probably want to experiment and see if the issue was fixed in their firmwares.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/408)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/lights/BoyueS62RootController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/lights/BoyueS62RootController.kt
@@ -1,0 +1,87 @@
+package org.koreader.launcher.device.lights
+
+import android.app.Activity
+import android.provider.Settings
+import android.util.Log
+import org.koreader.launcher.device.LightsInterface
+
+/* handle frontlight within the activity, without affecting other activities */
+
+class BoyueS62RootController : LightsInterface {
+    companion object {
+        private const val TAG = "Lights"
+        private const val BRIGHTNESS_MAX = 254
+        private const val BRIGHTNESS_MIN = 0
+    }
+
+    override fun getPlatform(): String {
+        return "boyue-s62-root"
+    }
+
+    override fun hasFallback(): Boolean {
+        return true
+    }
+
+    override fun hasWarmth(): Boolean {
+        return false
+    }
+
+    override fun needsPermission(): Boolean {
+        return false
+    }
+
+    override fun enableFrontlightSwitch(activity: Activity): Int {
+        return 1
+    }
+
+    override fun getBrightness(activity: Activity): Int {
+        var brightness = 0
+        activity.runOnUiThread {
+            try {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/koreader/launcher/device/lights/BoyueS62RootController.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    override fun setBrightness(activity: Activity, brightness: Int) {
+        if (brightness < BRIGHTNESS_MIN || brightness > BRIGHTNESS_MAX) {
+            Log.w(TAG, "brightness value out of range: $brightness")
+            return
+        }
+        Log.v(TAG, "setBrightness: $brightness")
+        activity.runOnUiThread {
+            try {
+                //Runtime.getRuntime().exec("su -c settings put system boyue_warm_light " + brightness) // slow
```
Frenzie, line 63: Is it still slower with something like this (or whatever the right function call was)?

```
Settings.System.putInt(activity.applicationContext.contentResolver, "boyue_warm_light", brightness)
```

RamKromberg: I get permission errors in logcat trying to putInt into that variable even after getting android.permission.WRITE_SETTINGS through the timeout menu.
It why root is required.

Frenzie: I'd also put that in the comment for future readers.

RamKromberg: There's just too many missing or inaccessible by incorrect permissions variables and out-right broken APIs on this device for me to start listing them in the source. I've mentioned some of the issues in the previous EDP pull request: https://github.com/koreader/android-luajit-launcher/pull/404

pazos (approved):

PR #415, opened 2023-05-10
android 13 cutout

yparitcher:
works for me on 13

Please test on <= 12

Fixes: koreader/koreader#9446

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/415)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/MainActivity.kt
+++ b/app/src/main/java/org/koreader/launcher/MainActivity.kt
@@ -44,8 +44,9 @@ class MainActivity : NativeActivity(), LuaInterface,
     // Path of last file imported
     private var lastImportedPath: String? = null
 
-    // Device cutout - only used on API 28+
-    private var topInsetHeight: Int = 0
+    // Surface height & width determined at runtime to account for device cutout
+    private var surfaceHeight: Int? = null
+    private var surfaceWidth: Int? = null
 
     // Fullscreen - only used on API levels 16-18
     private var fullscreen: Boolean = true
@@ -163,24 +164,10 @@ class MainActivity : NativeActivity(), LuaInterface,
             "surface changed {\n  width:  %d\n  height: %d\n format: %s\n}",
             width, height, pixelFormatName(format))
         )
+        surfaceWidth = width
+        surfaceHeight = height
         super.surfaceChanged(holder, format, width, height)
         drawSplashScreen(holder)
-    }
-
-    override fun onAttachedToWindow() {
-        Log.d(TAG_SURFACE, "onAttachedToWindow()")
-        super.onAttachedToWindow()
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-            val cut: DisplayCutout? = window.decorView.rootWindowInsets.displayCutout
-            if (cut != null) {
-                val cutPixels = cut.safeInsetTop
-                if (topInsetHeight != cutPixels) {
-                    Log.v(TAG_SURFACE,
-                        "top $cutPixels pixels are not available, reason: window inset")
-                    topInsetHeight = cutPixels
-                }
-            }
-        }
     }
 
     /* Called just before the activity is resumed by an intent */
@@ -436,18 +423,7 @@ class MainActivity : NativeActivity(), LuaInterface,
     }
 
     override fun getScreenHeight(): Int {
-        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-            // We need to handle the notch in Portrait
-            // NOTE: getScreenAvailableHeight does it automatically, but it also excludes the nav bar, when there's one :/
-            if (getOrientationCompat(screenIsLandscape).and(1) == 0) {
-                // getScreenOrientation returns LinuxFB rotation constants, Portrait rotations are always even
-                getHeight() - topInsetHeight
-            } else {
-                getHeight()
-            }
-        } else {
-            getHeight()
-        }
+        return surfaceHeight ?: getHeight()
     }
 
     override fun getScreenMaxBrightness(): Int {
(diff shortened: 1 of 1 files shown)


pazos:
Excellent, nicely done BTW :)

> Please test on <= 12

It should work just fine. Unfortunately I don't have my dev machine with me until the weekend. Could you upload the APK somewhere?

> works for me on 13

And should work for everybody :partying_face: 

Could you test what happens on devices that require a view?

https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/MainActivity.kt#L115-L123

You just need to flip the conditions. IIRC it should work (ie: the NativeSurfaceView's surfaceChanged shouldn't override the activity's surfaceChanged) but I'm not sure anymore :p 

Anyhow, none of the devices with NativeSurfaceView have a camera hole, so we're good nevertheless.

pazos (approved):
lgtm

NiLuJe:
I should be able to build/test tomorrow, FWIW ;).

yparitcher:
> Excellent, nicely done BTW :)

I just followed your instructions :)

https://user.fm/files/v2-ee3f4dfb7fe9f25004ddbcf525818773/koreader-android-arm-debug-v2023.04-26-gb4f453eb5_2023-05-07.apk

> Could you test what happens on devices that require a view?
> 
> https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/MainActivity.kt#L115-L123
> 
> You just need to flip the conditions. IIRC it should work (ie: the NativeSurfaceView's surfaceChanged shouldn't override the activity's surfaceChanged) but I'm not sure anymore :p
> 
> Anyhow, none of the devices with NativeSurfaceView have a camera hole, so we're good nevertheless.

sorry, not sure exactly what you want here.

yparitcher:
Please also test landscape, to make sure i did not mess that up.

pazos:
> sorry, not sure exactly what you want here.

1. Flip the condition by adding a `!`, so it ends like

```kotlin
 val surfaceKind: String = if (!device.needsView) {
            view = NativeSurfaceView(this)
            window.takeSurface(null)
            view?.holder?.addCallback(this)
            setContentView(view)
            "SurfaceView"
        } else {
            "Native Content"
        }
```

so your device ends using "SurfaceView" instead of "Native Content".

2. Rebuild the APK and test if you get the same results with both.

Again, not so important as there're no devices that require a view and have a cutout. So totally fine if you want to skip it. I can test it myself when I finally update my tablet to android 13

yparitcher:
the cutout on SurfaceView works, however when i try to open the top menu i get an error:
```
05-10 21:02:04.064 18158 18190 E NativeThread: Failed to run script: ffi/util.lua:693: bad argument #1 to 'pairs' (table expected, got nil)
05-10 21:02:04.109 18158 18158 V NativeGlue: Pause: 0xe7beb0f0
05-10 21:02:04.162  1203  1488 W ActivityManager: setHasOverlayUi called on unknown pid: 18158
```

yparitcher:
aarch64 refuses to build `jni/luajit/luajit`

NiLuJe:
> aarch64 refuses to build `jni/luajit/luajit`

It's a GCC issue on *debug* builds, a release one should go through just fin (but will crash in fun and interesting ways at runtime ;p).

NiLuJe:
> the cutout on SurfaceView works, however when i try to open the top menu i get an error:
> 
> ```
> 05-10 21:02:04.064 18158 18190 E NativeThread: Failed to run script: ffi/util.lua:693: bad argument #1 to 'pairs' (table expected, got nil)
> 05-10 21:02:04.109 18158 18158 V NativeGlue: Pause: 0xe7beb0f0
> 05-10 21:02:04.162  1203  1488 W ActivityManager: setHasOverlayUi called on unknown pid: 18158
> ```

Do we get a Lua stacktrace somewhere in these cases?

yparitcher:
> > aarch64 refuses to build `jni/luajit/luajit`
> 
> It's a GCC issue on *debug* builds, a release one should go through just fine (but will crash in fun and interesting ways at runtime ;p).

I thought we have a aarch64 build?

yparitcher:
> > the cutout on SurfaceView works, however when i try to open the top menu i get an error:
> > 
> > ```
> > 05-10 21:02:04.064 18158 18190 E NativeThread: Failed to run script: ffi/util.lua:693: bad argument #1 to 'pairs' (table expected, got nil)
> > 05-10 21:02:04.109 18158 18158 V NativeGlue: Pause: 0xe7beb0f0
> > 05-10 21:02:04.162  1203  1488 W ActivityManager: setHasOverlayUi called on unknown pid: 18158
> > ```
> 
> Do we get a Lua stacktrace somewhere in these cases?

That is all i got in the adb logs. The koreader bomb page is empty. This was using surfaceview instead of native context. Native context works fine.

NiLuJe:
Okay, finally tested on Android 12 ;).

First boot was... wonky on several aspects (some of it unrelated: e.g., the DB migration hung until the next input event), at which point it booted up... with the bottom cutoff ;).

Everything's fine after a restart, and Landscape behaves (as does switching from a Landscape book to the Portrait FM).

NiLuJe:
The urfaceView implementation also appears to behave... except for that first boot issue where it's also cutoff.

Funnily enough, I *couldn't* reproduce it the second time around when switching back to the Native Content imp.

(And I got to see the spinner on the splash screen, which I never did before, I think?).

Frenzie:
You should generally see a spinner (as in 3 dots) the first time after any upgrade. It's about 2-3 seconds on my Realme 7 Pro with a Snapdragon 720G.

NiLuJe:
Yeah, I'm not sure I did in the cases where the first boot was buggy (and the spinner is low, but I don't think it's low enough to fall off-screen in case the splash is affected by a similar issue?).

That said, it was late, I hadn't slept much, so it's entirely possible I just missed it ;).

pazos:
> the cutout on SurfaceView works, however when i try to open the top menu i get an error:

sorry, I missed the notification :p

Who knows? My tablet is going to update to lineage20 soon, so I can try to reproduce there. It seems that android 13 introduced quite a bit of changes that mess with some of our hacks. Work for future us :p

Feel free to merge when you want. @yparitcher

PR #418, opened 2023-06-24
simplify `FileExtensions.uncompress` implementation

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/418)


Frenzie (approved):
Fine by me though I don't know that I necessarily agree with the description.

benoit-pierre:
How would you describe it?

Frenzie:
"Use Apache IOUtils in FileExtensions.uncompress" :-)

It's a simplification of sorts if you're doing this multiple times but here it's just a minor abstraction for the same thing. But no matter.

pazos:
It's easier to read, so simplification works for me :)

pazos (approved):

PR #419, opened 2023-06-24
update to use NDK 23c

benoit-pierre:
The API level is now encoded in the toolchain executable names, and can be used in place.

Additionally, rework `jni/luajit/mk-luajit.sh` so each architecture has its own build directory (no need to clean when switching architecture). Note: the patches are applied to each build directory, not the original source checkout.

Bump minimum SDK version to 18 in `build.gradle`: the NDK version 23c minimal supported version is 16, but our builds only really work on 18 and above.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/419)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,26 +25,10 @@ BUILD_TASK?=assemble$(ARCH)$(FLAVOR)
 LINT_TASK?=lint$(ARCH)$(FLAVOR)
 
 # find the path where the SDK is installed
-ifdef SDK
-        ANDROID_SDK_FULLPATH?=$(SDK)
-else
-        ifdef ANDROID_HOME
-                ANDROID_SDK_FULLPATH?=$(ANDROID_HOME)
-        else
-                ANDROID_SDK_FULLPATH?=$(shell realpath ../../../base/toolchain/android-sdk-linux)
-        endif
-endif
+ANDROID_SDK_FULLPATH ?= $(or $(ANDROID_HOME),$(ANDROID_SDK_ROOT),$(shell realpath ../../../base/toolchain/android-sdk-linux))
 
 # find the path where the NDK is installed
-ifdef NDK
-	ANDROID_NDK_FULLPATH?=$(NDK)
-else
-	ifdef ANDROID_NDK_HOME
-		ANDROID_NDK_FULLPATH?=$(ANDROID_NDK_HOME)
-	else
-		ANDROID_NDK_FULLPATH?=$(shell realpath ../../../base/toolchain/android-ndk-r15c)
-	endif
-endif
+ANDROID_NDK_FULLPATH ?= $(or $(ANDROID_NDK_HOME),$(ANDROID_NDK_ROOT),$(shell realpath ../../../base/toolchain/android-ndk-r23c))
 
 # override android:versionName="string"
 ifdef ANDROID_NAME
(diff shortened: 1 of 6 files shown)


jni/luajit/mk-luajit.sh
```diff
@@ -2,99 +2,144 @@
...
     armeabi-v7a)
         # Android/ARM, armeabi-v7a (ARMv7 VFP)
-        NDKABI=${NDKABI:-14}
-        check_NDK
-        TCVER=("${NDK}"/toolchains/arm-linux-androideabi-4.*)
-        NDKP=${TCVER[0]}/prebuilt/${HOST_ARCH}/bin/arm-linux-androideabi-
-        NDKF="--sysroot ${NDK}/platforms/android-${NDKABI}/arch-arm"
-        NDKARCH="-march=armv7-a -mfloat-abi=softfp -Wl,--fix-cortex-a8"
-        make -C luajit amalg install HOST_CC="gcc -m32" CFLAGS="-O2 -pipe" HOST_CFLAGS="-O2 -pipe -mtune=generic" LDFLAGS="" HOST_LDFLAGS="" TARGET_CFLAGS="${CFLAGS}" TARGET_LDFLAGS="${LDFLAGS}" TARGET_LIBS="${EXTRA_LIBS}" TARGET_SONAME="libluajit.so" INSTALL_SONAME="libluajit.so" CROSS="$NDKP" TARGET_FLAGS="${NDKF} ${NDKARCH}" TARGET_SYS=Linux DESTDIR="${DEST}" PREFIX=
+        makeenv[CROSS]+=" armv7a-linux-androideabi${NDKABI:-16}-"
```
pazos, line 85: any reason to use ndkabi 16 instead of 18?.

Makefile
```diff
@@ -25,26 +25,10 @@ BUILD_TASK?=assemble$(ARCH)$(FLAVOR)
...
-ifdef NDK
-	ANDROID_NDK_FULLPATH?=$(NDK)
-else
-	ifdef ANDROID_NDK_HOME
-		ANDROID_NDK_FULLPATH?=$(ANDROID_NDK_HOME)
-	else
-		ANDROID_NDK_FULLPATH?=$(shell realpath ../../../base/toolchain/android-ndk-r15c)
-	endif
-endif
+ANDROID_NDK_FULLPATH ?= $(or $(ANDROID_NDK_HOME),$(ANDROID_NDK_ROOT),$(shell realpath ../../../base/toolchain/android-ndk-r15c))
```
pazos, line 31: I guess `base/toolchain/android-ndk-r15c` needs to be updated to the path where r23 is in

pazos (commented):
Looks good at first glance

Makefile
benoit-pierre: Good catch, amended.

jni/luajit/mk-luajit.sh
benoit-pierre: Nope, amended.

app/src/main/java/org/koreader/launcher/Assets.kt
```diff
@@ -18,6 +18,7 @@ class Assets {
     private val tag = this::class.java.simpleName
 
     init {
+        Log.i(tag, "loading lib7z")
```
pazos, line 21: For the sake of completeness please do the same with https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/device/Ioctl.kt#L9

pazos (approved):
lgtm

poire-z:
Can you merge it ? So @benoit-pierre can go on with his other PRs ?

benoit-pierre:
This would allow me to update the base PR, and check the CI builds correctly, even if we wait for a new point release before merging it.

benoit-pierre:
Wait, never mind, the CI build does create the APK, and luajit-launcher is a submodule of the main repo…

PR #439, opened 2023-09-24
Reverted cutout and fix rotations

hugleo:
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/439)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/MainActivity.kt
+++ b/app/src/main/java/org/koreader/launcher/MainActivity.kt
@@ -135,8 +135,6 @@ class MainActivity : NativeActivity(), LuaInterface,
         if (!hasMandatoryPermissions()) {
             requestMandatoryPermissions()
         }
-
-        device.epd.create()
     }
 
     /* Called when the activity has become visible. */
--- a/app/src/main/java/org/koreader/launcher/MainActivity.kt
+++ b/app/src/main/java/org/koreader/launcher/MainActivity.kt
@@ -46,7 +46,6 @@ class MainActivity : NativeActivity(), LuaInterface,
 
     // Device cutout - only used on API 28+
     private var insetsHeight: Int = 0
-    private var insetsWidth: Int = 0
 
     // Fullscreen - only used on API levels 16-18
     private var fullscreen: Boolean = true
--- a/app/src/main/java/org/koreader/launcher/MainActivity.kt
+++ b/app/src/main/java/org/koreader/launcher/MainActivity.kt
@@ -45,7 +45,7 @@ class MainActivity : NativeActivity(), LuaInterface,
     private var lastImportedPath: String? = null
 
     // Device cutout - only used on API 28+
-    private var insetsHeight: Int = 0
+    private var topInsetHeight: Int = 0
 
     // Fullscreen - only used on API levels 16-18
     private var fullscreen: Boolean = true
@@ -176,18 +176,21 @@ class MainActivity : NativeActivity(), LuaInterface,
         if (Build.VERSION.SDK_INT >= 33) {
             val cut = windowManager.defaultDisplay.cutout
             if (cut != null && cut.boundingRects.isNotEmpty()) {
-                insetsHeight = cut.safeInsetTop + cut.safeInsetBottom
-                Log.v(TAG_SURFACE,
-                    "edges $insetsHeight pixels are not available, reason: window inset")
+                val cutPixels = cut.safeInsetTop
+                if (topInsetHeight != cutPixels) {
+                    Log.v(TAG_SURFACE,
+                        "top $cutPixels pixels are not available, reason: window inset")
+                    topInsetHeight = cutPixels
+                }
             }
         } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
             val cut: DisplayCutout? = window.decorView.rootWindowInsets.displayCutout
             if (cut != null) {
                 val cutPixels = cut.safeInsetTop
-                if (insetsHeight != cutPixels) {
+                if (topInsetHeight != cutPixels) {
                     Log.v(TAG_SURFACE,
                         "top $cutPixels pixels are not available, reason: window inset")
-                    insetsHeight = cutPixels
+                    topInsetHeight = cutPixels
                 }
             }
         }
@@ -451,7 +454,7 @@ class MainActivity : NativeActivity(), LuaInterface,
             // NOTE: getScreenAvailableHeight does it automatically, but it also excludes the nav bar, when there's one :/
             if (getOrientationCompat(screenIsLandscape).and(1) == 0) {
                 // getScreenOrientation returns LinuxFB rotation constants, Portrait rotations are always even
-                getHeight() - insetsHeight
+                getHeight() - topInsetHeight
             } else {
                 getHeight()
             }
@@ -490,7 +493,7 @@ class MainActivity : NativeActivity(), LuaInterface,
             // NOTE: getScreenAvailableWidth does it automatically, but it also excludes the nav bar, when there's one :/
             if (getOrientationCompat(screenIsLandscape).and(1) == 1) {
                 // getScreenOrientation returns LinuxFB rotation constants, Landscape rotations are always odd
-                getWidth() - insetsHeight
+                getWidth() - topInsetHeight
             } else {
                 getWidth()
             }
--- a/app/src/main/java/org/koreader/launcher/MainActivity.kt
+++ b/app/src/main/java/org/koreader/launcher/MainActivity.kt
@@ -135,6 +135,8 @@ class MainActivity : NativeActivity(), LuaInterface,
         if (!hasMandatoryPermissions()) {
             requestMandatoryPermissions()
         }
+
+        device.epd.create()
     }
 
     /* Called when the activity has become visible. */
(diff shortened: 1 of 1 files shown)


hugleo:
Wondering why the changes for Android 14 don't work for me after I merged the changes with the checks and then I went to logging.

09-24 09:50:34.558 20926 20926 V Android_Version: onAttachedToWindow Android version is: 33
09-24 09:50:34.595 20926 20926 V Android_Version: surfaceChanged Android version is: 33
09-24 09:50:34.649 20926 20959 V Android_Version: getScreenWidth Android version is: 33
09-24 09:50:34.649 20926 20959 V Android_Version: getScreenHeight Android version is: 33

It is because my device is Android 13. DUH, DUH, DUH. I feel so dumb. That Xiaomi version screen really tricks me out."

Screenshot_2023-09-24-09-57-03-291_com android settings


Now the discussion about Android 14 doesn't make any sense since I actually have an Android 13 device.

@pazos Should I then apply my working changes to Android 13 then since I simulate the problem now on a real device?

Oh boy....

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -165,12 +169,46 @@ class MainActivity : NativeActivity(), LuaInterface,
...
+        Log.d(TAG_SURFACE, "onAttachedToWindow()")
+        super.onAttachedToWindow()
+
+        // ONLY FOR ANDROID 13
+        if (android.os.Build.VERSION.SDK_INT == 33) {
+            // nothing to do
+        }
+        // ONLY FOR ANDROID 14
+        else if (android.os.Build.VERSION.SDK_INT == 34) {
+            val cut = windowManager.defaultDisplay.cutout
```
pazos, line 193: Does this chunk of code works for android 9 - 12?

It would be wonderful to have a single codepath for all versions.

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -165,12 +169,46 @@ class MainActivity : NativeActivity(), LuaInterface,
...
+            if (cut != null && cut.boundingRects.isNotEmpty()) {
+                // TODO: we can handle various kinds of cutouts: getSafeInsetLeft, getSafeInsetRight, getSafeInsetTop, getSafeInsetBottom
```
pazos, line 195: Not a fan of random TODO lists.

Do you have something in mind to do with those cutouts?

app/src/main/java/org/koreader/launcher/MainActivity.kt
hugleo: > Does this chunk of code works for android 9 - 12?
> 
> It would be wonderful to have a single codepath for all versions.

That's what I would like to show you!

for android <= 12 as you say so works.
val cut: DisplayCutout? = window.decorView.rootWindowInsets.displayCutout

for android 13 this seems to be the secret:
windowManager.defaultDisplay.cutout
Following the doc this method should work for >= P

For now I've tested on devices without the hole on some devices < 13 and works.
But I can test some more on the emulator enabling holes in the dev options.

Did you remember to try the method windowManager.defaultDisplay.cutout before?

app/src/main/java/org/koreader/launcher/MainActivity.kt
hugleo: > Not a fan of random TODO lists.
> 
> Do you have something in mind to do with those cutouts?

Yes. Today we only deal with topInsetHeight when the hole is at the top. 
On the emulator, we can enable holes in any direction: top, bottom, left, or right. 
When you guys changed surfaceWidth = width; surfaceHeight = height inside the surfaceChanged method, it enabled support for all kinds of holes. 
When I reverted the changes I realized that it was not implemented in the old version ;-)

app/src/main/java/org/koreader/launcher/MainActivity.kt
pazos: > Did you remember to try the method windowManager.defaultDisplay.cutout before?

I actually don't remember. I think I didn't try anything. Just read the documentation, find something useful, went with that and pushed an APK for others to test. Since the test were ok I never bothered again :)

So please go with that for everything.

app/src/main/java/org/koreader/launcher/MainActivity.kt
pazos: Ok, so remove the TODO and add a NOTE instead.
Somebody else that reads the code in the future might find that useful.

Not so with a TODO.

app/src/main/java/org/koreader/launcher/MainActivity.kt
hugleo: > > Did you remember to try the method windowManager.defaultDisplay.cutout before?
> 
> I actually don't remember. I think I didn't try anything. Just read the documentation, find something useful, went with that and pushed an APK for others to test. Since the test were ok I never bothered again :)
> 
> So please go with that for everything.

Tested on emulator and doesn't work for Android 9. Need Android >= 10

pazos: Ok, thanks!.

Then from 9 up to 12 the legacy method and from 13 onwards the new one.

hugleo: > Ok, thanks!.
> 
> Then from 9 up to 12 the legacy method and from 13 onwards the new one.

Already there ;-)

pazos:
@hugleo: sorry for the delay.

two nitpicks and this can be merged.

Please add a log statement in the new condition, too. Use exactly the same string, so logs provide the same feedback in all android versions.

Please fix https://app.codacy.com/gh/koreader/android-luajit-launcher/pullRequest?prid=12685762

hugleo:
@pazos nitpicks addressed

app/src/main/java/org/koreader/launcher/MainActivity.kt
```diff
@@ -165,12 +164,35 @@ class MainActivity : NativeActivity(), LuaInterface,
...
     }
 
+    override fun onAttachedToWindow() {
+        Log.d(TAG_SURFACE, "onAttachedToWindow()")
+        super.onAttachedToWindow()
+
+        if (Build.VERSION.SDK_INT >= 33) {
+            val cut = windowManager.defaultDisplay.cutout
+            if (cut != null && cut.boundingRects.isNotEmpty()) {
+                insetsHeight = cut.safeInsetTop + cut.safeInsetBottom
```
pazos, line 179: any reason to take in account safeInsetBottom?. Did you find a device (I mean *any* device outside an emulator with fake cutouts) where it is not 0?.

It's not a vetto. Just trying to figure out what's the best option:

a) ignore `cut.safeInsetBottom` on both branches.
b) count `cut.safeInsetBottom` on both branches.

Anyhow, I'm fine if you go with *b** but then please use the same log statement.

hugleo: > any reason to take in account safeInsetBottom?. Did you find a device (I mean _any_ device outside an emulator with fake cutouts) where it is not 0?.
> 
> It's not a vetto. Just trying to figure out what's the best option:
> 
> a) ignore `cut.safeInsetBottom` on both branches. b) count `cut.safeInsetBottom` on both branches.
> 
> Anyhow, I'm fine if you go with _b_* but then please use the same log statement.

No, see only in the emulator. Fine with a) since it makes more sense for the log as well.

hugleo: I think can get rid of cut.boundingRects.isNotEmpty() as well since I don't remember if it's needed and it would save some duplication

pazos (approved):
Looks much better, thanks!

jospalau:
This broke the screen for my Xiaomi. I don't read on the phone. Just letting you know. MIUI 14 too.

I am using the patch I used to use previously to the original fix that this is reverting to add the missing height.

hugleo:
Do you have a screenshot of the problem?

jospalau:
Sure. 

Screenshot_2024-02-14-22-11-36-977_org koreader launcher

hugleo:
If you rotate to landscape in KOReader options and then back to portrait, does the problem go away?

hugleo:
Also tell which device model

hugleo:
This device has a hole at the top just like mine. The black bar is supposed to fill the hole and should be drawn at the top. However, on yours, it is being drawn at the bottom. Unless, of course, you have a rotation configured like the picture below, this should be the normal behavior.

rotation

jospalau:
I have it by default but I don't know why the order of the icons is different than in yours.
IMG_20240214_224916

I think I know what you mean but I want it full screen like the rests of the apps.

Previous code reverted and the patch makes it work:
local screen = require("android").screen
screen.height = screen.height +90

hugleo:
> I have it by default but I don't know why the order of the icons is different than in yours. IMG_20240214_224916
> 
> I think I know what you mean but I want it full screen like the rests of the apps.
> 
> Previous code reverted and the patch makes it work: local screen = require("android").screen screen.height = screen.height +90

Your default setting is correct; in my case, I intentionally changed it. Your black bar should be at the top for this rotation mode.

jospalau:
It is not a problem to be honest. Just curious.
Thanks

PR #441, opened 2023-10-07
fix native libraries search path

benoit-pierre:
Cf. https://github.com/koreader/koreader/issues/10533.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/441)

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -2654,9 +2654,10 @@ local function run(android_app_state)
 
     -- load the dlopen() implementation
     android.dl = require("dl")
+    android.dl.system_libdir = ffi.abi("64bit") and "/system/lib64" or "/system/lib"
     android.dl.library_path = android.nativeLibraryDir..":"..
         android.dir..":"..android.dir.."/libs:"..
-        "/lib:/system/lib:/lib/lib?.so:/system/lib/lib?.so"
+        string.gsub("@:@/lib?.so", "@", android.dl.system_libdir)
 
     -- register the dependency lib loader
     table.insert(package.loaders, 3, android.deplib_loader)
--- a/assets/dl.lua
+++ b/assets/dl.lua
@@ -54,6 +54,9 @@ end
 local dl = {
     -- set this to search in certain directories
     library_path = '/lib/?;/usr/lib/?;/usr/local/lib/?',
+    -- set this to the directory of system libraries
+    -- (to be ignored when loading dependencies).
+    system_libdir = nil,
 }
 
 local function sys_dlopen(library, global, padding)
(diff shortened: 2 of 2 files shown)


Frenzie (approved):
lgtm in principle, but are you sure they have to be excluded?

assets/android.lua
```diff
@@ -2654,9 +2654,10 @@ local function run(android_app_state)
 
     -- load the dlopen() implementation
     android.dl = require("dl")
+    android.dl.system_libdir = ffi.abi("64bit") and "/system/lib64" or "/system/lib"
     android.dl.library_path = android.nativeLibraryDir..":"..
         android.dir..":"..android.dir.."/libs:"..
-        "/lib:/system/lib:/lib/lib?.so:/system/lib/lib?.so"
+        string.gsub("@:@/lib?.so", "@", android.dl.system_libdir)
```
NiLuJe, line 2660: I don't recall why the `/lib` & `/lib/lib` stuff was in there in the first place, but is it safe to drop those?

(For context, it might have been for broken Chinese ROMs or something).

NiLuJe: (FWIW, I'm in the "fuck it" train if that was the case, just wanted to know if you had thoughts on that ;)).

NiLuJe:
> but are you sure they have to be excluded?

Yep, see the comments, the dynamic loader will bounce you if you try ;).

hugleo:
Here the /lib & /lib/lib commit: https://github.com/koreader/android-luajit-launcher/commit/64caed00259bb23b789f2a8051e77efa9bcde9cc

benoit-pierre:
> Here the /lib/lib commit: [64caed0](https://github.com/koreader/android-luajit-launcher/commit/64caed00259bb23b789f2a8051e77efa9bcde9cc)
> 
> An here the lib commit [9c95c44](https://github.com/koreader/android-luajit-launcher/commit/9c95c44cdfe462dbe5f259daf7867cb2fa1719f5)

Going back further: 7d0c1fcc2a1811af43ac42b8db76ca1c2268fb06

So `/lib` was [already there](https://github.com/koreader/android-luajit-launcher/blob/7d0c1fcc2a1811af43ac42b8db76ca1c2268fb06/jni/android-main.c#L45), no idea why. The canonical directory is [`/system/lib` on 32bits and `/system/lib64` for 64bits](https://source.android.com/docs/core/permissions/namespaces_libraries), with the possible addition of `/vendor/lib` / `/vendor/lib64`.

I checked with the oldest available emulator image for x86 (API level 10, Gingerbread), and `/system/lib` is already used, no `/lib`.

Without a clear example of "broken Chinese ROMs or something", I would stick to the default.

PR #445, opened 2023-10-26
update Gradle to 7.3.3

benoit-pierre:
So Java 17 can be used, the minimum required version to use the most recent Android SDK tools like `sdkmanager`.

The `koandroid` docker image will have to be updated to Focal (20.04), which is good idea anyway as the currently used version reached EOL in May (Bionic: 18.04).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/445)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,8 +3,8 @@ buildscript {
     ext.targetSdk = 30
     ext.minSdk = 18
 
-    ext.gradle_plugin_version = '4.2.1'
-    ext.kotlin_plugin_version = '1.5.20'
+    ext.gradle_plugin_version = '7.4.2'
+    ext.kotlin_plugin_version = '1.6.21'
     ext.androidx_core_version = '1.6.0'
     ext.androidx_appcompat_version = '1.3.1'
     ext.androidx_supportv4_version = '1.0.0'
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists


Frenzie:
We're not willfully on 18.04 ftr, but it fails (hopefully now failed) in CI with 20.04 or 22.04 while the exact same Docker image worked fine locally.
https://github.com/koreader/virdevenv/commit/9052cee45b019180d787d61b6d56b5428f847590

Frenzie (approved):

pazos:
> So Java 17 can be used, the minimum required version to use the most recent Android SDK tools like sdkmanager.

I would prefeer java 11 (the minimum required version nowadays) and keep the sdk as is for a while.

But that's because I have no real idea of the benefits of a new sdk :smile: 

In both cases the docker image needs to be updated and this PR is ok as is 

Also works fine with slightly newer versions:

```diff
diff --git a/build.gradle b/build.gradle
index 213823c..0ff5834 100644
--- a/build.gradle
+++ b/build.gradle
@@ -3,8 +3,8 @@ buildscript {
     ext.targetSdk = 30
     ext.minSdk = 18

-    ext.gradle_plugin_version = '4.2.1'
-    ext.kotlin_plugin_version = '1.5.20'
+    ext.gradle_plugin_version = '7.4.2'
+    ext.kotlin_plugin_version = '1.6.21'
     ext.androidx_core_version = '1.6.0'
     ext.androidx_appcompat_version = '1.3.1'
     ext.androidx_supportv4_version = '1.0.0'
diff --git a/gradle/wrapper/gradle-wrapper.properties b/gradle/wrapper/gradle-wrapper.properties
index 4d9ca16..8049c68 100644
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
```

benoit-pierre:
> I would prefeer java 11 (the minimum required version nowadays) and keep the sdk as is for a while.

The SDK is unchanged, it's just the commandline tools ( apkanalyzer, avdmanager, sdkmanager).

> But that's because I have no real idea of the benefits of a new sdk 😄

It's annoying to have to keep an old version of Java and the those tools just for koreader (independently of the NDK / SDK versions). 

Additionally, JDK 17 is available on Ubuntu 20.04 LTS too (so it's not that fresh).

> In both cases the docker image needs to be updated and this PR is ok as is

See https://github.com/koreader/virdevenv/pull/93.

> Also works fine with slightly newer versions:

Yep, thank, applied.

PR #472, opened 2024-03-10
Add Linfiny eNote / Avalue ENT-13T1 / QuirkLogic Papyr

davide125:
These are all OEM variations of the same hardware, which itself originates from the Sony DPT-RP1 design that was licensed to Linfiny.

From https://github.com/koreader/koreader/issues/8482#issuecomment-1986743868
> Commercial name of the product: QuirkLogic Papyr
> Android version: 5.1
> Device has no lights
> Freescale/NXT is the only eink driver that works
> Report:
> [test.log](https://github.com/koreader/koreader/files/14545151/test.log)

I tested this on my QuirkLogic Papyr and it seems to work fine.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/472)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -49,6 +49,7 @@ object DeviceInfo {
         HANVON_960,
         INKBOOK,
         JDREAD,
+        LINFINY_ENOTE,
         MEEBOOK_P6,
         NABUK,
         NOOK,
@@ -148,6 +149,7 @@ object DeviceInfo {
     enum class QuirkDevice {
         NONE,
         EMULATOR,
+        LINFINY_ENOTE,
         ONYX_MAX,
         ONYX_NOTE,
         ONYX_POKE2,
@@ -186,6 +188,7 @@ object DeviceInfo {
     private val HANVON_960: Boolean
     private val INKBOOK: Boolean
     private val JDREAD: Boolean
+    private val LINFINY_ENOTE: Boolean
     private val MEEBOOK_P6: Boolean
     private val NABUK_REGAL_HD: Boolean
     private val NOOK: Boolean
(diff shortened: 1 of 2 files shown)


davide125:
Report with this PR applied: [test.log](https://github.com/koreader/android-luajit-launcher/files/14548863/test.log)

I noticed is says "Lights: generic" despite having set the quirk for no lights, not sure if that's normal or not.

pazos:
> Report with this PR applied: [test.log](https://github.com/koreader/android-luajit-launcher/files/14548863/test.log)
> 
> I noticed is says "Lights: generic" despite having set the quirk for no lights, not sure if that's normal or not.

Yup, the LightsFactory always returns a LightsController. If no match happens the else branch returns a GenericController in https://github.com/koreader/android-luajit-launcher/blob/master/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt#L82-L85

What the `QUIRK_NO_LIGHTS` does is tell the frontend there're no lights, so the program doesn't add an action to the relevant settings in KO.

Given that the setting was never tested by any maintainer I don't recall if gestures are affected (ie: skipped) when a device declares no lights. I think you can test that doing the regular gesture (bright changes by vertical scrolling the left border of the screen).

On a device with lights, like a phone, you'll see it changes the screen brightness and a notification pops up. No idea what happens on your device with no lights.

pazos (approved):
LGTM, thanks!

app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
```diff
@@ -185,6 +187,7 @@ object DeviceInfo {
     private val FIDIBOOK: Boolean
     private val HANVON_960: Boolean
     private val INKBOOK: Boolean
+    private val LINFINY_ENOTE: Boolean
```
pazos, line 190: minor nitpick: sorting

davide125: Fixed, thanks!

davide125:
> Given that the setting was never tested by any maintainer I don't recall if gestures are affected (ie: skipped) when a device declares no lights. I think you can test that doing the regular gesture (bright changes by vertical scrolling the left border of the screen).
> 
> On a device with lights, like a phone, you'll see it changes the screen brightness and a notification pops up. No idea what happens on your device with no lights.

I wasn't able to trigger the gesture on my device, so I assume it's being skipped. fwiw there _is_ a brightness slider in the device Settings app under Display, defaulting to around 25%, but it doesn't seem to do anything at all when I change it, so I assume the vendor was just lazy and left it around.

PR #474, opened 2024-04-05
Add support for C64P

ovehbe:
commercial name of product: C64P (Boyue P6 Knockoff) android version: 8.1
driver(s) that work;
Rockchip RK3368
Rockchip RK3026

Other:
None of the light drivers work in the compatibility checker but light works in the app. My device only works in A2 mode.
*Tolino/NTX driver doesnt work.

Device info:
Manufacturer: c64p
Brand: c64p
Model: c64p
Device: c64p
Product: c64p
Hardware: rk30board
Platform: rk3326

This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/474)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -101,7 +101,6 @@ object DeviceInfo {
 
     enum class LightsDevice {
         NONE,
-        BOYUE_C64P,
         BOYUE_S62,
         CREMA_CARTA_G,
         MEEBOOK_P6,
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -41,6 +41,7 @@ object DeviceInfo {
         BOYUE_T80D,
         BOYUE_T80S,
         BOYUE_T103D,
+        C64P,
         CREMA,
         CREMA_0650L,
         CREMA_CARTA_G,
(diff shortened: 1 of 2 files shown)


Frenzie (approved):

pazos (changes requested):
thanks for the patch, @ovehbe!

 A few changes required:

- A better name for the device
- Remove the device from LightsDevice, since you're not using it.

You can get the idea from https://github.com/koreader/android-luajit-launcher/pull/475, which is exactly yours with these little changes :)

app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
```diff
@@ -100,6 +101,7 @@ object DeviceInfo {
 
     enum class LightsDevice {
         NONE,
+        BOYUE_C64P,
```
hugleo, line 104: remove

app/src/main/java/org/koreader/launcher/device/EPDFactory.kt
```diff
@@ -18,6 +18,7 @@ object EPDFactory {
     val epdController: EPDInterface
         get() {
             return when (DeviceInfo.EINK) {
+                DeviceInfo.EinkDevice.BOYUE_C64P,
```
hugleo, line 21: I would prefer add to RK3368 driver since you got more recent device. My test apk was RK3368. But it's OK if both works. for you.

ovehbe: Got it, I'll modify accordingly.

hugleo (approved):

pazos (approved):
thanks! lgtm

Issue #477, opened 2024-04-14
Brightness and Warmth support for Xiaomi Duokan Pro II 


cheehongw:
Brightness/warmth controls currently don't work at all for this e-reader. 

Commercial Product Name: `Xiaomi Duokan Pro II` / `Xiaomi Ereader Pro II`
Model: `miduokanreaderproii`
Product: `moaan`
Device: `rk3566_eink`
Hardware: `rk30board`
Platform: `rk356x`
Android Version: 11

I have tried my hands at making a driver for it, but have encountered a roadblock, so this is me sharing my findings. 

The brightness/warmth can be controlled through the adb shell, but it appears that these values are located under `Settings.Global`, which means only system apps can set these values. If anyone has any tricks to set `Settings.Global`, please let me know!

```
rk3566_eink:/ $ settings list global | grep mogu                                                                                                                                                           
mogu_audio_status=1                       
mogu_audio_value=2        
mogu_cold_led_status=1                  //only applicable when backlight is on. From testing, this follows mogu_warm_led_status
mogu_cold_led_value=10                 //determines the warmth level
mogu_contrast_value=8
mogu_init_app_config=2
mogu_notification_status=1
mogu_notification_value=3
mogu_warm_led_status=1               //determines whether the backlight is on or off
mogu_warm_led_value=13              //determines the brightness level
mogu_wifi_status=0
```

I also investigated `/sys/class/backlight` but the files there are rather inconclusive and dont contain any real values.

```
rk3566_eink:/sys/class/backlight/aw99703 $ ls
actual_brightness  bl_power  brightness  device  max_brightness  power  subsystem  type  uevent
rk3566_eink:/sys/class/backlight/aw99703 $ cat actual_brightness                                                                                                                                           
0
rk3566_eink:/sys/class/backlight/aw99703 $ cat brightness                                                                                                                                                  
0
rk3566_eink:/sys/class/backlight/aw99703 $ cat bl_power                                                                                                                                                    
0
rk3566_eink:/sys/class/backlight/aw99703 $ cat max_brightness                                                                                                                                              
2047
```

pazos:
ADB is not a reliable method to figure out what unprivilegied apps can do.

You'll need to RE the binaries to find which methods are called by the stock app. If you look in open/closed tickets and previous implemented drivers you'll find a lot of info on how it can be done.

jkl16:
[...]

PR #478, opened 2024-04-23
EPD & Lights drivers for Xiaomi RK3566 based devices

jkl16:
- Added EPD driver implementing full refresh for Xiaomi 7 Reader. Refresh is using mode configured in System UI. 
- Added Lights & Warmth driver implementing _getting_ & _setting_ values, however _setting_ is not enabled since it requires root.

I have provided more details in https://github.com/koreader/android-luajit-launcher/issues/477

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/478)


jkl16:
Since complexity check is failing, should I split out EPD and light changes to a separate PRs?

Frenzie:
Nah, that check seems kind of dumb, at least in the context of this PR.

pazos:
> Added Lights & Warmth driver implementing getting & setting values, however setting is not enabled since it requires root.

I'm not very fan of this. It would be better to use an Intent to launch the native dialog used by the system, if something like this exists. Gestures won't work but at least the dialog would be useful to change levels.

Until now all drivers rely on our own widget, so we need a little change to make possible to call external widgets instead.

What do you think, @jkl16?

jkl16:
In the device that I own, brightness & warm levels can only be controlled from native Android notification panel. It looks as on the following photo:


The panel can be accessed when running KOReader at any time by pulling it down from the top edge.  From UX perspective that seems good enough, so maybe there is no need to implement that part at all.
The biggest issue is the fact that the notification panel gesture collides with KOReader top menubar.

Frenzie:
> From UX perspective that seems good enough, so maybe there is no need to implement that part at all.

Better not indeed.

jakkli-mend:
Then based on the discussion I removed all lights related code from the branch.

jkl16:
Then based on the discussion I removed all lights related code from the branch.

Frenzie (approved):

pazos (approved):
lgtm, thanks!

pazos:
> The biggest issue is the fact that the notification panel gesture collides with KOReader top menubar.

You can setup KO top menu to be activated only when tapping and use the swipe gesture to activate system menu.

PR #483, opened 2024-05-19
Support for a bunch of rockchip devices

hugleo:
Device info:
Manufacturer: haoqing
Brand: haoqing
Model: m6
Device: m6
Product: rk3566_eink
Hardware: rk30board
Platform: rk356x

Working EPD Driver: RK3566

Device info:
Manufacturer: haoqing
Brand: haoqing
Model: m7
Device: m7
Product: rk3566_eink
Hardware: rk30board
Platform: rk356x

Working EPD Driver: RK3566

Device info:
Manufacturer: rockchip
Brand: rockchip
Model: moaanmix7
Device: rk3566_eink
Product: moaan
Hardware: rk30board
Platform: rk356x

Working EPD Driver: RK3566

Device info:
Manufacturer: rockchip
Brand: rockchip
Model: inkpalmplus
Device: rk3566_eink
Product: moaan
Hardware: rk30board
Platform: rk356x

Working EPD Driver: RK3566

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/483)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -49,9 +49,13 @@ object DeviceInfo {
         FIDIBOOK,
         HANVON_960,
         INKBOOK,
+        INKPALM_PLUS,
         JDREAD,
         LINFINY_ENOTE,
+        MEEBOOK_M6,
+        MEEBOOK_M7,
         MEEBOOK_P6,
+        MOAAN_MIX7,
         NABUK,
         NOOK,
         NOOK_GL4,
@@ -191,9 +195,13 @@ object DeviceInfo {
     private val FIDIBOOK: Boolean
     private val HANVON_960: Boolean
     private val INKBOOK: Boolean
+    private val INKPALM_PLUS: Boolean
     private val JDREAD: Boolean
     private val LINFINY_ENOTE: Boolean
+    private val MEEBOOK_M6: Boolean
+    private val MEEBOOK_M7: Boolean
     private val MEEBOOK_P6: Boolean
+    private val MOAAN_MIX7: Boolean
     private val NABUK_REGAL_HD: Boolean
     private val NOOK: Boolean
     private val NOOK_GL4: Boolean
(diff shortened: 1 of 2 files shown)


Frenzie (approved):

Bluemoondragon07:
Are these changes in effect for recent nightlies of KOreader? The Meebook M7 is still registered as an unsupported device in the most recent nightly release.

benoit-pierre:
Yes, since v2024.04-206-gb7fc048bb.

hugleo:
Nothing was removed here. @Bluemoondragon07, can you run the KOReader eink test tool to confirm your device? 
Follow instructions here: https://github.com/koreader/koreader/issues/8482

Bluemoondragon07:
The Rockchip rk3566 works when I test it. For some reason, the device just doesn't register as supported. I will see if downloading the last stable release, 2024.07, fixes anything.

Bluemoondragon07:
Uninstalling and reinstalling 2024.07 didn't do the trick, unfortunately.

benoit-pierre:
Could it be some sort of trailing whitespace in some fields that would trip one or more of the string equality tests?

Bluemoondragon07:
So,  with the latest nightly, the Meebook M7 now shows up as supported, but the full screen refresh function does not work using gestures. Setting the page refresh in e-ink settings to refresh every page, nothing happens. 



So, still can't trigger refreshes yet, but it's showing up as supported now. 

Also, I tried in both regal/normal mode and A2/fast mode.

hugleo:
> So, with the latest nightly, the Meebook M7 now shows up as supported, but the full screen refresh function does not work using gestures. Setting the page refresh in e-ink settings to refresh every page, nothing happens.
> 
> 
> 
> So, still can't trigger refreshes yet, but it's showing up as supported now.
> 
> Also, I tried in both regal/normal mode and A2/fast mode.

in eink-test is Working EPD Driver RK3566 or another one?

Bluemoondragon07:
@hugleo Yes, in compatibility test Rockchip Rk3566  triggers a full refresh in all refresh modes.

Bluemoondragon07:
https://github.com/user-attachments/assets/3ee0dd09-30f7-4b33-848b-56ce66e604db

hugleo:
@pazos One of those hard cases that works on E Ink but not KOReader. Any advice?

Bluemoondragon07:
It might have something to do with KOreader not using typical Android elements? I can't get anything in KOreader, except in compatibility test, to trigger a refresh, even with native device settings. In most apps, such as Moon Reader, if I have the device refresh rate set to 1, then every tap on an Android element or app event triggered by a physical button press causes a refresh.

When I go into compatibility test, I know it must be using Android elements, because any button I press there will follow the device refresh rate settings. Of course, pressing the  Rk3566 test will make it refresh every time anyway regardless of device settings. But with that refresh-rate-of-1 setting, any of those buttons trigger a refresh if I tap on them. 

But as soon as I exit that compatibility test in KOreader, there is nothing within the app that can cause a refresh. I can only refresh it manually through the OS from there.

Frenzie:
> But with that refresh-rate-of-1 setting, any of those buttons trigger a refresh if I tap on them.

That's a very irrelevant test then. ;-) Try it with the system's reset setting to something significantly higher than 1.

Bluemoondragon07:
@Frenzie my tests have been with different refresh rates, I was just giving a specific example of how the UI in compatibility test reacts to device system refresh rate but the rest of the app does not. The RK3566 test always refreshes the screen, regardless of refresh rate or even refresh mode. It's just that I notice other buttons in the test can trigger system refreshes because they are registered as real Android buttons.

PR #486, opened 2024-06-06
speedup gradle

benoit-pierre:
- increase JVM heap space
- disable `lintVitalArmRocksRelease` on CI
- enable caching, parallelization and configuration on demand

Locally, starting with no `~/.gradle` directory and pristine `luajit-launcher` checkout, the time spent running `./gradlew …` is:

|                | no daemon & no cache | no daemon & cache | daemon & cache |
-----------------|---------------------:|------------------:|---------------:|
| master         |                2m52s |             1m27s |            48s |
| PR             |                2m56s |               16s |             7s |
| PR (`CI=true`) |                1m44s |               16s |            N/A |

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/486)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,3 @@
-# Prevent OOM in Java heap space
-org.gradle.jvmargs=-Xmx2048m
 org.gradle.warning.mode=all
 org.gradle.caching=true
 org.gradle.configuration-cache=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,10 @@
 # Prevent OOM in Java heap space
 org.gradle.jvmargs=-Xmx2048m
 org.gradle.warning.mode=all
+org.gradle.caching=true
+org.gradle.configuration-cache=true
+org.gradle.configureondemand=true
+org.gradle.parallel=true
 
 # Use AndroidX compat libraries
 android.useAndroidX=true
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,10 @@ NAME?=1.0
 VERSION?=1
 APPNAME?="luajit-launcher"
 
+ifneq (,$(CI))
+  GRADLE_FLAGS ?= --console=plain --no-daemon -x lintVitalArmRocksRelease
+endif
+
 update:
 	@echo "Updating sources"
 	git submodule init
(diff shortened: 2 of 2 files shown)


gradle.properties
```diff
@@ -1,6 +1,9 @@
 # Prevent OOM in Java heap space
-org.gradle.jvmargs=-Xmx2048m
+org.gradle.jvmargs=-Xmx3072m
```
Frenzie, line 2: But all the various CI solutions "only" supply 2 GB?

benoit-pierre: - [AppVeyor](https://www.appveyor.com/docs/build-environment/): 6-7 GB
- [CircleCI](https://circleci.com/product/features/resource-classes/): 2-8 GB (medium has 4 GB)
- [Github Actions](https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories): 7-16 GB
- [Gitlab](https://docs.gitlab.com/ee/ci/runners/hosted_runners/linux.html): 8-16 GB
- [Travis CI](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system): 4-8 GB

So unless we plan on building APKs on CircleCI with the smallest resource classes, I think it should be fine.

benoit-pierre: We could also override the value when running in CI.

Frenzie: Alright. I guess Gitlab upped it to 8. I think it was 4 a couple of years ago, possibly 2. Should we also set Xms=1500m or some such or does Xmx already reduce time spent dynamically allocating?

benoit-pierre: Re-testing only that change locally, I'm not seeing any difference… Since this was originally done some time ago, before the JVM/gradle update, maybe something changed. Or maybe it never had any real impact? I also tested without the setting: again, no impact.

So I dropped that change.

benoit-pierre:
It also occurred to me that `$PARALLEL_JOBS` should be honored, in case we end-up in a situation similar to CircleCI.

gradle.properties
Frenzie: I believe the default value is (or was) 512, which led to issues as per the comments. I'm not sure if the value has any meaning other than just being larger.

benoit-pierre: Default is apparently: `-Xms256m -Xmx512m`.

Frenzie: If the default is no longer an issue that's fine by me.

benoit-pierre: So remove it?

Frenzie: Yup! 👍

Frenzie: @pazos to double check, do you agree?

pazos: sorry, I miss this one.

Totally cool :+1:

pazos (approved):

PR #489, opened 2024-06-20
more ci

pazos:
To actually fix the remaining codacy warnings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/489)

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -2,7 +2,6 @@
 engines:
   duplication:
     exclude_paths:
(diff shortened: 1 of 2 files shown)


Frenzie:
That strikes me as less legible but I suppose it does make it easier to find several of the same without potential false positives.

pazos:
> That strikes me as less legible but I suppose it does make it easier to find several of the same without potential false positives.

Yay. I'm too lazy to setup detekt locally and tweak its config until I find a way to skip that particular file. And the check itself makes sense elsewhere.

PR #490, opened 2024-06-24
Update luajit to match koreader luajit version

hugleo:
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/490)


Frenzie:
Was updated in https://github.com/koreader/koreader-base/pull/1828

hugleo:
@benoit-pierre , @Frenzie  Can we bump android-luajit-launcher on koreader already?

pazos:
It is safe to bump

hugleo:
Mostly because of this: https://github.com/koreader/koreader/issues/11723#issuecomment-2083360146

benoit-pierre:
The `koreader/koandroid:0.8.1-20.04` docker image has everything needed, so we can bump.

Frenzie:
The next release in that context is 2024.04 which I tagged mere hours later. I didn't want to risk any delays in case something didn't work. I think the image should be fine but if not I'll fix it (or @benoit-pierre will, since he has this habit of being incredibly fast, or in any case available when I'm not).

Frenzie:
Also GH has this habit of not showing new replies immediately.

PR #495, opened 2024-07-11
Add support for Crema Grande

Ki-qu:
Commercial name of product: Crema Grande
Android Version: 4.4.2
Light: Tolino Ntx
Display: Freescale/NTX

Manufacturer: iwg
Brand: crema
Model: crema-0710c
Device: ntx_6sl
Product: ntx_6sl
Hardware: e60qn0
Platform: imx6

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/495)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -49,6 +49,7 @@ object DeviceInfo {
         BOYUE_T103D,
         CREMA,
         CREMA_0650L,
+        CREMA_0710C,
         CREMA_CARTA_G,
         ENERGY,
         FIDIBOOK,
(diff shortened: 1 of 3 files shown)


Ki-qu:
I found better solution that works with this device without artificial delay by reverse engineering stock apps: by writing a value to `/sys/class/graphics/fb0/full_refresh_request`.

- `1` will refresh the screen after screen is updated (This is what I'd like to use)
- `2` will immediately refresh the screen

This might be only relevant to my device. Can I implement with this method?

Frenzie:
Related, does that require root?

Ki-qu:
No. It doesn't require root.

app/src/main/java/org/koreader/launcher/device/epd/TolinoAltEPDController.kt
```diff
@@ -0,0 +1,63 @@
+/* Tested on Crema Grande */
```
Frenzie, line 1: What else would it be tested on. ;-)

```suggestion
```

Frenzie (approved):
lgtm, pinging @pazos

Ki-qu:
Sorry. Upon further testing I found that my hacky approach works rather inconsistently. I changed it back to NookEPDController.

PR #505, opened 2024-08-07
gradle: support changing the location of the assets / libraries

benoit-pierre:
This allow for out-of-tree builds (when combined with `--project-cache-dir` and `--project-cache-dir`).

Necessary for https://github.com/koreader/koreader/pull/12285.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/505)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -24,8 +24,8 @@ android {
 
     sourceSets {
         main {
-            assets.srcDirs = [ '../assets' ]
-            jniLibs.srcDirs = [ '../libs' ]
+            assets.srcDirs = [ '../assets', assetsPath ]
+            jniLibs.srcDirs = [ libsPath ]
         }
     }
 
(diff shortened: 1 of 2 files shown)


app/build.gradle
```diff
@@ -24,8 +24,8 @@ android {
 
     sourceSets {
         main {
-            assets.srcDirs = [ '../assets' ]
-            jniLibs.srcDirs = [ '../libs' ]
+            assets.srcDirs = [ '../assets', assetsPath ]
+            jniLibs.srcDirs = [ libsPath ]
```
Frenzie, line 28: How does that work for lzma and ioctl if you override it for luajit?

benoit-pierre: I did try leaving out that patch out, but could not get something working when not building in the source tree.

PR #516, opened 2024-09-30
jni boilerplate: c -> cpp

pazos:
It isn't intended to go in for a while as it is untested. 

pros:
- It is easier to read for a ramdom lurker. 
- C/JNI boilerplate should be banned from existence.

cons:
- It is c++

but we already use c++ for the rest of boilerplate, so why not here too? :)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/516)


Frenzie (approved):
No real opinion. I suppose "real" objects might have some advantages over pointers but I thought it was really just a slightly different syntax for the exact same thing.

pazos:
> No real opinion. I suppose "real" objects might have some advantages over pointers but I thought it was really just a slightly different syntax for the exact same thing.

They do the same thing. The c version also showcases what the cpp does under the hood.

From a reading perspective the cpp version is way easier. Modern "source readers" are used to deal with instances of things and are `self` aware, so the construct: 
```(vm->AttachCurrentThread(&env, NULL) == 0))``` is easier to grasp than
```((*vm)->AttachCurrentThread(vm, &env, NULL) == 0)))```

They are also easier to think and write. Hence cpp elsewhere :)


But, yeah: `really just a slightly different syntax for the exact same thing.` is 100% correct :)

Frenzie:
Not convinced after all?

PR #518, opened 2024-10-02
assets: remove unused workarounds

pazos:
They seem no longer needed.
Pinging @benoit-pierre?

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/518)


Frenzie (approved):
Although other users (if there are any) might be less appreciative of removing the option.

Frenzie:
> Maybe archive the current master as a legacy branch?

In that case I'd suggest to merely tag it. A branch can always be spun off if someone desires to do so.

PR #520, opened 2024-10-05
add light controller for Tolino Epos 3

dboleslawski:
Built a specific light controller for the Tolino Epos 3. The NTX controller does not seem to change the warmth value (I guess without root) and the lighting max value is 255 there while the Epos 3 caps out at 100.

Has been tested now for a few weeks, can confirm it works rock-solid on my device. Only thing that I had to do, is to manually allow "Modify system settings" for KOReader in the Android settings. Not sure if there is supposed to be a Popup when opening the App the first time..

### DIFF
--- a/app/src/main/java/org/koreader/launcher/TestActivity.kt
+++ b/app/src/main/java/org/koreader/launcher/TestActivity.kt
@@ -27,6 +27,7 @@ import org.koreader.launcher.device.lights.OnyxWarmthController
 import org.koreader.launcher.device.lights.TolinoRootController
 import org.koreader.launcher.device.lights.TolinoNtxController
 import org.koreader.launcher.device.lights.TolinoNtxNoWarmthController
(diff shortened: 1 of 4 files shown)


pazos:
Thanks for your contribution!

> Has been tested now for a few weeks, can confirm it works rock-solid on my device. Only thing that I had to do, is to manually allow "Modify system settings" for KOReader in the Android settings. Not sure if there is supposed to be a Popup when opening the App the first time..

The expected behaviour isn't wired yet as no other driver use it. But you driver correctly declares `needsPermission`  and drivers with that enabled should prompt for the `WRITE_SETTINGS` permission the first time they're invoked (via light dialog or via gestures)

Give a me few days, since I'm currently busy, to wire up things, so you can test this PR against it and confirm it works correctly without having to enable "Modify system settings" manually.

hanakomisa:
#570 has superseded this PR, I assume this can be safely closed now?

Frenzie:
Thanks, yes

PR #525, opened 2024-11-02
Fix Rockchip devices that are not mapped

hugleo:
fix: https://github.com/koreader/android-luajit-launcher/pull/483#issuecomment-2452608471

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -743,9 +743,13 @@ object DeviceInfo {
         deviceMap[EinkDevice.ENERGY] = ENERGY
         deviceMap[EinkDevice.FIDIBOOK] = FIDIBOOK
         deviceMap[EinkDevice.INKBOOK] = INKBOOK
+        deviceMap[EinkDevice.INKPALM_PLUS] = INKPALM_PLUS
         deviceMap[EinkDevice.JDREAD] = JDREAD
         deviceMap[EinkDevice.LINFINY_ENOTE] = LINFINY_ENOTE
+        deviceMap[EinkDevice.MEEBOOK_M6] = MEEBOOK_M6
+        deviceMap[EinkDevice.MEEBOOK_M7] = MEEBOOK_M7
         deviceMap[EinkDevice.MEEBOOK_P6] = MEEBOOK_P6
+        deviceMap[EinkDevice.MOAAN_MIX7] = MOAAN_MIX7
         deviceMap[EinkDevice.MOOINKPLUS2C] = MOOINKPLUS2C
         deviceMap[EinkDevice.NABUK] = NABUK_REGAL_HD
         deviceMap[EinkDevice.NOOK] = NOOK


benoit-pierre:
Not a fan of `device/DeviceInfo.kt`'s code, to put it mildly. Those booleans and weird temporary hash maps…

- are the `TOLINO` and `TOLINO_EPOS3` booleans exclusive?
- what about other `TOLINO_XXX` booleans?
- the `TOLINO_VISION6` boolean is set, but `EINK` is never set to `EinkDevice.TOLINO_VISION6` (but checked in `device/EPDFactory.kt`)
- `EINK` is never set to `EinkDevice.HANVON_960` (but checked in `device/EPDFactory.kt`)
- same with `HYREAD_MINI6`
- `LIGHTS` is never set to `LightsDevice.ONYX_NOTE_PRO` (but checked in `device/LightsFactory.kt`)

benoit-pierre:
Tentative [rework](https://github.com/benoit-pierre/android-luajit-launcher/blob/fdab28c70f8f1374e68e2496e07c3362b30b349d/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt) ([diff](https://github.com/koreader/android-luajit-launcher/compare/master...benoit-pierre:android-luajit-launcher:pr/DeviceInfo.kt_rework)).

pazos:
> Not a fan of device/DeviceInfo.kt's code, to put it mildly. Those booleans and weird temporary hash maps…

Yeah, it was good enough for a few devices but it doesn't scale :)

AFAICT all identifiers are exclusive except `NOOK` and `TOLINO` which are used as a catch them all for EPD routines.

> Tentative [rework](https://github.com/benoit-pierre/android-luajit-launcher/blob/pr/DeviceInfo.kt_rework/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt)

It looks that you got rid of `LightsDevice` and `QuirksDevice`. Maybe search and replace too much? :)

benoit-pierre:
> > Not a fan of device/DeviceInfo.kt's code, to put it mildly. Those booleans and weird temporary hash maps…
> 
> Yeah, it was good enough for a few devices but it doesn't scale :)
> 
> AFAICT all identifiers are exclusive except `NOOK` and `TOLINO` which are used as a catch them all for EPD routines.

How is that suppose to work with the `deviceMap` hashmap? Since the loop only honors the first match?

Also, is the iteration order deterministic? Does it matches the insertion order? The official documentation is confusing: the `keys` accessor is an unordered set, but with a `reversed` method, so I'm assuming the iteration order is not random… ¯\\_(ツ)_/¯

> > Tentative [rework](https://github.com/benoit-pierre/android-luajit-launcher/blob/pr/DeviceInfo.kt_rework/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt)
> 
> It looks that you got rid of `LightsDevice` and `QuirksDevice`. Maybe search and replace too much? :)

No, see the diff for `device/EPDFactory.kt` and `device/LightsFactory.kt`.

pazos:
> How is that suppose to work with the deviceMap hashmap? Since the loop only honors the first match?

I don't understand the question :). It honors the first match. It will return whatever controller or fallback to default if none.

> Also, is the iteration order deterministic? Does it matches the insertion order?

There is no gurantee elements will appear in the specific order they were insterted. They should in a case like this, IIRC.

If that's what you want you might want to look at https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-linked-hash-map/. 

But why?

AFAICT there's no device  with two or more different controllers for the same kind of driver.

> Nope, both gone, see the [diff](https://github.com/koreader/android-luajit-launcher/compare/master...benoit-pierre:android-luajit-launcher:pr/DeviceInfo.kt_rework).

Ok, sorry. That totally works for me if you think it is easier to maintain. Maybe rename `DeviceInfo.EinkDevice` to `Device.Id` since that's what the variable holds?

hugleo:
Isn't it a little weird to put all these TOLINO variant stuff in this generic Device file? https://github.com/benoit-pierre/android-luajit-launcher/blob/c41aa5ee8ee4c6b0c4557f7daf02dde64872a4aa/app/src/main/java/org/koreader/launcher/device/Device.kt#L41C1-L49C52

benoit-pierre:
> > How is that suppose to work with the deviceMap hashmap? Since the loop only honors the first match?
> 
> I don't understand the question :). It honors the first match. It will return whatever controller or fallback to default if none.

If both the `TOLINO` & `TOLINO_EPOS3` can be true, then that code is problematic.

> 
> > Also, is the iteration order deterministic? Does it matches the insertion order?
> 
> There is no gurantee elements will appear in the specific order they were insterted. They should in a case like this, IIRC.
> 
> If that's what you want you might want to look at https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-linked-hash-map/.
> 
> But why?

See above.

> 
> AFAICT there's no device with two or more different controllers for the same kind of driver.
> 
> > Nope, both gone, see the [diff](https://github.com/koreader/android-luajit-launcher/compare/master...benoit-pierre:android-luajit-launcher:pr/DeviceInfo.kt_rework).
> 
> Ok, sorry. That totally works for me if you think it is easier to maintain. Maybe rename `DeviceInfo.EinkDevice` to `Device.Id` since that's what the variable holds?

OK.

benoit-pierre:
> Isn't it a little weird to put all these TOLINO variant stuff in this generic Device file? https://github.com/benoit-pierre/android-luajit-launcher/blob/c41aa5ee8ee4c6b0c4557f7daf02dde64872a4aa/app/src/main/java/org/koreader/launcher/device/Device.kt#L41C1-L49C52

Yeah, I don't like that `is_tolino` flag, a dedicated quirk would be cleaner.

pazos:
If the only usage is https://github.com/koreader/koreader/blob/1e55dda4c765be3316d7ca87be3052943cb1be86/frontend/device/android/device.lua#L141C3-L145C8 we could get rid of it completely.

It should be moved to the wiki as yet another example in https://github.com/koreader/koreader/wiki/Android-tips-and-tricks#customize-keys

pazos:
> If both the TOLINO & TOLINO_EPOS3 can be true, then that code is problematic.

`TOLINO_EPOS3`  shouldn't be a valid `EinkDevice`

hugleo:
They work today because TOLINO is used only for device map and TOLINO_XXX is used only for light map.



> > If both the TOLINO & TOLINO_EPOS3 can be true, then that code is problematic.
> 
> either a `TOLINO_EPOS3` is not catched as a `TOLINO` or the `TOLINO_EPOS3` shouldn't be a valid `EinkDevice` ¿?
> 
> But yeah, that's ultra annoying to deal with. Thanks for the refactor 👍

There is a bug in the code. TOLINO_EPOS3 uses NGL4EPDController in EPDFactory.kt. A similar problem exists for TOLINO_VISION6.
We need to add an exclusion for the TOLINO match.

```
        // Tolino (catch them all)
        TOLINO = !MODEL.contentEquals("tolino epos 3") && !MODEL.contentEquals("tolino vision 6")
        
        &&
        
        (
        
			BRAND.contentEquals(STR_TOLINO) && MODEL.contentEquals("imx50_rdp")
            || MODEL.contentEquals(STR_TOLINO) && (DEVICE.contentEquals("tolino_vision2")
            || DEVICE.contentEquals(STR_NTX))
        )
```

benoit-pierre:
In any case, the `TOLINO` check is now down after more targeted `TOLINO_XXX` checks.

benoit-pierre:
> If the only usage is https://github.com/koreader/koreader/blob/1e55dda4c765be3316d7ca87be3052943cb1be86/frontend/device/android/device.lua#L141C3-L145C8 we could get rid of it completely.
> 
> It should be moved to the wiki as yet another example in https://github.com/koreader/koreader/wiki/Android-tips-and-tricks#customize-keys

Agreed, I updated the wiki page.

hugleo:
superseded by https://github.com/koreader/android-luajit-launcher/pull/526

PR #526, opened 2024-11-04
rework DeviceInfo.kt

benoit-pierre:
Reduce duplication to make it easier to maintain.

Fix the following issues:

- possible problem with `EINK` value on `TOLINO_EPOS3` devices (since both `TOLINO` and `TOLINO_EPOS3` booleans would be set)
- `EINK` was never set to `EinkDevice.TOLINO_VISION6`, but checked in `device/EPDFactory.kt`
- same with `HANVON_960` & `HYREAD_MINI6`
- `LIGHTS` was never set to `LightsDevice.ONYX_NOTE_PRO`, but checked in `device/LightsFactory.kt`

Still not fixed:

~- `LIGTS` could be set to `LightsDevice.ONYX_PALMA`, but was never checked in `device/LightsFactory.kt`~

Also address https://github.com/koreader/android-luajit-launcher/pull/483#issuecomment-2452608471, will conflict with #525.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/526)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/Device.kt
+++ b/app/src/main/java/org/koreader/launcher/device/Device.kt
@@ -10,7 +10,6 @@ class Device(activity: Activity) {
     @Suppress("unused")
     val product = DeviceInfo.PRODUCT
     val needsWakelocks = DeviceInfo.QUIRK_NEEDS_WAKELOCKS
-    val bugRotation = DeviceInfo.QUIRK_NO_HW_ROTATION
     val bugLifecycle = DeviceInfo.QUIRK_BROKEN_LIFECYCLE
     val hasColorScreen = DeviceInfo.HAS_COLOR_SCREEN
 
@@ -31,13 +30,12 @@ class Device(activity: Activity) {
     val einkPlatform = epd.getPlatform()
 
     val properties: String
-      get() = String.format("%s;%s;%s;%s;%s;%s;%b;%b",
+      get() = String.format("%s;%s;%s;%s;%s;%s",
           DeviceInfo.MANUFACTURER,
           DeviceInfo.BRAND,
           DeviceInfo.MODEL,
           DeviceInfo.DEVICE,
           DeviceInfo.PRODUCT,
           DeviceInfo.HARDWARE,
-          DeviceInfo.BOYUE,
-          DeviceInfo.TOLINO)
+      )
 }
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -7,10 +7,12 @@
 package org.koreader.launcher.device
 
 import android.os.Build
-import kotlin.collections.HashMap
+import android.util.Log
 import java.util.Locale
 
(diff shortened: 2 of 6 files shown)


benoit-pierre:
> Still not fixed:
> 
>     * `LIGTS` could be set to `LightsDevice.ONYX_PALMA`, but was never checked in `device/LightsFactory.kt`

According to https://github.com/koreader/android-luajit-launcher/pull/432, the generic controller is the right choice for the Onyx Palma.

Frenzie (approved):
Fine by me

(but note, I didn't check everything in detail ;-) )

hugleo:
> > Still not fixed:
> > ```
> > * `LIGTS` could be set to `LightsDevice.ONYX_PALMA`, but was never checked in `device/LightsFactory.kt`
> > ```
> 
> According to #432, the generic controller is the right choice for the Onyx Palma.

Yes,  lightsMap[LightsDevice.ONYX_PALMA] = ONYX_PALMA was added by mistake

app/src/main/java/org/koreader/launcher/device/Device.kt
```diff
@@ -31,13 +31,23 @@ class Device(activity: Activity) {
...
+      get() = String.format("%s;%s;%s;%s;%s;%s;%s",
           DeviceInfo.MANUFACTURER,
           DeviceInfo.BRAND,
           DeviceInfo.MODEL,
           DeviceInfo.DEVICE,
           DeviceInfo.PRODUCT,
           DeviceInfo.HARDWARE,
-          DeviceInfo.BOYUE,
-          DeviceInfo.TOLINO)
+          when (DeviceInfo.ID) {
```
hugleo, line 41: forgot to remove tolino stuff

benoit-pierre: Yep, good catch.

app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
```diff
@@ -292,621 +138,449 @@ object DeviceInfo {
...
+            Id.SONY_RP1,
+            -> true else -> false
         }
 
         // 4.4+ device without native surface rotation
-        QUIRK_NO_HW_ROTATION = when (QUIRK) {
-            QuirkDevice.EMULATOR -> true
-            else -> false
+        QUIRK_NO_HW_ROTATION = when (ID) {
+            // FIXME: why?
```
pazos, line 561: No real reason. When I did the hw rota thing I found it was broken until, at least, 4.3. So I enabled it on 4.4+.

Also assumed some devices were going to be broken above 4.4, so added the category just in case and the emulator as the single case.

Feel free to remove the emulator or the entire category :)

pazos (approved):

app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
hugleo, line 1: to get rid of codacy warning you can create a constant  CREMA = BRAND == "crema", similar to BOYUE

hugleo (approved):

PR #513, opened 2024-09-27
Tolino Epos 1 warmth lights support

hugleo:
Manufacturer: rakuten kobo inc
    Brand: rakutenkobo
    Model: tolino
    Device: ntx_6sl
    Product: ntx_6sl
    Hardware: e70q20
    Platform: imx6
    Friendly name:
    [test.log](https://github.com/user-attachments/files/17102791/test.log)
    Tolino Epos 1
    Android 4.4
    Tolino Ntx, Freescale/NTX

This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/513)

PR #531, opened 2024-11-26
fix Tolino Epos 2 to the correct light controller

hugleo:
fix https://github.com/koreader/koreader/issues/8482#issuecomment-2500443034

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/531)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
+++ b/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
@@ -74,7 +74,6 @@ object LightsFactory {
                 DeviceInfo.Id.CREMA_CARTA_G,
                 DeviceInfo.Id.MEEBOOK_P6,
                 DeviceInfo.Id.RIDI_PAPER_3,
-                DeviceInfo.Id.TOLINO_EPOS1,
                 DeviceInfo.Id.TOLINO_SHINE3,
                 DeviceInfo.Id.TOLINO_VISION4,
                 DeviceInfo.Id.TOLINO_VISION5,
@@ -88,6 +87,7 @@ object LightsFactory {
                     TolinoNtxNoWarmthController()
                 }
                 DeviceInfo.Id.NOOK_GL4,
+                DeviceInfo.Id.TOLINO_EPOS1,
                 DeviceInfo.Id.TOLINO_EPOS2,
                 -> {
                     logController("TolinoRoot")
--- a/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
+++ b/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
@@ -74,6 +74,7 @@ object LightsFactory {
                 DeviceInfo.Id.CREMA_CARTA_G,
                 DeviceInfo.Id.MEEBOOK_P6,
                 DeviceInfo.Id.RIDI_PAPER_3,
+                DeviceInfo.Id.TOLINO_EPOS1,                
                 DeviceInfo.Id.TOLINO_SHINE3,
                 DeviceInfo.Id.TOLINO_VISION4,
                 DeviceInfo.Id.TOLINO_VISION5,
@@ -87,7 +88,6 @@ object LightsFactory {
                     TolinoNtxNoWarmthController()
                 }
                 DeviceInfo.Id.NOOK_GL4,
-                DeviceInfo.Id.TOLINO_EPOS1,
                 DeviceInfo.Id.TOLINO_EPOS2,
                 -> {
                     logController("TolinoRoot")
--- a/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
+++ b/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
@@ -74,7 +74,7 @@ object LightsFactory {
                 DeviceInfo.Id.CREMA_CARTA_G,
                 DeviceInfo.Id.MEEBOOK_P6,
                 DeviceInfo.Id.RIDI_PAPER_3,
-                DeviceInfo.Id.TOLINO_EPOS1,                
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
```diff
@@ -88,6 +87,7 @@ object LightsFactory {
                     logController("TolinoNTXNoWarmth")
                     TolinoNtxNoWarmthController()
                 }
+                DeviceInfo.Id.TOLINO_EPOS2,
                 DeviceInfo.Id.NOOK_GL4,
```
benoit-pierre, line 91:
```suggestion
                DeviceInfo.Id.NOOK_GL4,
                DeviceInfo.Id.TOLINO_EPOS2,
```

Frenzie (approved):

pazos:
Makes no sense to revert. It must be a FW thingy. @rola25: are you on the lastest firmware for your device? Which one?

Maybe I'm missing something,

pazos:
> Makes no sense to revert.

I mean, `TolinoRoot` was the initial implementation for warmth lights at all. Made by @zwim. It works but requires root.
Since then a new warmth driver is available. IIRC we never changed a driver for a device without an user report. So if every device is now on `TolinoNtx` it must be because they were reports on them (and because, if both work, the non-root option is better).

mitol388:
@rola25 already reports warmth doesn't work without TolinoRoot driver here: https://github.com/koreader/koreader/issues/8482#issuecomment-2500443034. I can also confirm warmth only works with TolinoRoot. Here is a log where I test brightness/warmth with both TolinoNTX and TolinoRoot (with Tolinot NTX only brightness works, with TolinoRoot both work): 
[test.log](https://github.com/user-attachments/files/17938700/test.log)

hugleo:
> 
>  IIRC we never changed a driver for a device without an user report. 
We did, it was my fault here: https://github.com/koreader/android-luajit-launcher/pull/513/files#diff-7d3fed5a40ea52bd9274153219933a8cac5bdacd5622db82e3d45c8ed5fcbcb4L13. 
You can go ahead and merge this PR, as it was a regression.

pazos:
Thanks for the info, @hugleo 

Now makes sense. Then please also change `TOLINO_EPOS1`

pazos:
> For TOLINO_EPOS1 TolinoNtx seems to be good. https://github.com/koreader/koreader/issues/8482#issuecomment-2369111458
At least until now nobody complains about this device.

Please no. Both on the same place. (Root!).
TOLINO EPOS 1 and 2 share warmth driver since the beginning :)

But the FW version thingy might apply. Newer firmwares might be compatible with the non-root implementation. The old root method might be invalid... 
If somebody reports that a current driver no longer works and a new one works we would need to notice it an have a policy :/
Otherwise we will bump and dump drivers based on the FW of the people reporting :/.

hugleo:
OK, done.

AFAIK, the root driver was developed and tested for the TOLINO EPOS 2.
For the TOLINO 1, I couldn't find any old reports indicating whether it works or doesn't work.

Good to know that they might still release new firmware for old TOLINOs even today. Next time, I'll buy a TOLINO! :D

pazos (approved):

hugleo:
> Thanks!
> 
> > For the TOLINO 1, I couldn't find any old reports indicating whether it works or doesn't work.
> 
> Both were "Root" before the regression.
> 
> > Good to know that they might still release new firmware for old TOLINOs even today. Next time, I'll buy a TOLINO! :D
> 
> :))) I mean in general, like Onyxes or new tolino "vision" devices.

But they really do: https://mytolino.com/software-updates-for-tolino-ereaders/
16.2.0 2024 version is also available for Tolino Epos 1.

rola25:
Hi, no I'm not on the latest firmware, mine is 16.1.0. I also have a Tolino Epos 1 - on weekend I can test koreader there.

hugleo:
Can you test the Tolino Epos 1 on the eInk test and tell if will work TolinoRoot or TolinoNtx lights? This way we will have at least two reports on the device and can commit to the best option.

rola25:
I tested it on the Tolino Epos 1 -> warmth is working there with both drivers, ntx and tolinoroot ... very strange ;-)
Do you need the device info? Hardware is e70q20 for Tolino Epos 1 and e80k00 for Tolino Epos 2

hugleo:
@pazos I think that's all, go for the root here as work for both devices.

hugleo:
Reading this comment: https://github.com/koreader/koreader/issues/12843 is implying root I guess.

pazos:
All non recent Nooks and Tolinos come with a closed version of android where their stock reading app is a big monolithic launcher.

Usually the way to "gain access" involves a way to get adb working and using adb to install an alternate launcher that lets you install and launch other apks.

New Tolino (vision?) comes with an Onyx/Mebook like launcher and with recent-ish android versions. Those devices are already "opened" but obtaining root in there might be difficult. No idea.

tl;dr: if v15 and v16 of the tolino epos1 firmware work with the non-root controller we can upgrade. Otherwise stay as is, root is easily achievable in "closed" tolinos.

hugleo:
What do you prefer? merge this as is or revert to ntx?

pazos:
@hugleo https://github.com/koreader/koreader/issues/12843#issuecomment-2518492774

Now I think the non-root is better. Hopefully works on v16 too. Who knows?

hugleo:
> @hugleo [koreader/koreader#12843 (comment)](https://github.com/koreader/koreader/issues/12843#issuecomment-2518492774)
> 
> Now I think the non-root is better. Hopefully works on v16 too. Who knows?

confirmed

PR #540, opened 2025-01-30
Update DeviceInfo.kt

Schizosaccaromyces:
Corrected typo in device product identifier for Inkbook Focus Plus

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/540)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -272,7 +272,7 @@ object DeviceInfo {
             -> Id.INKBOOKFOCUS
 
             // InkBook Focus Plus
-            PRODUCT == "r07802" && MODEL == "focus plus"
+            PRODUCT == "ro7802" && MODEL == "focus plus"
             -> Id.INKBOOKFOCUS_PLUS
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
```diff
@@ -272,7 +272,7 @@ object DeviceInfo {
             -> Id.INKBOOKFOCUS
 
             // InkBook Focus Plus
-            PRODUCT == "r07802" && MODEL == "focus plus"
+            PRODUCT == "ro7802" && MODEL == "focus plus"
```
hugleo, line 275: I think should be better something like:

```
            // InkBook Focus
            DEVICE == "px30_eink" && MODEL == "focus"
            -> Id.INKBOOKFOCUS

            // InkBook Focus Plus
            DEVICE == "rk3566_eink" && MODEL == "focus plus"
            -> Id.INKBOOKFOCUS_PLUS
```

Schizosaccaromyces:
So as long as enough fields are unique it doesn't matter which ones are specified?

hugleo (approved):

PR #546, opened 2025-03-13
handle uris from browsers

pazos:
Or any app that states `android.intent.category.BROWSABLE` on their intents.

Restricted to `file` and `content` schemes from any host.

Also added a filter by extension, in case the sending app is nuts (I never needed it here but I see it is commonly used)

Also added markdown to the list.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/546)


Frenzie (approved):
> Also added a filter by extension, in case the sending app is nuts (I never needed it here but I see it is commonly used)

I'm not sure if I entirely understand, but alright. :-)

Frenzie:
I see, that makes sense.

PR #563, opened 2025-06-18
Removed ONYX Book GO COLOR 7 From Onyx/Qualcomm EDP

komadorirobin:
This is in response to @jadehawk fork of the same file. I've removed Go Color 7 as well, for the exact same reason, namely:

"Due to intermittent black screen when device awakes from sleep. I was able to duplicate the issue consistently when the EDP driver was assigned to Onyx/Qualcomm even if it did work on the developer Test Page."

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/563)


hugleo (approved):

Frenzie:
It is suggested here that the situation with it in there is preferable.
https://www.mobileread.com/forums/showthread.php?p=4538072#post4538072

> Sure, Koreader sometimes turns into a black screen on many Onyx devices, which is resolved by restarting it. However, I don't see how removing the driver is a solution. If I can't set full refresh to every page turn, I can't read at all.
> 
> Occasionally just killing the app and starting it again, while slightly annoying, is a price I'm willing to pay until it gets fixed.

Issue #532, opened 2024-12-11
Tolino Shine 4 EPD support


pazos:
continuing from the [post](https://github.com/koreader/koreader/issues/8482#issuecomment-2533638374) 

> One more thing need to mention: There is a property 'persist.mode.global' which the default value is 8000 after reboot. The NGL4 driver dose not work with value 8000. If i set the value to 0 (after i root the device and use adb shell setprop), the NGL4 driver will work.

When I  for this driver it looked like any internal code indeed tried to reset it to zero everywhere, probably disabling this global mode of refresh to be active. On my Nook Glowlight 4e it is zero on start, probably depends on the manufacturer. Random thoughts:

- The problem with these system properties that they require root access to be changed so not so great to be used in KOReader
- Just refreshed my understanding, according to [the docs](https://source.android.com/docs/core/architecture/configuration/add-system-properties), persist properties survive reboot, to theoretically this can be some last software that you used before the reboot that forgot to reset it. 
- For ones with root access there's always several ways to force it to zero on start (independently of KOReader), but this solution is not ok for mass use.  
- I recall that I seemed to notice the initial value for this property in some system start up script, but I'm not sure. Just looks at some variant of Android image from my device, could not find it in `default.prop` and `system/build.prop` so it can be false memories.
- can not currently decode this 8000 value, at least for the sources I managed to quickly find.

_Originally posted by @Codereamp in https://github.com/koreader/koreader/issues/11110#issuecomment-2533776202_

Codereamp:
A suggestion for @f-coco (from [the post](https://github.com/koreader/koreader/issues/11110#issuecomment-1963409521)). Have you noticed the same effect on your Tolino 4 (If it is ... Shine, I'm not very familiar with this manufacturer's models :) ? Maybe you can add something? Thanks

f-coco:
> A suggestion for @f-coco (from [the post](https://github.com/koreader/koreader/issues/11110#issuecomment-1963409521)). Have you noticed the same effect on your Tolino 4 (If it is ... Shine, I'm not very familiar with this manufacturer's models :) ? Maybe you can add something? Thanks

Ok, I will check all my B300 devices made by Netronix this weekend.

f-coco:
For Mooink S, run getprop from adb shell after a reboot.
persist.mode.global=0
For Tolino Shine 4, run getprop from adb shell after a reboot.
persist.mode.global=0
For Tolino Vision 6, run getprop from adb shell after a reboot.
[...]

PR #568, opened 2025-08-31
Adding better light controls for Tolino Vision 6 based on PR #520

hanakomisa:
I lifted the light control additions from https://github.com/koreader/android-luajit-launcher/pull/520, only slightly dusted it up to make it work with the current codebase. All credits to @dboleslawski .  

This comes with some caveats:
1. As the original PR stated, this needs the "Modify System Settings" permission toggled on for KOReader for the sliders to work. Though considering there are root only lighting controls as well as ADB based ones, I feel like this is an okay tradeoff. (Should modify the wiki to mention this?)
2. The Vision 6 has inverted warmth control, but there are no other known issues otherwise.
~~3. I do not have a Shine 4 personally to test, but I'm operating under the assumption that it will work just fine seeing how architecturally similar the Epos3/Vision6/Shine4 all are. Testers would be crucial here.~~

~~Additionally, I've gone ahead and added the missing info for the Tolino Shine 4, using the information from this comment https://github.com/koreader/koreader/issues/8482#issuecomment-2533638374 as well as this issue: https://github.com/koreader/android-luajit-launcher/issues/532 .~~

~~Seeing how the Epos 3 and Vision 6 already has proper e-ink support using NGL4 EPD (with some caveats that I'll get into), I feel like it's fair to add support to the Shine 4 as well. I'm not entirely sure if I've completely added everything though, so would appreciate extra eyes as I'm completely new to this and this is my first time ever really contributing to anything :lul:~~

Now, I don't know why `persist.mode.global` gets set to 8000 when the official Tolino app is launched, but as I mentioned here in https://github.com/koreader/android-luajit-launcher/issues/532#issuecomment-3239575034, using the AOSP settings app can reliably set it back to 0.

I also recently observed said prop being set to `1800010`, I have no clue how that happened and it's still something I'm looking into. It seems to have enabled the "A2" mode of the screen as contrast was quite poor and has a lot of ghosting, but responded very quickly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/568)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -67,6 +67,8 @@
                 <data android:mimeType="application/rtf" />
                 <data android:mimeType="application/tcr" />
                 <data android:mimeType="application/vnd.amazon.mobi8-ebook" />
+                <data android:mimeType="application/vnd.comicbook-rar" />
+                <data android:mimeType="application/vnd.rar" />
                 <data android:mimeType="application/vnd.comicbook+tar" />
                 <data android:mimeType="application/vnd.comicbook+zip" />
                 <data android:mimeType="application/vnd.ms-htmlhelp" />
@@ -117,6 +119,7 @@
                 <data android:pathPattern=".*\\.pdf" />
                 <data android:pathPattern=".*\\.rtf" />
                 <data android:pathPattern=".*\\.mobi" />
+                <data android:pathPattern=".*\\.cbr" />
                 <data android:pathPattern=".*\\.cbt" />
                 <data android:pathPattern=".*\\.cbz" />
                 <data android:pathPattern=".*\\.odt" />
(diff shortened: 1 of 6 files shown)


Frenzie (approved):
Nothing jumps out to me, so if it's tested. ;-)

hanakomisa:
Forgot to add the Shine 4's ID under enum, whoops

hugleo:
Since is tested only for Vision 6 then add only Vision 6 in the new driver should be good to go.

hanakomisa:
> Since is tested only for Vision 6 then add only Vision 6 in the new driver should be good to go.

I have removed all of the Shine 4's support and removed the Epos 3 mapping. I'll see if I can get some testers for those devices but in the meantime this should satisfy all?

hugleo:
@pazos Since you've commented on the based PR #520 see if you're okay with it aside from the "Modify System Settings" permission being toggled on as OP mentioned.

PR #573, opened 2025-10-19
add light controller for Rockchip3566 (Inkbook Focus Plus)

matejbart:
> [!NOTE]  
> Testing device:
> MANUFACTURER: inkbook
> BRAND       : rockchip
> MODEL       : focus plus
> DEVICE      : rk3566_eink
> PRODUCT     : r0782a
> HARDWARE    : rk30board

This PR adds a new light controller for the [`InkBOOK Focus Plus`](https://inkbook.eu/de/products/inkbook-focus-plus) e-reader, which is based on the `Rockchip3566` device. The generic (or any other existing) controller doesn't work for this model at all.

The light is controlled by two hardware LEDs called `leda` and `ledb`. The input for the led values needs to be computed based on the brightness and warmth. The computation function has been reverse engineered from the `inkBOOKSettings.apk`.

Controlling the light is done by writing to the "files" directly which works without root access. I also added a fallback by using reflection in case that the write/read should not be permitted on some other devices even though I think it's not that versatile as it relies on the existence of a specific class.

I tried to mimic the behavior of the system app as good as possible, i.e. as soon as a new value is set, the value is also persisted in the corresponding system property:
- `brightness`: `persist.inkbook.frontlight.brightness`
- `warmth`: `persist.inkbook.frontlight.temperature`
- `leda`: `persist.yitoa.backlight.leda.brightness`
- `ledb`: `persist.yitoa.backlight.ledb.brightness`

And finally, a broadcast action `com.android.systemui.statusbar.action.light_change` is sent out so that the systemUI app bar updates the light icon. However, the light handlers in the inkBOOK related .apks are not listening to that broadcast action, they only send it out in case something changes. 
Furthermore, they also cache their last known brightness in the RAM, so when the koreader manipulates some light settings, then the inkBOOK apps are unaware of that, resulting in an unsynced state.
___
This section just contains details about the device specific APKs and light handling.

I started my investigations by reverse engineering the following system apks to find out how the light is being controlled:
- `/system_ext/priv-app/SystemUI_inkBook/SystemUI_inkBook.apk` - the system UI which e.g. displays the status bar. 
- `/odm/bundled_persist-app/inkBook3566Setting/inkBook3566Setting.apk` an app that mainly is responsible for displaying context menus for the status bar. it listens to broadcast receiver actions from the status bar:
   - click on light icon --> listens to the action, but does nothing.
   - long click on light icon --> starts `LightActivity` of the `inkBOOKSettings.apk` where the user can set the brightness and temperature with UI sliders.
- `/odm/bundled_persist-app/inkBOOKSettings/inkBOOKSettings.apk` - the main settings app, which also listens to broadcast receiver actions from the status bar.
   - this apk uses `android.yitoa.rk3566.SysUtil` from the platform's `framework.jar` to control the light. That util class provides methods to write to the leda/ledb files.
- `lightProvider.apk` (a provider which seems to handle the light as well but it's not exposed, so I guess it's just dead code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/573)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ app/build
 assets/libs
 assets/module
 jni/luajit/build
+.DS_Store
--- a/app/src/main/java/org/koreader/launcher/TestActivity.kt
+++ b/app/src/main/java/org/koreader/launcher/TestActivity.kt
@@ -28,6 +28,7 @@ import org.koreader.launcher.device.lights.TolinoRootController
 import org.koreader.launcher.device.lights.TolinoNtxController
 import org.koreader.launcher.device.lights.TolinoNtxNoWarmthController
 import org.koreader.launcher.device.lights.BoyueS62RootController
+import org.koreader.launcher.device.lights.Rockchip3566Controller
 import org.koreader.launcher.dialog.LightDialog
 import org.koreader.launcher.dialog.ToolTip
 
@@ -81,6 +82,7 @@ class TestActivity: AppCompatActivity() {
         lightsMap["Tolino Root"] = TolinoRootController()
         lightsMap["Tolino Ntx"] = TolinoNtxController()
         lightsMap["Tolino Ntx (no warmth)"] = TolinoNtxNoWarmthController()
+        lightsMap["Rockchip RK3566"] = Rockchip3566Controller()
 
         // Device ID
         binding.info.append("Manufacturer: ${DeviceInfo.MANUFACTURER}\n")
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -18,6 +18,7 @@ object DeviceInfo {
     private const val STR_NTX = "ntx_6sl"
     private const val STR_ROCKCHIP = "rockchip"
     private const val STR_TOLINO = "tolino"
+    const val DEVICE_RK3566_EINK = "rk3566_eink"
 
     val MANUFACTURER: String
     val BRAND: String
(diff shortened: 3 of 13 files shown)


jkl16:
Unfortunately, the driver for `XIAOMI_READER` isn't functional based on the logs - the attached stack trace shows both FileNotFoundException and ClassNotFoundException errors.

[logcat.log](https://github.com/user-attachments/files/31852199/logcat.log)

For context, when I implemented the RK3566 EPD driver two years ago, I looked into light control, but on my specific device, it required root access due to permission restrictions.

matejbart:
> Unfortunately, the driver for `XIAOMI_READER` isn't functional based on the logs - the attached stack trace shows both FileNotFoundException and ClassNotFoundException errors.
> 
> [logcat.log](https://github.com/user-attachments/files/31852199/logcat.log)
> 
> For context, when I implemented the RK3566 EPD driver two years ago, I looked into light control, but on my specific device, it required root access due to permission restrictions.

Thanks for testing, my guess is that the controller I have implemented only works for the particular manufacturer and model. Probably, I should name it differently as it clearly doesn't work for all Rockchip3566 based devices out of the box.

Could you please share the full spec of your testing device? (manufacturer, brand, model, etc.)

jkl16:
Here is my device specs:
- MANUFACTURER: "xiaomi"
- BRAND: "xiaomi"
- MODEL: "xiaomi_reader"
- DEVICE: "rk3566_eink"
- HARDWARE: "rk30board"

The Moaan Mix7 and Mix7s share very similar hardware and likely run the same software stack.

When I investigated frontlight control on this device previously, the system class `android.os.MoanLedControl` was responsible for managing the lights. While reading values via reflection works fine, calling setter methods throws a `SecurityException` because it requires system-level permissions that KOReader cannot obtain.

PR #580, opened 2026-02-11
EventReceiver: use Unix domain sockets with socketpair() instead of mkfifo()

Frenzie:
This is extremely similar to the approach described in <https://github.com/koreader/koreader/discussions/14358#discussioncomment-14500895>.

Without this change events like android.intent.action.ACTION_POWER_CONNECTED don't reach the frontend, but I'm primarily interested in TEXT_INPUT from the IME/onscreen-keyboard.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/580)

### DIFF
--- a/jni/android_native_app_glue/android_native_app_glue.c
+++ b/jni/android_native_app_glue/android_native_app_glue.c
@@ -22,6 +22,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <sys/socket.h>
 #include <sys/stat.h>
 
 #include <android/log.h>
@@ -30,7 +31,7 @@
 
 #define  TAG "NativeGlue"
 
-#define FIFO_NAME "alooper.fifo"
+static int g_event_write_fd = -1;
 
 #define LOGE(...) ((void)__android_log_print(ANDROID_LOG_ERROR, TAG, __VA_ARGS__))
 #define LOGV(...) ((void)__android_log_print(ANDROID_LOG_VERBOSE, TAG, __VA_ARGS__))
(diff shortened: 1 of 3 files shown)


jni/android_native_app_glue/android_native_app_glue.c
```diff
@@ -510,3 +511,10 @@ void ANativeActivity_onCreate(ANativeActivity* activity, void* savedState, size_
 
     activity->instance = android_app_create(activity, savedState, savedStateSize);
 }
+
+JNIEXPORT jint JNICALL
+Java_org_koreader_launcher_EventReceiver_nativeGetEventSocketFd(JNIEnv* env, jclass clazz) {
+    (void)env;
+    (void)clazz;
+    return g_event_write_fd;
+}
```
Frenzie, line 541: The diff looks quite large, but conceptually the only real change is this here, with the associated `System.loadLibrary("luajit-launcher")` and `private external fun nativeGetEventSocketFd(): Int`.

Frenzie:
@pazos @hugleo Any thoughts? Since those events don't currently arrive anyway it's a clear improvement in the actually working department, but I'm mildly worried I might be missing something obvious to more Android-attuned people.

hugleo:
Did you check the native code to see if it's safer to close the FD so the native code doesn't keep a old reference?

Frenzie:
I'm not sure if I quite understand what you mean. Some potential issue when the program is shutting down?

Frenzie:
@hugleo Are you satisfied by the change?

hugleo:
Yes. I didn’t check deeply or test against my devices but at first glance it looks safe if it has been tested.

Frenzie:
I'll merge it then to reduce the chance of conflicts.

PR #585, opened 2026-03-05
bump sdk, gradle and deps

pazos:
superseedes https://github.com/koreader/android-luajit-launcher/pull/572

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/585)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
     name: ${{ matrix.variant }}
     runs-on: ubuntu-latest
     container:
-      image: koreader/koandroid:0.9.1-22.04
+      image: benoitpierre/koandroid:0.10.0-22.04
 
     steps:
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,9 +13,5 @@ insert_final_newline = true
 indent_size = 2
 trim_trailing_whitespace = false
 
-[Makefile]
-indent_style = tab
-indent_size = 8
-
 [*.{js,json,css,scss,sass,html,handlebars,tpl}]
 indent_size = 2
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ A good number of Android native API headers are readily presented via FFI alread
 
 Have a look at KOReader's [llapp_main.lua](https://github.com/koreader/koreader/blob/master/platform/android/llapp_main.lua) file. You can use it as a starting point for your own app.
 
-The real starting point, called from JNI/C, is the run() function in android.lua. It sets up a few things, namely FFI definitions for the Android native API (since it uses that itself for a few things) and some wrapper functions for logging. Also, it registers the "android" module in package.loaded, so you can access it in your own code via require("android"). It also registers a new package loader which can load Lua code from the activity's asset store, so you can use require() for Lua code stored there.
+The real starting point, called from JNI/C, is the run() function in android.lua. It sets up a few things, namely FFI definitions for the Android native API (since it uses that itself for a few things) and some wrapper functions for logging. Also, it registers the "android" module in `package.loaded`, so you can access it in your own code via require("android"). It also registers a new package loader which can load Lua code from the activity's asset store, so you can use require() for Lua code stored there.
 
 ## Disclaimer
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,20 +1,18 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+
 apply plugin: 'com.android.application'
-apply plugin: 'kotlin-android'
 
 android {
-    namespace 'org.koreader.launcher'
-    ndkVersion '23.2.8568313'
-    ndkPath System.getenv('ANDROID_NDK_HOME')
-    compileSdk rootProject.ext.compileSdk
+    namespace = 'org.koreader.launcher'
+    ndkVersion = '23.2.8568313'
+    ndkPath = System.getenv('ANDROID_NDK_HOME')
+    compileSdk = rootProject.ext.compileSdk
     def baseVersionCode = versCode as Integer
 
-    aaptOptions {
-        noCompress '7z'
-    }
 
     defaultConfig {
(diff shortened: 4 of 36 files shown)


pazos:
bumping targetSdk is always error-prone :/

benoit-pierre:
Try without [this line](https://github.com/koreader/android-luajit-launcher/blob/master/.github/workflows/build.yml#L49) first.

benoit-pierre:
The SDK version should be bumped in https://github.com/koreader/koreader-base/blob/master/toolchain/Makefile#L34, and an updated koandroid docker image need to be cut.

pazos:
```
> Failed to install the following SDK components:
      build-tools;36.0.0 Android SDK Build-Tools 36
      platforms;android-36 Android SDK Platform 36
  The SDK directory is not writable (/opt/android-sdk-linux)
```

.github/workflows/build.yml
benoit-pierre: I've pushed an updated image to my registry, based on your PR to base.

```suggestion
      image: benoitpierre/koandroid:0.10.0-22.04
```

.github/workflows/build.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       - name: Build
         run: |
           # Ensure the build fail if gradle tries to install/update an SDK package.
-          echo 'android.builder.sdkDownload=false' >>gradle.properties
+          # echo 'android.builder.sdkDownload=false' >>gradle.properties
```
benoit-pierre:
```suggestion
          echo 'android.builder.sdkDownload=false' >>gradle.properties
```

benoit-pierre:
For the builtin Kotlin / new DSL stuff:
```diff
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,6 +1,5 @@
 //file:noinspection GrDeprecatedAPIUsage
 apply plugin: 'com.android.application'
-apply plugin: 'kotlin-android'
 
 android {
     namespace = 'org.koreader.launcher'
@@ -60,8 +59,10 @@ android {
         targetCompatibility = JavaVersion.VERSION_17
     }
 
-    kotlinOptions {
-        jvmTarget = '17'
+    kotlin {
+	compilerOptions {
+	    jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_17
+	}
     }
 
     flavorDimensions = [ 'ABI', 'CHANNEL'  ]
@@ -121,11 +122,13 @@ android {
     androidResources {
         noCompress '7z'
     }
+}
 
-    applicationVariants.configureEach { variant ->
-        variant.outputs.configureEach {
-            outputFileName = "NativeActivity.apk"
-        }
+androidComponents {
+    onVariants(selector().all()) {
+	outputs.forEach {
+	    it.outputFileName.set("NativeActivity.apk")
+	}
     }
 }
 
diff --git a/gradle.properties b/gradle.properties
index 1e55d0c..1856671 100644
--- a/gradle.properties
+++ b/gradle.properties
@@ -25,5 +25,3 @@ android.uniquePackageNames=false
 android.dependency.useConstraints=true
 android.r8.strictFullModeForKeepRules=false
 android.r8.optimizedResourceShrinking=false
-android.builtInKotlin=false
-android.newDsl=false
```

benoit-pierre:
Do we need all those new gradle properties? I've had success locally building and running an ARM APK with just:
```ini
org.gradle.warning.mode=all
org.gradle.caching=true
org.gradle.configuration-cache=true
org.gradle.configureondemand=true
org.gradle.parallel=true

# Use AndroidX compat libraries
android.useAndroidX=true

# Fallback values used in gradle build
projectName=luajit-launcher
versCode=1
versName=1.0
assetsPath=../assets
libsPath=../libs

# JNI libraries
sevenZipLib=z

# Compilation features
android.defaults.buildfeatures.shaders=false
```

pazos:
oh, I saw the builtin Kotlin / new DSL stuff but I thought it was going to be hard.  Do you want to commit it as part of this PR?


> Do we need all those new gradle properties?

At least `android.generateSyncIssueWhenLibraryConstraintsAreEnabled=false` for gh actions

benoit-pierre:
> > Do we need all those new gradle properties?
> 
> At least `android.generateSyncIssueWhenLibraryConstraintsAreEnabled=false` for gh actions

What does that do? I can't find any documentation.

benoit-pierre:
I updated my docker image with your latest change from https://github.com/koreader/koreader-base/pull/2286, and now of course the build is broken…

benoit-pierre:
I guess we would need to bump to `platforms;android-36.1` too? Can't find any useful information, release notes, on SDK 36.1… Man, android development is such a shit show…

benoit-pierre:
OK, I re-tagged it to the previous version, let's forget about `build-tools;36.1.0` for now. ¯\\_(ツ)_/¯

benoit-pierre:
> > > Do we need all those new gradle properties?
> > 
> > 
> > At least `android.generateSyncIssueWhenLibraryConstraintsAreEnabled=false` for gh actions
> 
> What does that do? I can't find any documentation.

It seems to build fine on GA wihout it: https://github.com/benoit-pierre/android-luajit-launcher/actions/runs/22800505004. Might be some interaction with some of those other properties…

benoit-pierre:
OK, so linked to `android.dependency.useConstraints`, which  now default to false:

> android.dependency.useConstraints
>
> Controls the use of dependency constraints between configurations.
>
> The default in AGP 9.0 is false which only uses constraints in application device tests (AndroidTest). Setting this to true will revert back to the 8.13 behavior. 	true → false 	Don't use dependency constraints everywhere unless you need them. Accepting the new default of this flag also enables optimizations in the project import process which should reduce the import time for builds with many Android library subprojects.

benoit-pierre:
We can get rid of `android.defaults.buildfeatures.shaders=false` too (false is the new default).

benoit-pierre:
And the APK seems to work fine without `android.nonTransitiveRClass=false` too (whose default was changed to true in AGP 8.0.0).

PR #586, opened 2026-03-05
Fix/rewrite onyx sdk lights controller

Tukks:
This PR addresses several issues encountered with frontlight and warmth (CTM) control on the Onyx Boox Go 7 when using ADB Mode.

I extracted and decompiled the APK from my device and tried to reproduce the logic here.  
In the APK, we can see that different kinds of devices are handled: frontlight-only devices, devices with warmth, etc.

The following device types are handled:

- IMX6Device  
- RK31XXDevice  
- IMX7Device  
- RK32XXDevice  
- RK33XXDevice  
- RK3026Device  

Here are the decompiled classes in a zip file (only the relevant ones):  
[extracted.zip](https://github.com/user-attachments/files/25780242/extracted.zip)

## Testing & Verification

Tested on **Boox Go 7**.

There are 4 modes available in the Boox UX: **Off mode**, **Custom mode**, **Soft mode**, and **Bright mode**.

- **Off Mode**: No slider is active in the Boox UX. KOReader can enable the frontlight, but a bug immediately sets the value to **20** instead of the value provided. It is impossible to make warmth work in this mode.

- **Custom Mode**: Only the frontlight slider is active in the UX, but KOReader can also control warmth. We override the system behavior.

- **Soft Mode and Bright Mode**: Both the warmth and frontlight sliders are active in the UX. KOReader can control the frontlight and warmth normally.

The minimum and maximum values are correctly retrieved. (Before it was 0 to 255, now 32 on the Go 7)

## Remarks

If needed, I can create a new controller class instead of overriding what already appears to work.  
This would allow users to test the driver through the developer options before replacing the current implementation.

## Implementation Notes

An LLM was used as an assistance tool while analyzing the decompiled APK and reproducing the logic.  
The final implementation was reviewed and adjusted manually.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/586)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
+++ b/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
@@ -17,7 +17,6 @@ object LightsFactory {
                 DeviceInfo.Id.ONYX_GALILEO2,
                 DeviceInfo.Id.ONYX_GO_COLOR7,
                 DeviceInfo.Id.ONYX_GO6,
-                DeviceInfo.Id.ONYX_GO7,
                 DeviceInfo.Id.ONYX_GO7GEN2,
                 DeviceInfo.Id.ONYX_NOTE_AIR_3C,
                 DeviceInfo.Id.ONYX_NOTE_AIR_4C,
@@ -82,7 +81,8 @@ object LightsFactory {
                 DeviceInfo.Id.ONYX_POKE4LITE,
                 DeviceInfo.Id.ONYX_TAB_ULTRA,
                 DeviceInfo.Id.STORYTEL_READER2,
-                -> {
+                DeviceInfo.Id.ONYX_GO7,
+                    -> {
                     logController("Onyx/Sdk")
                     OnyxSdkLightsController()
                 }
--- a/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
+++ b/app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
@@ -82,7 +82,7 @@ object LightsFactory {
                 DeviceInfo.Id.ONYX_TAB_ULTRA,
                 DeviceInfo.Id.STORYTEL_READER2,
                 DeviceInfo.Id.ONYX_GO7,
-                    -> {
+                -> {
                     logController("Onyx/Sdk")
                     OnyxSdkLightsController()
                 }
--- a/app/src/main/java/org/koreader/launcher/device/lights/OnyxSdkLightsController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/lights/OnyxSdkLightsController.kt
@@ -8,14 +8,13 @@ import java.lang.reflect.Method
 
 // ─── Constants mirroring FrontLightController / BaseBrightnessProvider ────────
 
-private const val LIGHT_TYPE_FL = 1 // FLBrightnessProvider — single channel
-private const val LIGHT_TYPE_CTM_WARM = 2 // WarmBrightnessProvider — warm channel
-private const val LIGHT_TYPE_CTM_COLD = 3 // ColdBrightnessProvider — cold channel
-private const val LIGHT_TYPE_CTM_ALL = 4 // open/close the whole CTM unit
-private const val LIGHT_TYPE_TEMP = 6 // CTMTemperatureProvider — CTM warmth param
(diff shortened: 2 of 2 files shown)


Tukks:
I think this PR is ready for review.

I tested it extensively on my Boox Go 7 and everything works smoothly (brightness, warmth, and frontlight control).

For reference, I also attached a decompiled `framework.jar`, which contains additional information compared to the KCB one:

[framework-out.zip](https://github.com/user-attachments/files/25795054/framework-out.zip)

Codacy reports **Complexity +65**, but this mainly comes from reproducing the same behaviour as the Onyx framework.
The logic is required to correctly support the different brightness modes (FL, WARM/COLD, CTM) and to mirror how `DeviceController` handles frontlight control internally.

Additional note:

It might be possible to exit the *Off* mode (where the bug appears) by sending the following broadcast:

```kotlin
private const val TOGGLE_BRIGHTNESS_STYLE_REQUEST_ACTION = "com.onyx.TOGGLE_BRIGHTNESS_STYLE_REQUEST"
private const val ARGS_TYPE = "type" // verify this constant name in BroadcastHelper

fun setCustomLightStyle(context: Context) {
    val intent = Intent(TOGGLE_BRIGHTNESS_STYLE_REQUEST_ACTION).apply {
        putExtra(ARGS_TYPE, "style_type_custom")
        addFlags(Intent.FLAG_INCLUDE_STOPPED_PACKAGES)
    }
    context.sendBroadcast(intent)
    Log.d(TAG, "sendBroadcast → style_type_custom")
}
```

However, this introduces inconsistent behaviour: the device briefly switches to *Light* mode and then sometimes returns to *Off* mode after a short time. In some cases it remains in *Light* mode as expected.

My guess is that the system service responsible for frontlight management resets the mode in certain situations.

Because of this behaviour, I did not include this approach in the implementation for now.

If someone wants to explore this further, everything seems to be handled by `LightStyleConfig`, which loads a configuration file and writes updates to it. The available styles appear to be:

* `style_type_custom`
* `style_type_bright`
* `style_type_soft`
* `style_type_auto_brightness` (if supported by the device, not the case for the Boox Go 7)
* `style_type_close_brightness` (the *Off* mode)

Example log output:

```
LightStyleConfig save {"currentStyle":"style_type_close_brightness","lightStyle":{"style_type_auto_brightness":{"autoLightBrightnessEnable":false,"autoTemperatureBrightnessEnable":false,"lightBrightness":0,"styleType":"style_type_auto_brightness","temperatureBrightness":0},"style_type_bright":{"autoLightBrightnessEnable":false,"autoTemperatureBrightnessEnable":false,"lightBrightness":20,"styleType":"style_type_bright","temperatureBrightness":0},"style_type_close_brightness":{"autoLightBrightnessEnable":false,"autoTemperatureBrightnessEnable":false,"lightBrightness":0,"styleType":"style_type_custom","temperatureBrightness":0},"style_type_custom":{"autoLightBrightnessEnable":false,"autoTemperatureBrightnessEnable":false,"lightBrightness":0,"styleType":"style_type_custom","temperatureBrightness":0},"style_type_soft":{"autoLightBrightnessEnable":false,"autoTemperatureBrightnessEnable":false,"lightBrightness":7,"styleType":"style_type_soft","temperatureBrightness":12}},"supportAutoBrightness":false,"versionCode":4}
```

Frenzie:
> Codacy reports Complexity +65, but this mainly comes from reproducing the same behaviour as the Onyx framework.

That particular metric is mostly quite useless as far as I'm concerned.

I think there might be some minor conflict with #585 so we'll have to check which to merge first.

app/src/main/java/org/koreader/launcher/device/LightsFactory.kt
```diff
@@ -82,7 +81,8 @@ object LightsFactory {
                 DeviceInfo.Id.ONYX_POKE4LITE,
                 DeviceInfo.Id.ONYX_TAB_ULTRA,
                 DeviceInfo.Id.STORYTEL_READER2,
-                -> {
+                DeviceInfo.Id.ONYX_GO7,
+                    -> {
```
Frenzie:
```suggestion
                -> {
```

Frenzie:
> I think there might be some minor conflict

Extremely minor, so merging this one first.

PR #593, opened 2026-05-19
Android network info

backcountrymountains:
I apologize if this is not good, I don't know how to program. 
I have a Nook Glowlight 4 plus that runs android 8.1.0. The device it lacks a navigation bar or a back button so when I wanted to turn on wifi, it would navigate to the wifi settings page but it was a struggle to get back into koreader. I asked Claude if there was a way to fix this since for android 9 and below it is possible to just toggle wifi. 

the Settings->Network->Network info menu didn't display anything but "Connected to Wi-Fi" so I had Claude do some magic so that it would show the SSID, Ip, and Mac. On my device it requires root to get the mac by running a dumpsys command, but Claude assured me that no issues would be caused if someone didn't have root.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/593)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -11,6 +11,8 @@
     <!-- common android permissions -->
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+    <uses-permission android:name="android.permission.CHANGE_WIFI_STATE" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"
         android:maxSdkVersion="29" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"
--- a/app/src/main/java/org/koreader/launcher/LuaInterface.kt
+++ b/app/src/main/java/org/koreader/launcher/LuaInterface.kt
@@ -65,7 +65,9 @@ interface LuaInterface {
     fun isWarmthDevice(): Boolean
     fun needsWakelocks(): Boolean
     fun openLink(url: String): Boolean
+    fun isWifiEnabled(): Boolean
     fun openWifiSettings()
+    fun setWifiEnabled(enable: Boolean): Boolean
     fun performHapticFeedback(constant: Int, force: Int)
     fun requestIgnoreBatteryOptimizations(rationale: String, okButton: String, cancelButton: String)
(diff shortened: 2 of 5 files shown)


Frenzie:
We'll have to reject this because it adds unnecessary permissions that'll raise a lot of questions ("why does this app need to toggle wifi?")

How would you normally get back to the home screen? There's no swipe from the bottom or something like that?

backcountrymountains:
That's fine, I understand. I have to use the home button to navigate on the eReader; it's super annoying! Not only do I only have the home button available to navigate; it's a capacitive button and it takes a long press to activate so I have to put my finger on it for around 1 second. If I touch the button for too little time, it does nothing; if I touch the button for too long, it also does nothing! Drives me nuts! I think the version of android (8.1) is so old that swipe gestures don't even exist for it. Android 8 was released in 2017! Yikes.

Anyway, I leave wifi off most of the time, to conserve battery, but sometimes I want to add a book to the device wirelessly from my phone using filebrowser, so the ability to turn on and off wifi simply is a nice usabilty improvement for me.

I think one of the AIs that I was using said that it might be possible to create some sort of settings dialog that wouldn't require deprecated permissions (wifi on/off was removed for android 10+) but would have some sort of back button or close dialog functionality. 

Sorry if I'm spamming with PR's I think I might have given Claude too many permissions to do things without approval.

Frenzie:
I'll close this one for now.

backcountrymountains:
> We'll have to reject this because it adds unnecessary permissions that'll raise a lot of questions ("why does this app need to toggle wifi?")
> 
> How would you normally get back to the home screen? There's no swipe from the bottom or something like that?

I just wanted to update that there is some functionality for setting the wifi state using the native com.nook.partner app that wouldn't require new permissions for koreader (although the permissions are deprecated in android 10+), but it has some issues with implementation that Claude documented [here](https://github.com/koreader/android-luajit-launcher/blob/66d42a2f9c88d140d6bd878101409e06e884efcb/docs/nook-gl4plus-wifi-analysis.md)

There doesn't seem to be a clean way to have koreader directly turn on and off wifi from within the app without synchronization issues between user toggling of wifi outside of koreader and koreader toggling wifi based on what koreader's last recorded wifi state. I'm not sure if it's an acceptable trade-off. 

Claude did say that changing the "Network info" to have the IP and MAC information shouldn't require added permissions to koreader, so maybe I can do that as a PR. It really helps me to know those two things so I can more easily connect to koreader to transfer files.

Thanks for your patience!

Frenzie:
I'm not super enthused about having something so singular to a specific device but that's a much more attractive proposition in any case.

PR #594, opened 2026-06-06
Android: add auto-rotation support

chofuhoyu:
This PR adds the Android-side changes needed for auto-rotation support in KOReader (see companion PR in the main repository).

## Changes

- **AndroidManifest.xml**: Change `screenOrientation` from `nosensor` to `fullSensor`, allowing Android to manage screen rotation via the accelerometer.
- **ActivityExtensions.kt**: Add `setAutoOrientation()` extension and a sensor mode guard in `setOrientationCompat` that passes `FULL_SENSOR` and related constants through without remapping.
- **MainActivity.kt**: Add `setScreenAutoRotation(enabled)` method to enable/disable auto mode via `requestedOrientation`.
- **LuaInterface.kt**: Declare the new JNI method.
- **assets/android.lua**: Add `android.orientation.setAuto(enabled)` binding.

## Dependencies

This PR must be merged **before** the companion PR in `koreader/koreader`, which updates the luajit-launcher submodule pointer and implements the Lua-side auto-rotation UI and logic.

Companion PR: https://github.com/koreader/koreader/pull/15507

## Testing

- Built and tested on an Android ARM64 device (Onyx Boox Leaf5)
- Auto-rotation follows device orientation correctly
- Manual rotation selection exits auto mode and locks to the selected orientation
- Physical page-turn buttons follow rotation correctly in both modes

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/594)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -39,7 +39,7 @@
             android:exported="true"
             android:theme="@style/SplashScreen"
             android:resizeableActivity="false"
-            android:screenOrientation="nosensor"
+            android:screenOrientation="fullSensor"
             android:launchMode="singleInstance"
             android:configChanges="colorMode|density|fontScale|keyboard|keyboardHidden|layoutDirection|locale|mcc|mnc|navigation|orientation|screenLayout|screenSize|smallestScreenSize|touchscreen|uiMode"
             tools:ignore="NonResizeableActivity">
--- a/app/src/main/java/org/koreader/launcher/extensions/ActivityExtensions.kt
+++ b/app/src/main/java/org/koreader/launcher/extensions/ActivityExtensions.kt
@@ -303,11 +303,11 @@ fun Activity.getOrientationCompat(isLandscape: Boolean): Int {
 }
 
 fun Activity.setOrientationCompat(isLandscape: Boolean, orientation: Int) {
-	    // Pass through sensor modes without remapping
-	    if (isPassthroughOrientation(orientation)) {
-	        requestedOrientation = orientation
-	        return
-	    }
+    // Pass through sensor modes without remapping
+    if (isPassthroughOrientation(orientation)) {
+        requestedOrientation = orientation
+        return
+    }
     val newOrientation = if (isLandscape) {
         when (orientation) {
             ANDROID_LANDSCAPE -> ANDROID_PORTRAIT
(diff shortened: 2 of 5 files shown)


pazos:
thanks for the contribution!!! all is looking great!

Please see https://app.codacy.com/gh/koreader/android-luajit-launcher/pull-requests/594/issues#issue-272fb7d1f4c1bcb5b713ccf61077d2de

I agree with codacy this time :)

app/src/main/AndroidManifest.xml
```diff
@@ -39,7 +39,7 @@
             android:exported="true"
             android:theme="@style/SplashScreen"
             android:resizeableActivity="false"
-            android:screenOrientation="nosensor"
+            android:screenOrientation="fullSensor"
```
pazos, line 42: isn't "fullsensor" the default when no screenOrientation is specified? in that case I would prefeer to remove the flag

chofuhoyu: According to the [document](https://developer.android.com/guide/topics/manifest/activity-element#screen), the default value will be `"unspecified"`. So setting it explicitly would be necessary.

app/src/main/java/org/koreader/launcher/extensions/ActivityExtensions.kt
```diff
@@ -331,6 +362,7 @@ private const val ANDROID_LANDSCAPE = 0
 private const val ANDROID_PORTRAIT = 1
 private const val ANDROID_REVERSE_LANDSCAPE = 8
 private const val ANDROID_REVERSE_PORTRAIT = 9
+	private const val ANDROID_FULL_SENSOR = 10
```
Frenzie: nit

Frenzie (approved):
Fine by me. (Mind, only relevant if the others are too busy for the next week.)

app/src/main/java/org/koreader/launcher/extensions/ActivityExtensions.kt
chofuhoyu: I've also found other inconsistent tabs added accidentally by editor settings. Also fixed that.

hugleo:
It might be worth testing on an emulator to verify that manual rotation still works as before on versions above 4.1, both with and without notch emulation. I can also test it on my devices and phone.

chofuhoyu:
> It might be worth testing on an emulator to verify that manual rotation still works as before on versions above 4.1, both with and without notch emulation. I can also test it on my devices and phone.

Thanks. It needs more tests. Please let me know if you tried this patch on Android emulator, devices or phone, and whether it works fine.

PR #595, opened 2026-06-15
Moaan W7 Support

kerivin:
[Apparently](https://github.com/koreader/koreader/pull/15542), KOReader treats Moaan W7 as if it had a color screen.
I've also added it to the no-light list since it has no lights.
Everything else (except probably Wi-fi) has been working fine with the default `Id.NONE` so I don't need to add the new Id anywhere else, do I?..

The specs were taken from logs:
```
MANUFACTURER: allwinner
BRAND       : allwinner
MODEL       : epd103
DEVICE      : virgo-perf1
PRODUCT     : virgo_perf1
HARDWARE    : sun8iw15p1
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/595)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -76,6 +76,7 @@ object DeviceInfo {
         MEEBOOK_P6,
         MOBISCRIBE_WAVE,
         MOAAN_MIX7,
+        MOAAN_W7,
         MOOINKPLUS2C,
         NABUK,
         NOOK,
@@ -355,6 +356,9 @@ object DeviceInfo {
             // Moaan Mix7
             MANUFACTURER == STR_ROCKCHIP && MODEL == "moaanmix7"
             -> Id.MOAAN_MIX7
+
+            BRAND == "allwinner" && MODEL == "epd103" && DEVICE == "virgo-perf1" && HARDWARE == "sun8iw15p1"
+            -> Id.MOAAN_W7
 
             // Mooink Plus 2c
             BRAND == "allwinner" && MODEL == "mooink plus 2c"
@@ -727,6 +731,7 @@ object DeviceInfo {
             Id.ONYX_NOTE,
             Id.SONY_CP1,
             Id.SONY_RP1,
+            Id.MOAAN_W7,
             -> true else -> false
         }
 
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -727,12 +727,12 @@ object DeviceInfo {
         // Android devices without lights
         QUIRK_NO_LIGHTS = when (ID) {
             Id.LINFINY_ENOTE,
+            Id.MOAAN_W7,
             Id.ONYX_MAX,
             Id.ONYX_MAX2_PRO,
             Id.ONYX_NOTE,
             Id.SONY_CP1,
             Id.SONY_RP1,
-            Id.MOAAN_W7,
             -> true else -> false
         }
 
--- a/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
+++ b/app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
@@ -357,6 +357,7 @@ object DeviceInfo {
             MANUFACTURER == STR_ROCKCHIP && MODEL == "moaanmix7"
             -> Id.MOAAN_MIX7
 
+            // Moaan W7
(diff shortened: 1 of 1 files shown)


Frenzie:
Did you do the compatibility test? File browser → tools → more tools → developer options → start compatibility test.

Edit: see https://github.com/koreader/koreader/issues/8482 for more info.

kerivin:
Here you go: 
[test.log](https://github.com/user-attachments/files/28973056/test.log)

Frenzie:
What about the driver(s) that work part?

kerivin:
I only see a bunch of errors and exceptions so I guess none of them work? x)
As I said, it works pretty much fine even with `Id.NONE`, the exception being fake light and color options (the device has none of it) + Wifi but that's not today.
So it's fine to let it use the default driver as it does now, isn't it?

Frenzie:
It's not a question of looking at the log but of looking at your screen. ;-)

> So it's fine to let it use the default driver as it does now, isn't it?

Not if one of the existing drivers works already. That'd be significantly better since then the correct screen update modes would work.

kerivin:
Ahh, the button "Test E-Ink" is supposed to refresh the screen as it says in the prompt! But I'm definitely not seeing that. Sorry for the confusion, I was trying to guess what the button should do. Nothing happens no matter which driver I choose. I guess that's a confirmation that none of them work x)

Frenzie:
Right, that should be okay then. That means the device will basically just guess when to do fast and slow refreshes. As long as it looks visually alright I suppose it doesn't matter but that's often not the case.

kerivin:
I mean, if there are some sources to look into for writing the device-specific controller, for sure I can try to figure out something. But first I would need to figure out what to look for in the first place 🥴

kerivin:
After doing some voodoo log gathering, I found this:
```
06-16 00:34:18.135  1000  1764  1877 V sunxihwc_eink: Layer:a73ba40c fd=47 refreshMode is EINK_GU16_MODE 
06-16 00:34:18.135  1000  1764  1877 V sunxihwc_eink: RGB
06-16 00:34:18.135  1000  1764  1877 V sunxihwc_eink: setupDisplayInfo: LINE:1490 handle->format=2,bpp0=4
06-16 00:34:18.135  1000  1764  1877 V sunxihwc_eink: Layer:a73ba04c fd=50 refreshMode is EINK_GU16_MODE 
06-16 00:34:18.135  1000  1764  1877 V sunxihwc_eink: RGB
06-16 00:34:18.135  1000  1764  1877 V sunxihwc_eink: setupDisplayInfo: LINE:1490 handle->format=1,bpp0=4
06-16 00:34:18.135  1000  1764  1877 V sunxihwc_eink: Layer:a73b91ec fd=69 refreshMode is EINK_GU16_MODE 
06-16 00:34:18.135  1000  1764  1877 V sunxihwc_eink: RGB
06-16 00:34:18.135  1000  1764  1877 V sunxihwc_eink: setupDisplayInfo: LINE:1490 handle->format=1,bpp0=4
06-16 00:34:18.135  1000  1764  1877 V sunxihwc_eink: de2commitToDisplay: frameCount=22124,w*h=(1872 ,1404)refreshMode=EINK_GU16_MODE,rectangleEnabled=0
```

```
mLayout=WM.LayoutParams{(0,0)(fillxfill) gr=#800033 ty=2013 fl=#10318 fmt=2 wanim=0x1030306 colorMode=0 refreshMode=84}
```

I have no idea what to do with this info now but it might be useful later

Frenzie:
Also see https://github.com/koreader/android-luajit-launcher/blob/cdb4d03f60b1e0b4ab67e6c0491a8acc35e17428/app/src/main/java/org/koreader/launcher/device/epd/NGL4EPDController.kt#L16

kerivin:
Yeah I'll look into it later, for now just a correct device detection would be enough since koreader already seems to work fine. It's still nice to get rid of fake light and color options. EPD stuff is too much for my non-Android brain x)

app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
```diff
@@ -727,6 +732,7 @@ object DeviceInfo {
             Id.ONYX_NOTE,
             Id.SONY_CP1,
             Id.SONY_RP1,
+            Id.MOAAN_W7,
```
hugleo: sort

hugleo (approved):

app/src/main/java/org/koreader/launcher/device/DeviceInfo.kt
kerivin: done

kerivin:
When I said 'I'll look into it later' I meant this PR can be merged, and if I ever get to the controller stuff it'll be in another PR x)

PR #599, opened 2026-07-12
Add Lenovo Smart/Yoga Paper support (EPD + lights)

4285f4:
Add full EPD refresh and brightness/warmth control for Lenovo Smart(Global version)/Yoga(China version) Paper.

  - EPD: full refresh via EinkManager.screenRefresh() reflection
  - Lights: brightness (0–99) and color temperature (0–24) encoded
    into Settings.System screen_brightness_float
  - Device detected by Brand=lenovo, Model=lenovo sp101fu
  
  Based on reverse-engineering of Lenovo's system EinkManager service
  and QuickSettingManager encoding scheme.

  Note: This patch has not been tested by building and running it as
  part of KOReader itself. However, the underlying EPD and lights
  control logic was verified with a standalone [test app](https://github.com/4285f4/smartpaper-eink-test).

  The test was performed on a Chinese-version device flashed with
  global firmware. The key difference between the two firmware
  variants:

  - Chinese version: only allows installing apps from the built-in
    app store and has ADB access locked down.
  - Global version: enables ADB by default and allows installation
    of third-party applications.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/599)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/lights/LenovoSmartPaperLightsController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/lights/LenovoSmartPaperLightsController.kt
@@ -69,7 +69,7 @@ class LenovoSmartPaperLightsController : LightsInterface {
         return try {
             val brightnessFloat = Settings.System.getFloat(
                 activity.contentResolver, "screen_brightness_float", 0f)
-            // Extract brightness raw value (0–99)
+            // Extract brightness raw value (0–99). The "rounding" logic purposefully follows original SystemUI.
             val raw = (brightnessFloat * 10000f).roundToInt() / 100
             raw.coerceIn(MIN, BRIGHTNESS_RAW_MAX)
         } catch (e: Exception) {
--- a/app/src/main/java/org/koreader/launcher/device/lights/LenovoSmartPaperLightsController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/lights/LenovoSmartPaperLightsController.kt
@@ -1,0 +1,145 @@
+package org.koreader.launcher.device.lights
+
+import android.app.Activity
+import android.provider.Settings
+import android.util.Log
+import org.koreader.launcher.device.LightsInterface
+import kotlin.math.roundToInt
+
+/* Lights Controller for Lenovo SmartPaper.
+ *
+ * Brightness and color temperature are encoded into a single float
+ * and written to Settings.System "screen_brightness_float", using the
+ * same encoding as Lenovo's system QuickSettingManager:
+ *
+ *   float value = (brightness / 100.0f) + (colorTemp / 10000.0f)
+ *
+ * Brightness raw range: 0–99
+ * Color temperature raw range: 0–24 (via PowerManager.BRIGHTNESS_COLORTEMP_MAX)
+ *
+ * Requires WRITE_SETTINGS permission (special permission, must be granted
+ * manually via Settings.ACTION_MANAGE_WRITE_SETTINGS).
+ */
+
+class LenovoSmartPaperLightsController : LightsInterface {
+
+    companion object {
+        private const val TAG = "Lights"
+        private const val BRIGHTNESS_RAW_MAX = 99
+        private const val DEFAULT_COLORTEMP_MAX = 24
+        private const val MIN = 0
+
+        // The UI exposes 0–24 steps for both brightness and warmth.
+        // This matches the color temp range and provides fine-grained
+        // brightness control within the 0–99 raw range.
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/koreader/launcher/device/lights/LenovoSmartPaperLightsController.kt
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    // ── Read ──────────────────────────────────────────────────────────────────
+
+    override fun getBrightness(activity: Activity): Int {
+        return try {
+            val brightnessFloat = Settings.System.getFloat(
+                activity.contentResolver, "screen_brightness_float", 0f)
+            // Extract brightness raw value (0–99)
+            val raw = (brightnessFloat * 10000f).roundToInt() / 100
```
Frenzie, line 73: Isn't that the same as:

```suggestion
            val raw = (brightnessFloat * 100f).roundToInt()
```

Frenzie (commented):
Seems fine besides the nit about the multiplication by 10k / 100 thing. @hugleo same for you?

app/src/main/java/org/koreader/launcher/device/lights/LenovoSmartPaperLightsController.kt
4285f4: Hi Frenzie, thanks for the suggestion.

The original logic comes from the SystemUI reverse‑engineering code [here](https://github.com/4285f4/smartpaper-eink-test/blob/b980f678df4b83eb6f8d9296074e3903953a28ab/temp/QuickSettingManager.java#L440):

I rewrote it in Kotlin to keep that exact behavior. Your suggested (value * 100).roundToInt() is not equivalent – it rounds to the nearest integer percentage, while the original rounds to two decimal places then truncates.

I tested this by looping from 0.0 to 1.0 with a step of 0.001; the original Kotlin matches the Java source 100%, while the suggestion mismatches on many values (e.g., 0.149f → 14 vs 15).

So I’d prefer to keep the original to preserve the SystemUI behavior. Let me know if you have further questions.

Frenzie: That sounds a bit odd. Is that a Kotlin peculiarity? The behavior you describe sounds like casting to int while `roundToInt` suggests… rounding.

Frenzie: > I tested this by looping from 0.0 to 1.0 with a step of 0.001; the original Kotlin matches the Java source 100%, while the suggestion mismatches on many values (e.g., 0.149f → 14 vs 15).

I see; I misunderstood what you were saying. It's the original that has the rounding error because it's working in integers.

```kotlin
import kotlin.math.*
fun main() {
    val brightnessFloat = 0.129f
    val original = (brightnessFloat * 10000f).roundToInt() / 100
    val simplified = (brightnessFloat * 100f).roundToInt()

    println("Original: $original")
    println("Simplified: $simplified")
}
```

```
Original: 12
Simplified: 13
```

If the original has what looks like a rounding error I guess that's how it should be? But I find it very strange.

4285f4: > If the original has what looks like a rounding error I guess that's how it should be? But I find it very strange.

Yeah I don't know why either. It does look odd, but it matches the SystemUI behavior and works fine, so I'd rather keep it. 😄

Frenzie: Maybe add a comment to that effect. :-)

app/src/main/java/org/koreader/launcher/device/lights/LenovoSmartPaperLightsController.kt
```diff
@@ -0,0 +1,145 @@
...
+        return false
+    }
+
+    // ── Read ──────────────────────────────────────────────────────────────────
+
+    override fun getBrightness(activity: Activity): Int {
+        return try {
+            val brightnessFloat = Settings.System.getFloat(
+                activity.contentResolver, "screen_brightness_float", 0f)
+            // Extract brightness raw value (0–99)
```
Frenzie: Something like this for example.
```suggestion
            // Extract brightness raw value (0–99). The "rounding" logic purposefully follows original SystemUI.
```

hugleo (approved):

PR #601, opened 2026-07-24
Set FONTDIR so MuPDF can find its fallback fonts

### LINKED ISSUE #15736: KOReader crashes on Android when browsing multiple dictionary results
# KOReader crashes on Android when browsing multiple dictionary results

## Summary

On Android, looking up a word in the dictionary and then navigating to another
dictionary result or word (next/previous dictionary button, or a new lookup
while the popup is open) reliably crashes the whole app.

## Environment

- Device: bigme hibreak (smartphone form factor), Android 14 (SDK 34)
- KOReader: built from `master` @ `a1b956d62` (also reproduces on current `master`)

## Steps to reproduce

1. Open any EPUB.
2. Long-press a word to open the dictionary popup.
3. Tap the "next dictionary result" (▷▷) button to move to a second
   dictionary source, particularly one that renders as HTML (e.g. a
   Wiktionary entry).
4. The app crashes and Android shows the "KOReader crashed" report screen.

## Root cause

Two independent bugs compound to cause this:

1. **MuPDF can't find its fallback fonts on Android.** MuPDF resolves its
   built-in fallback fonts (e.g. `noto/NotoSans-Regular.ttf`, used to render
   HTML content) via a path relative to `$FONTDIR`, defaulting to `./fonts`
   when unset (see `base/thirdparty/mupdf/external_fonts.patch`,
   `get_font_file()`). On Android, the process' cwd is `chdir()`'d to the
   app's private data directory (`getFilesDir()`) at startup
   (`android-luajit-launcher`'s `assets/android.lua`), not to the koreader
   home directory on external storage (`/sdcard/koreader`) where the actual
   `fonts` directory lives. `$FONTDIR` is never set, so every MuPDF-rendered
   HTML fallback-font lookup fails with:
   ```
   MuPDF cannot open document from text: cannot open ./fonts/noto/NotoSans-Regular.ttf: No such file or directory (0)
   ```

2. **That failure isn't always handled.** `HtmlBoxWidget:setContent()`
   (`frontend/ui/widget/htmlboxwidget.lua`) catches the first failure via
   `pcall`, retries with a plain-text fallback, but the retry hits the exact
   same font-resolution error and this time re-`error()`s. The very first
   dictionary lookup goes through `Trapper`'s `xpcall`
   (`readerdictionary.lua`), so that error is just logged as a warning. But
   `DictQuickLookup:update()` (`dictquicklookup.lua`) — used when switching
   between dictionary results or looking up another word while the popup is
   open — calls `setContent()` with **no** `pcall`/`xpcall` protection. The
   resulting unhandled Lua error propagates past the `NativeThread` script
   runner on Android ("Failed to run script"), which then calls `exit()`,
   killing the whole process instead of just failing to refresh the popup.

## Fix

Two companion PRs address both layers:

- `android-luajit-launcher`: set `$FONTDIR` at startup to the real koreader
  `fonts` directory on external storage, so MuPDF can always find its fonts
  regardless of the process' cwd.
- `koreader/koreader`: guard `DictQuickLookup:update()`'s `setContent()` call
  with `pcall`, so a rendering failure degrades gracefully (dictionary entry
  stays unrefreshed, warning logged) instead of crashing the app — this is
  defense-in-depth regardless of the font-path issue, since `setContent()`
  can fail for other reasons too (e.g. malformed HTML from a third-party
  dictionary).

Both were verified together on-device (crash no longer reproduces; the
previously-failing Wiktionary HTML entry now renders correctly).

jperon:
Companion fix to koreader/koreader#15737 (which fixes koreader/koreader#15736).

# PR (koreader/android-luajit-launcher) — Set FONTDIR so MuPDF can find its fallback fonts

**Branch:** `fix/fontdir-relative-path-crash` (base: `master`)
**Commit:** `0eccce26c79d81f0eb32c33997bb2dc4ab232743`

## Title

Set FONTDIR so MuPDF can find its fallback fonts

## Summary

- MuPDF resolves its built-in fallback fonts (e.g. `noto/NotoSans-Regular.ttf`,
  used whenever it renders HTML content — notably KOReader's HTML dictionary
  popup) via a path relative to the `$FONTDIR` environment variable,
  defaulting to `./fonts` when unset (see
  `base/thirdparty/mupdf/external_fonts.patch`, `get_font_file()`, in the
  koreader/koreader repo).
- That default is relative to the process' **current working directory**,
  which `assets/android.lua` `chdir()`s to the app's private data directory
  (`getFilesDir()`) very early at startup — not to the koreader home
  directory on external storage (`/sdcard/koreader` by default) where the
  actual `fonts` directory is extracted/lives.
- On devices without legacy full external-storage access (i.e. most modern
  Android installs, including this one — `MANAGE_EXTERNAL_STORAGE` was not
  granted), this makes **every** MuPDF HTML-fallback-font lookup fail with
  `cannot open ./fonts/noto/NotoSans-Regular.ttf: No such file or directory`,
  which in turn crashes koreader (see companion PR in koreader/koreader:
  `DictQuickLookup:update()` propagates this as an unhandled error on
  Android).
- Fix: explicitly `setenv("FONTDIR", ...)` right after the existing `chdir()`
  call, pointing at `$KO_HOME/fonts` (respecting `KO_HOME` if set, exactly
  like `datastorage.lua`'s `getDataDir()` does), before `llapp_main.lua` — and
  therefore any MuPDF usage — ever runs.

## Test plan

- [x] Built `arm64Rocks` debug APK against this branch + the koreader.git
      companion fix, installed on-device (Android 14, bigme hibreak).
- [x] Confirmed a previously-crashing HTML dictionary entry (Wiktionary FR-FR)
      now renders with correct formatting (bold/italic/lists), proving MuPDF
      resolves the fallback font correctly.
- [x] No regression to the existing `chdir()` behavior (app still starts,
      reads/writes `/sdcard/koreader` as before).

## Notes for reviewers

`FONTDIR` is only consulted by the `NOBUILTINFONT` build of MuPDF (see
`koreader/koreader`'s `base/thirdparty/mupdf/CMakeLists.txt`, which does
define `-DNOBUILTINFONT`), which is what koreader's Android build actually
uses — so this isn't dead configuration on this platform.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/601)

### DIFF
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -2731,15 +2731,12 @@ local function run(android_app_state)
         error(err)
     end
 
-    -- MuPDF resolves its built-in fallback fonts (eg. noto/NotoSans-Regular.ttf, used to
-    -- render the HTML dictionary popup) via a path relative to $FONTDIR, defaulting to
-    -- "./fonts" (see base/thirdparty/mupdf/external_fonts.patch) when unset. That default
-    -- is relative to the process' cwd, which we just changed to the app's private data
-    -- directory above -- not to the koreader home directory on external storage where the
-    -- actual "fonts" directory lives. Point FONTDIR there explicitly so MuPDF can always
-    -- find its fonts regardless of the process' cwd.
-    local koreader_home = os.getenv("KO_HOME") or (android.getExternalStoragePath() .. "/koreader")
-    C.setenv("FONTDIR", koreader_home .. "/fonts", 1)
+    -- MuPDF looks up its fallback fonts (eg. noto/NotoSans-Regular.ttf, used to render the
+    -- HTML dictionary popup) relative to $FONTDIR, defaulting to "./fonts" when unset (see
+    -- base/thirdparty/mupdf/external_fonts.patch). Point it at android.dir/fonts, where
+    -- Assets.kt's bootstrap() extracts the app's bundled "fonts" dir -- guaranteed present,
+    -- no external-storage access required.
+    C.setenv("FONTDIR", android.dir .. "/fonts", 1)
 
     dofile(android.dir.."/llapp_main.lua")
 end
--- a/assets/android.lua
+++ b/assets/android.lua
@@ -31,6 +31,7 @@ local ffi = require("ffi")
 ffi.cdef[[
 // posix:
 int chdir(const char *path);
+int setenv(const char *name, const char *value, int overwrite);
 // logging:
 int __android_log_print(int prio, const char *tag,  const char *fmt, ...);
 typedef enum android_LogPriority {
@@ -2729,6 +2730,16 @@ local function run(android_app_state)
         android.LOGE(err)
         error(err)
     end
+
+    -- MuPDF resolves its built-in fallback fonts (eg. noto/NotoSans-Regular.ttf, used to
+    -- render the HTML dictionary popup) via a path relative to $FONTDIR, defaulting to
+    -- "./fonts" (see base/thirdparty/mupdf/external_fonts.patch) when unset. That default
+    -- is relative to the process' cwd, which we just changed to the app's private data
+    -- directory above -- not to the koreader home directory on external storage where the
+    -- actual "fonts" directory lives. Point FONTDIR there explicitly so MuPDF can always
+    -- find its fonts regardless of the process' cwd.
+    local koreader_home = os.getenv("KO_HOME") or (android.getExternalStoragePath() .. "/koreader")
+    C.setenv("FONTDIR", koreader_home .. "/fonts", 1)
 
     dofile(android.dir.."/llapp_main.lua")
 end


assets/android.lua
```diff
@@ -2741,6 +2731,16 @@ local function run(android_app_state)
         error(err)
     end
 
+    -- MuPDF resolves its built-in fallback fonts (eg. noto/NotoSans-Regular.ttf, used to
+    -- render the HTML dictionary popup) via a path relative to $FONTDIR, defaulting to
+    -- "./fonts" (see base/thirdparty/mupdf/external_fonts.patch) when unset. That default
+    -- is relative to the process' cwd, which we just changed to the app's private data
+    -- directory above -- not to the koreader home directory on external storage where the
+    -- actual "fonts" directory lives. Point FONTDIR there explicitly so MuPDF can always
+    -- find its fonts regardless of the process' cwd.
```
Frenzie: I'm not sure if this makes sense. If it wants to find fonts we ship along it should point to something like `android.dir` something something `/fonts`, not to some external location.

jperon: Good catch — switched to android.dir/fonts (630c8ad), where Assets.kt already extracts the bundled fonts on install. No external storage needed.

assets/android.lua
```diff
@@ -2741,6 +2731,13 @@ local function run(android_app_state)
         error(err)
     end
 
+    -- MuPDF looks up its fallback fonts (eg. noto/NotoSans-Regular.ttf, used to render the
+    -- HTML dictionary popup) relative to $FONTDIR, defaulting to "./fonts" when unset (see
+    -- base/thirdparty/mupdf/external_fonts.patch). Point it at android.dir/fonts, where
+    -- Assets.kt's bootstrap() extracts the app's bundled "fonts" dir -- guaranteed present,
+    -- no external-storage access required.
+    C.setenv("FONTDIR", android.dir .. "/fonts", 1)
```
Frenzie, line 2739: If it's indeed that simple I'd think it belongs in <https://github.com/koreader/koreader/blob/6296a3cc7f2fecec0b5f4d4d4a54ff65b3f5f7c4/platform/android/llapp_main.lua> since it doesn't really seem to be a luajit-launcher concern. But maybe hold off on that for a minute; I want to take a look at what that referenced patch does exactly first.

jperon:
Sounds good, waiting on your look at the patch before moving anything.

Frenzie:
Alright, so the AI did come up with something that seems fairly elegant. Just like I said, provided it works it seems more at home in <https://github.com/koreader/koreader/blob/6296a3cc7f2fecec0b5f4d4d4a54ff65b3f5f7c4/platform/android/llapp_main.lua> seeing how it's a KOReader concern.

jperon:
Moving this to koreader.git's platform/android/llapp_main.lua per the discussion above — it's a KOReader/MuPDF concern, not a luajit-launcher one. Closing in favor of that.

jperon:
Moved to koreader/koreader#15738.

PR #602, opened 2026-08-02
Sunxi controller

kerivin:
With the help of the good old AI slop I was able to find a working controller for Moaan W7.
I've tested it: in the compatibility test it correctly flashes the entire screen, and refresh rate menu is here and seems to work correctly.

The only problem I'm not to sure about is the persistent refresh mode that stays after the refresh. It's fixed with a delay here, but without the delay the screen would full flash every minute in the compatibility test window when the time updates in the top bar (other windows are full-screen so no top bar and no time updates). I don't like the delay but haven't found anything better, so I guess the delay it is.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/602)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/TestActivity.kt
+++ b/app/src/main/java/org/koreader/launcher/TestActivity.kt
@@ -18,6 +18,7 @@ import org.koreader.launcher.device.epd.OnyxEPDController
 import org.koreader.launcher.device.epd.RK3026EPDController
 import org.koreader.launcher.device.epd.RK3368EPDController
 import org.koreader.launcher.device.epd.RK3566EPDController
+import org.koreader.launcher.device.epd.SunxiEPDController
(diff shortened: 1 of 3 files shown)


Frenzie (approved):
I hope a name like just `sunxi` won't come back to haunt in a year or two, but that's fine if there isn't anything more precise.

kerivin:
`MoaanW7` would be too specific, `sun8iw15p1` is a hardware for Tolino as well, so it would be misleading, 'Moaan' would be incorrect since there is another `Moaan` that runs on a rockchip... So I would prefer to keep 'Sunxi' unless you have an idea x)

hugleo:
It would be good to also include the driver in the eInk test.

Frenzie:
Very good point.

kerivin:
You mean the compatibility test? It's already added

PR #603, opened 2026-08-03
Sunxi: added binary A2 for fast refresh and other constants

kerivin:
Sorry for the fuss x) I've run some adb logging and found the actual constants used on the device and also found that faster binary refresh exists. So I replaced partial-grey GU16 with this binary A2. This A2 is apparently used for a drawing app from the vendor so it might be useful to have actually fast refresh in KOReader as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/603)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/epd/SunxiEPDController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/epd/SunxiEPDController.kt
@@ -25,22 +25,8 @@ class SunxiEPDController : EPDInterface {
         const val SUNXI_EINK_GLR16_MODE = 0x40
         const val SUNXI_EINK_GLD16_MODE = 0x80
         const val SUNXI_EINK_GU16_MODE = 0x84
-        const val SUNXI_EINK_GCK16_MODE = 0x90
-        const val SUNXI_EINK_GLK16_MODE = 0x94
-        const val SUNXI_EINK_CLEAR_MODE = 0x88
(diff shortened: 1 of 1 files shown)


Frenzie:
I didn't say anything earlier because it was completely different, but now it's starting to look more similar to this other driver?

https://github.com/koreader/android-luajit-launcher/blob/f1ab2b18fabc3c4d814a6e67c837e2ea7a2866ac/app/src/main/java/org/koreader/launcher/device/epd/NGL4EPDController.kt#L16-L46

kerivin:
But this driver doesn't work on my device unfortunately (because of "full-only" mode perhaps?). I would be less surprised if you mentioned [NookEmperorEPDController.kt](https://github.com/koreader/android-luajit-launcher/blob/f1ab2b18fabc3c4d814a6e67c837e2ea7a2866ac/app/src/main/java/org/koreader/launcher/device/epd/NookEmperorEPDController.kt) x)

Frenzie:
Same difference. My point is only suddenly it looks a lot more like a duplicate, which may be worth investigating.

kerivin:
Nah they use different refresh API, and my device doesn't have what NGL4 has. The same constants are probably because they all are Allwinner. The NookEmperor is much closer and even might work (but it was merged after Sunxi so I don't have it locally yet), but it tries to fix the same persistent-mode problem with `resume()` while I fixed it with the timer (the problem that also doesn't exist on NGL4 apparently)

PR #608, opened 2026-08-07
Fix EPD full refresh not working and add test entries for Lenovo Smart Paper

4285f4:
Fix EPD full refresh not working and add test entries for Lenovo Smart Paper

- EPD: fix waveform constants returning EPD_AUTO (0) which caused einkUpdate(0) to be discarded as 'invalid' in MainActivity.einkUpdate(). Changed all getWaveform* methods to return WAVEFORM_FULL (1) so the refresh request reaches setEpdMode().
- TestActivity: add EPD and lights test entries for Lenovo Smart Paper, with a tooltip for lights reminding users to grant WRITE_SETTINGS permission.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/608)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/epd/LenovoSmartPaperEPDController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/epd/LenovoSmartPaperEPDController.kt
@@ -4,6 +4,12 @@ import android.annotation.SuppressLint
 import android.util.Log
 import android.view.View
 import org.koreader.launcher.device.EPDInterface
+import org.koreader.launcher.device.epd.rockchip.RK33xxEPDController.Companion.EINK_MODE_FAST
+import org.koreader.launcher.device.epd.rockchip.RK33xxEPDController.Companion.EINK_MODE_FULL
+import org.koreader.launcher.device.epd.rockchip.RK33xxEPDController.Companion.EINK_MODE_FULL_UI
+import org.koreader.launcher.device.epd.rockchip.RK33xxEPDController.Companion.EINK_MODE_PARTIAL
+import org.koreader.launcher.device.epd.rockchip.RK33xxEPDController.Companion.EINK_MODE_PARTIAL_UI
+import org.koreader.launcher.device.epd.rockchip.RK33xxEPDController.Companion.EINK_WAVEFORM_DELAY
 
 /* EPD Controller for Lenovo SmartPaper (RK3566).
  *
@@ -19,12 +25,6 @@ class LenovoSmartPaperEPDController : EPDInterface {
 
     companion object {
         private const val TAG = "EPD"
-
-        // Waveform value for full-screen refresh. Must match the EPD_FULL constant
-        // in MainActivity.einkUpdate() (value 1) so the refresh request is not
-        // discarded as "invalid". The actual waveform used by EinkManager is
-        // determined inside setEpdMode(), not by this constant.
-        private const val WAVEFORM_FULL = 1
     }
 
     override fun getPlatform(): String {
@@ -36,35 +36,35 @@ class LenovoSmartPaperEPDController : EPDInterface {
     }
 
     override fun getWaveformFull(): Int {
-        return WAVEFORM_FULL
+        return EINK_MODE_FULL
     }
 
     override fun getWaveformPartial(): Int {
-        return WAVEFORM_FULL
+        return EINK_MODE_PARTIAL
     }
 
     override fun getWaveformFullUi(): Int {
-        return WAVEFORM_FULL
+        return EINK_MODE_FULL_UI
     }
 
     override fun getWaveformPartialUi(): Int {
-        return WAVEFORM_FULL
+        return EINK_MODE_PARTIAL_UI
     }
 
     override fun getWaveformFast(): Int {
-        return WAVEFORM_FULL
+        return EINK_MODE_FAST
     }
 
     override fun getWaveformDelay(): Int {
-        return WAVEFORM_FULL
+        return EINK_WAVEFORM_DELAY
     }
 
     override fun getWaveformDelayUi(): Int {
-        return WAVEFORM_FULL
+        return EINK_WAVEFORM_DELAY
     }
 
     override fun getWaveformDelayFast(): Int {
-        return WAVEFORM_FULL
+        return EINK_WAVEFORM_DELAY
     }
 
     override fun needsView(): Boolean {
@@ -72,10 +72,11 @@ class LenovoSmartPaperEPDController : EPDInterface {
     }
 
     @SuppressLint("WrongConstant")
-    override fun setEpdMode(targetView: View,
-                            mode: Int, delay: Long,
-                            x: Int, y: Int, width: Int, height: Int, epdMode: String?)
-    {
+    override fun setEpdMode(
+        targetView: View,
+        mode: Int, delay: Long,
+        x: Int, y: Int, width: Int, height: Int, epdMode: String?
+    ) {
         try {
             val einkManagerClass = Class.forName("android.os.EinkManager")
             val einkManager: Any? = targetView.context.getSystemService("eink")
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/koreader/launcher/device/epd/LenovoSmartPaperEPDController.kt
```diff
@@ -20,6 +19,12 @@ class LenovoSmartPaperEPDController : EPDInterface {
 
     companion object {
         private const val TAG = "EPD"
+
+        // Waveform value for full-screen refresh. Must match the EPD_FULL constant
+        // in MainActivity.einkUpdate() (value 1) so the refresh request is not
+        // discarded as "invalid". The actual waveform used by EinkManager is
+        // determined inside setEpdMode(), not by this constant.
+        private const val WAVEFORM_FULL = 1
```
Frenzie: The way of doing this seem a bit weird. Wouldn't this mean you should add a new EPD driver or something?

4285f4: Hi Frenzie,

This change does look a bit weird, but here's the background:

I downloaded the latest nightly build to test and found that calling full refresh does nothing at all. After digging into the code, I realized the request gets filtered out as "invalid" in `MainActivity#einkUpdate(int)` — only values 1, 2, 3, and 4 are accepted. So I just picked one of them (1) to make sure the call goes through.

As for why the full refresh value was 0 before: `LenovoSmartPaperEPDController` was written by following `RK3566EPDController`, and all methods that return waveform mode values (like `getWaveformFull()`) return `EPD_AUTO`, which is 0. In [framebuffer_android.lua](https://github.com/koreader/koreader-base/blob/533cc8235fd0b380b92f68fb9b2ce8f54970d59f/ffi/framebuffer_android.lua#L29), when the platform is "rockchip", _updateFull() calls `android.einkUpdate(full)`, and `full` here ultimately comes from `getWaveformFull()`. So `mode=0` plus `platform=rockchip` means the full refresh request never reaches `LenovoSmartPaperEPDController#setEpdMode()` — it gets blocked earlier.

I'm not sure why it's designed this way, but this was the simplest workaround I could think of.

Frenzie (approved):
I suppose. Pinging @hugleo

app/src/main/java/org/koreader/launcher/device/epd/LenovoSmartPaperEPDController.kt
hugleo: > As for why the full refresh value was 0 before: `LenovoSmartPaperEPDController` was written by following `RK3566EPDController`, and all methods that return waveform mode values (like `getWaveformFull()`) return `EPD_AUTO`, which is 0. In [framebuffer_android.lua](https://github.com/koreader/koreader-base/blob/533cc8235fd0b380b92f68fb9b2ce8f54970d59f/ffi/framebuffer_android.lua#L29), when the platform is "rockchip", _updateFull() calls `android.einkUpdate(full)`, and `full` here ultimately comes from `getWaveformFull()`. So `mode=0` plus `platform=rockchip` means the full refresh request never reaches `LenovoSmartPaperEPDController#setEpdMode()` — it gets blocked earlier.
> 
> I'm not sure why it's designed this way, but this was the simplest workaround I could think of.

IIRC, that's hardcoded because back then there was a bug where Rockchip devices didn't work correctly with drivers that had partial mode support, so it was easier to hardcode it here than create a separate driver just for that.

Currently, it should be better to create a separate driver, I think. But if there are a lot of drivers that support partial mode and aren't used by Rockchip devices, we'd have to duplicate a lot of drivers.

One day, if someone wants to pick up this task, they could check it, remove this from `android.lua`, and handle it directly in the drivers.

hugleo:
Since it is using:

```kotlin
override fun getMode(): String {
    return "full-only"
}
```

The only relevant method is `getWaveformFull`. Whatever you put in the other partial methods will suffice, since they won't be used anyway.

hugleo:
If you want, you can import `org.koreader.launcher.device.epd.rockchip.RK30xxEPDController.kt` or `RK33xxEPDController.kt`.

They have all the constants:

```kotlin
const val EINK_MODE_FULL = 1
const val EINK_MODE_PARTIAL = 2
const val EINK_MODE_FULL_UI = 4
const val EINK_MODE_PARTIAL_UI = 4
const val EINK_MODE_FAST = 3
const val EINK_WAVEFORM_DELAY = 0
```

That might work for you.

All drivers are importing a parent like `import org.koreader.launcher.device.epd.*`, which I think is a good practice, for example.

Another one that doesn't follow this practice, which I didn't notice during the review, is `SunxiEPDController.kt`. It implements `setEpdMode` directly instead of defining a new driver family.

Frenzie:
@kerivin Would you mind looking at the Sunxi remark above?

4285f4:
@hugleo Thanks for the review and suggestions! I've updated the PR to use the standard constants from RK33xxEPDController instead of hardcoding my own. All getWaveform* methods now return the appropriate EINK_MODE_* constants. Please take another look when you have a moment.

hugleo:
Doesn’t that work by just importing

import org.koreader.launcher.device.epd.rockchip.RK35xxEPDController

and using:

override fun setEpdMode(
    targetView: android.view.View,
    mode: Int,
    delay: Long,
    x: Int,
    y: Int,
    width: Int,
    height: Int,
    epdMode: String?
) {
    requestEpdMode(targetView)
}

?

I see that you implement 

You can define the additional constants in the RK35xxEPDController file if needed.



You can set additional parameters for requestEpdMode and implement the logic in RK35xxEPDController.


Maybe define a flag in RK35xxEPDController requestEpdMode method: if it is true, call sendOneFullFrame(), and if it is false, call your screenRefresh() implementation, for example.

Or you could even pass the mode value from setEpdMode() to requestEpdMode(targetView, mode). That should be enough to differentiate between the two implementations in RK35xxEPDController.

@pazos, is that what you had in mind when implementing this Template Method pattern?

4285f4:
@hugleo I see — you're suggesting extending `requestEpdMode()` with a flag or parameter to support both `sendOneFullFrame()` and my `screenRefresh()`.

I'd prefer not to modify `RK35xxEPDController` since `screenRefresh()` might be device-specific. Would it be okay to keep my implementation separate?

hugleo:
Yes,

In this case, maybe it’s better to define all the constants defined here as well.

If other drivers can benefit from them in the future, we can avoid duplicating code instead of making `RK35xxEPDController.kt` more complex, it might be better to create a separate `RK3566EPDController.kt` under `epd/rockchip.

I just asked @pazos whether it’s okay to keep the whole implementation here for now, or if it would be better to create `RK3566EPDController.kt`.

pazos (changes requested):
please just inherit from rk35xx controller, so you don't need to import companion stuff.

Override the setEpdMode and you are good.

Rest of changes are good to go

**Edit** If im missing something please let me know, I just did a flash skim over this PR :)

hugleo:
@pazos All right, Was just for confirming the program pattern with you ;-) 
It’s not required to import a parent package like `org.koreader.launcher.device.epd.rockchip`.

Just import the common stuff if needed, and if the driver is too specific, you can implement the EPD directly in the driver file.

@4285f4 Just import the full `rk35xx` package and you will remove a lot of import lines to pass codacy ;-)

4285f4:
> please just inherit from rk35xx controller, so you don't need to import companion stuff.

@pazos Since RK35xxEPDController doesn't have the EINK_MODE_* constants (only EPD_AUTO), would you prefer that I keep importing them from RK33xxEPDController, or should I just define them locally in this class to avoid the extra imports?

hugleo:
@4285f4 Is it OK to just import RK35xxEPDController. Let someone else define those variables if they ever need to use them there.

4285f4:
@pazos Thanks for the review! I've updated the PR to inherit from RK35xxEPDController and define the constants locally. Could you please take another look when you have a chance?

hugleo:
@4285f4 Now you’ve complicated things.
The previous code was good.

You just need to combine those multiple companion imports into a single import:

import org.koreader.launcher.device.epd.rockchip.RK33xxEPDController

If the constants don’t work for you, for example, EPD_FULL = 1 but you need EPD_FULL = 2, it’s OK to import RK35xxEPDController.kt and define the new constants there.

4285f4:
@hugleo Got it, I was overthinking it. I've reverted to using RK33xxEPDController with a single import. The constants are working correctly now. Thanks for the clarification.

hugleo (approved):

Frenzie (approved):

PR #614, opened 2026-08-25
Onyx: support BOOX all refresh mode

crypt0nX:
# Onyx: support BOOX all refresh mode

This changes `OnyxEPDController` from `full-only` to `all` mode.

BOOX may issue a delayed system refresh after an application refresh, which causes a second flash. Before KOReader sends its explicit refresh, the Onyx controller calls `ViewUpdateHelper.debouncer(false, 0, 0, 0, 0)`. This is an internal BOOX method, so `HiddenApiBypass` is needed on Android R and newer to allow the reflection call. `resumeBooxSystemRefresh()` is kept as a future hook for restoring the debouncer state, but it is not used yet.

The coordinate conversion follows the official BOOX SDK definition of `BaseDevice.refreshScreenRegion(View, left, top, width, height, UpdateMode)` ([onyxsdk-device 1.3.5.2](https://repo.boox.com/repository/maven-public/com/onyx/android/sdk/onyxsdk-device/1.3.5.2/onyxsdk-device-1.3.5.2.aar)). KOReader starts with `x, y, width, height`, but its Android framebuffer bridge sends `x, y, x + width, y + height`, and `MainActivity` forwards those values unchanged. The Onyx controller therefore converts the received right and bottom coordinates back to width and height with `width - x` and `height - y` before calling the region refresh method.

The BOOX-specific changes stay in `OnyxEPDController`; the shared Qualcomm controller is not used to handle the BOOX debouncer.

References:

- [BOOX OnyxAndroidDemo](https://github.com/onyx-intl/OnyxAndroidDemo)
- [BOOX EPD screen update documentation](https://github.com/onyx-intl/OnyxAndroidDemo/blob/master/doc/EPD-Screen-Update.md)  
- [BOOX EPD update modes](https://github.com/onyx-intl/OnyxAndroidDemo/blob/master/doc/EPD-Update-Mode.md)
- [AndroidHiddenApiBypass](https://github.com/LSPosed/AndroidHiddenApiBypass)

The `debouncer` method is not described in the public SDK. It was recovered from the BOOX T10C Plus firmware by decompiling `/system/framework/framework.jar` (`classes2.dex`), class `android.onyx.ViewUpdateHelper` (`ViewUpdateHelper.java`). The decompiled method has the signature `debouncer(boolean, int, int, int, int)` and writes its five arguments to the SurfaceFlinger debouncer transaction.

Tested on a BOOX device with EPUB documents.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/614)

### DIFF
--- a/app/src/main/java/org/koreader/launcher/device/epd/OnyxEPDController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/epd/OnyxEPDController.kt
@@ -3,7 +3,6 @@
 package org.koreader.launcher.device.epd
 
 import android.util.Log
-import org.koreader.launcher.BuildConfig
 import org.koreader.launcher.device.EPDInterface
 import org.koreader.launcher.device.epd.qualcomm.QualcommEPDController
 
@@ -36,17 +35,20 @@ class OnyxEPDController : QualcommEPDController(), EPDInterface {
 
     private fun resumeBooxSystemRefresh(): Boolean {
         return try {
-            // The firmware restores the configured SF debouncer through this
-            // internal OnyxEpdBypassManager method, rather than fixed values.
-            val managerClass = Class.forName("android.onyx.optimization.OnyxEpdBypassManager")
-            val manager = managerClass.getMethod("sharedInstance").invoke(null)
-            managerClass.getDeclaredMethod("resumeSFDebouncer", String::class.java).apply {
-                isAccessible = true
-            }.invoke(manager, BuildConfig.APPLICATION_ID)
-            Log.i(TAG, "called OnyxEpdBypassManager.resumeSFDebouncer(${BuildConfig.APPLICATION_ID})")
+            // Ask BOOX's EInkHelper to reapply the current debouncer setting.
+            val helperClass = Class.forName("android.onyx.optimization.EInkHelper")
+            val duration = (helperClass.getMethod("getAnimationDuration")
+                .invoke(null) as Number).toInt()
+            if (duration < 0) {
+                Log.w(TAG, "BOOX E-Ink service is not ready")
+                return false
+            }
+            helperClass.getMethod("setAnimationDuration", Integer.TYPE)
+                .invoke(null, duration)
+            Log.i(TAG, "reapplied BOOX debouncer setting (animationDuration=$duration)")
             true
         } catch (e: Exception) {
-            Log.e(TAG, "OnyxEpdBypassManager.resumeSFDebouncer reflection failed: ${e.message}", e)
+            Log.e(TAG, "EInkHelper debouncer restore failed: ${e.message}", e)
             false
         }
     }
--- a/app/src/main/java/org/koreader/launcher/device/epd/OnyxEPDController.kt
+++ b/app/src/main/java/org/koreader/launcher/device/epd/OnyxEPDController.kt
@@ -2,19 +2,39 @@
 
 package org.koreader.launcher.device.epd
 
+import android.os.Build
 import android.util.Log
 import org.koreader.launcher.device.EPDInterface
 import org.koreader.launcher.device.epd.qualcomm.QualcommEPDController
+import org.lsposed.hiddenapibypass.HiddenApiBypass
 
 class OnyxEPDController : QualcommEPDController(), EPDInterface {
 
     private companion object {
         const val TAG = "EPD"
+
+        private val hiddenApiAccess: Boolean by lazy {
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
+                true
+            } else {
+                runCatching {
+                    HiddenApiBypass.addHiddenApiExemptions(
+                        "Landroid/onyx/ViewUpdateHelper;",
+                        "Landroid/onyx/optimization/EInkHelper;",
+                    )
+                }.onFailure { error ->
+                    Log.e(TAG, "BOOX EPD hidden API exemption failed", error)
+                }.getOrDefault(false)
+            }
+        }
     }
 
     private var debouncerPrevented = false
 
     private fun preventBooxSystemRefresh(): Boolean {
+        if (!hiddenApiAccess) {
+            return false
+        }
         return try {
             // Official BOOX API: ViewUpdateHelper.debouncer(false, 0, 0, 0, 0).
             Class.forName("android.onyx.ViewUpdateHelper").getMethod(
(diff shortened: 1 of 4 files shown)


Frenzie:
@Galunid Does this still mean anything to you? :-)

app/src/main/java/org/koreader/launcher/device/epd/OnyxEPDController.kt
```diff
@@ -55,9 +99,21 @@ class OnyxEPDController : QualcommEPDController(), EPDInterface {
                             mode: Int, delay: Long,
                             x: Int, y: Int, width: Int, height: Int, epdMode: String?)
     {
-        requestEpdMode(targetView, mode, delay, x, y, width, height)
+        if (!debouncerPrevented) {
+            debouncerPrevented = preventBooxSystemRefresh()
+        }
+        // KOReader passes right/bottom, while current Onyx firmware expects width/height.
+        requestEpdMode(targetView, mode, delay, x, y, width - x, height - y)
```
Frenzie, line 131: If I understand correctly this would break it on older firmware while the effect of this change is required for newer firmware?

hugleo:
If it's mimicking the official Onyx SDK, it should be okay for any onyx device.

But the problem is the inclusion of the external LSPosed library. Since it's causing Android "crashes" to expose reflection methods, it might not be a good idea to include it. For example, in the e-ink test, we're  "crashing" all Android devices that is trying to test this drivers.

This should work without the library by setting the following ADB option:

adb shell settings put global hidden_api_policy 1

app/src/main/java/org/koreader/launcher/device/epd/OnyxEPDController.kt
crypt0nX: You are referring to the coordinate conversion on this line, right?

This is not intended to be a newer-firmware-only change. KOReader's Android framebuffer bridge passes the region as:

`x, y, x + width, y + height`

However, BOOX's View.refreshScreen() expects:

`left, top, width, height, mode`

Therefore the controller converts the received right/bottom values back to width/height with `width - x` and `height - y`.

I also checked the older Max3 Android 9 firmware. Its `View.refreshScreen()` implementation uses the same width/height semantics, so this conversion is valid there as well. 

I don't have an earlier BOOX device available for testing; the oldest firmware I could inspect is the Max3 running BOOX OS 3.5.

crypt0nX:
> If it's mimicking the official Onyx SDK, it should be okay for any onyx device.
> 
> But the problem is the inclusion of the external LSPosed library. Since it's causing Android "crashes" to expose reflection methods, it might not be a good idea to include it. For example, in the e-ink test, we're "crashing" all Android devices that is trying to test this drivers.
> 
> This should work without the library by setting the following ADB option:
> 
> adb shell settings put global hidden_api_policy 1

There is another possible implementation that avoids the LSPosed dependency:

```kotlin
val data = Parcel.obtain()
return try {
    val serviceManager = Class.forName("android.os.ServiceManager")
    val surfaceFlinger = serviceManager.getMethod(
        "getService",
        String::class.java
    ).invoke(null, "SurfaceFlinger") as? IBinder ?: return false

    data.writeInterfaceToken("android.ui.ISurfaceComposer")
    repeat(5) { data.writeInt(0) }

    surfaceFlinger.transact(0x00ff0064, data, null, 0)
} finally {
    data.recycle()
}
```

The transaction code `0x00ff0064` and the five integer arguments correspond to the BOOX firmware's internal implementation of:

```kotlin
ViewUpdateHelper.debouncer(false, 0, 0, 0, 0)
```

This avoids modifying Android Runtime or using `HiddenApiBypass`. However, it relies on undocumented BOOX Binder details, including the transaction code and Parcel layout, so it may be difficult to maintain and could become incompatible with future firmware updates.

The dependency-free approach also has limitations: it does not provide a way to resume the debouncer, although that is not currently needed. It may also vary between firmware versions, since the transaction code and Parcel layout are undocumented.

In my opinion, as newer e-ink tablets increasingly ship with newer Android versions, hidden APIs may become necessary for some vendor-specific features, especially when the required functionality is not exposed in the public SDK. However, I understand if adding this dependency is not considered appropriate at this stage. If so, I can try the direct Binder approach instead.

hugleo:
We unfortunately cannot include the LSPosed (HiddenApiBypass) library at the time. Some reasons that I can think of:

Maintenance: It can be a complex dependency to maintain and is not guaranteed to work on future Android versions.
Scope & Security: Adding it to main affects all devices. It unlocks all hidden methods globally by default, rather than strictly whitelisting only the necessary ones.
Stability: This specific library is known to cause crashes on certain devices (e.g., some Huawei models).

Even if we were to use a bypass, it would need to be strictly isolated to the specific driver that requires it, not applied globally.

The Binder approach can be an alternative, but it must be implemented as a new driver exclusively for your specific Onyx device, as there is no guarantee this Parcel layout will work across all Onyx models.

crypt0nX:
> We unfortunately cannot include the LSPosed (HiddenApiBypass) library at the time. Some reasons that I can think of:
> 
> Maintenance: It can be a complex dependency to maintain and is not guaranteed to work on future Android versions. Scope & Security: Adding it to main affects all devices. It unlocks all hidden methods globally by default, rather than strictly whitelisting only the necessary ones. Stability: This specific library is known to cause crashes on certain devices (e.g., some Huawei models).
> 
> Even if we were to use a bypass, it would need to be strictly isolated to the specific driver that requires it, not applied globally.
> 
> The Binder approach can be an alternative, but it must be implemented as a new driver exclusively for your specific Onyx device, as there is no guarantee this Parcel layout will work across all Onyx models.

Thanks. I’ve changed it so `HiddenApiBypass` is no longer enabled globally.

It’s now only used lazily in `OnyxEPDController`, and only for the two BOOX classes needed here:

```kotlin id="h8ceq1"
Landroid/onyx/ViewUpdateHelper;
Landroid/onyx/optimization/EInkHelper;
```

If the bypass fails, the BOOX debouncer handling is skipped and the normal refresh path still works.

I kept the dependency for now instead of using the direct SurfaceFlinger transaction, since that would mean hard-coding undocumented BOOX Binder details.

Would this narrower use be acceptable?

hugleo:
Even if the dependency is specific to an individual driver, the other concerns I mentioned still apply (and there are others, such as the driver being tested in the E Ink test  which allows all devices to be tested with this driver,  the library targets API 37 and we  try to target 34).

I think @pazos will agree with me that the LSPosed library can’t be included at this time.

PR #615, opened 2026-08-27
Apply android.supports_size_changes to the main activity

xeropresence:
Workaround for https://github.com/koreader/koreader/issues/14423

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/android-luajit-launcher/615)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -44,6 +44,7 @@
             android:configChanges="colorMode|density|fontScale|keyboard|keyboardHidden|layoutDirection|locale|mcc|mnc|navigation|orientation|screenLayout|screenSize|smallestScreenSize|touchscreen|uiMode"
             tools:ignore="NonResizeableActivity">
             <meta-data android:name="android.app.lib_name" android:value="luajit-launcher" />
+            <meta-data android:name="android.supports_size_changes" android:value="true" />
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />


Frenzie (approved):
I think it's probably fine.

hugleo:
Maybe this will throw APP_CMD_WINDOW_RESIZED events. We currently don't handle APP_CMD_WINDOW_RESIZED.

Need to test if there are any side effects. I think if throw APP_CMD_WINDOW_RESIZED the effects will be null. We need to see if APP_CMD_CONFIG_CHANGED is also thrown on other devices to update the dimensions when rotate.

Last time I had fun with this, I ran some tests. At first glance, APP_CMD_WINDOW_RESIZED should implement the same thing as APP_CMD_CONFIG_CHANGED.

https://github.com/hugleo/koreader/blob/16b4781c061c1ed61c6394de32261d392367bd0b/frontend/device/android/device.lua#L168

https://github.com/koreader/koreader/blob/master/frontend/device/android/device.lua#L160

xeropresence:
Been using this locally since the Pr was put in, haven't noticed any issues.

hugleo:
It would be good to test this on a few devices by rotating them multiple times—especially those with a notch—to observe the behavior.

Normally, the `C.APP_CMD_CONFIG_CHANGED` event is generated and the resize proceeds, but on some devices only `C.APP_CMD_WINDOW_RESIZED` is thrown (in which case that breaks notch handling or inverted orientation, so we need the resize on this event as well).

If any problems occur, check if this fixes them:

```lua
elseif ev.code == C.APP_CMD_CONFIG_CHANGED
    or ev.code == C.APP_CMD_WINDOW_RESIZED then
        -- the stuff we already do

```

xeropresence:
I installed it on my p9, and couldn't get it to rotate under any condition, installed stock koreader and same situation. 
Tried executing the adb commands to force a rotation, eg settings put system user_rotation 3, had no effect on the koreader app.
My only other device that forces rotation is a boox and it doesn't have a notch.

hugleo:
I've tried it on my device with a notch, and things are working normally, so it's probably fine.

hugleo (approved):

